PR #26, opened 2020-09-01
Added documentation to the bulk of the api; fixed the run target to reset db; returns ecoverse name correctly - closes #21

### LINKED ISSUE #21: Documentation of api
## Description
As a developer I want documentation within the graphql schema so that I can effectively use it. 

## Acceptance criteria

- [ ] All types to have descriptions, judged by the reviewer to be understandable by a developer
- [ ] Non-trivial fields to also have descriptions

techsmyth:

Adds documentation to the bulk of the graphql api; fixed run target for resetting db script

### Additional context

Add any other context

### Governance 

- [X] Documentation is added
- [x] Reset db script follow by run + executing a sample query as below working:
query  {
  name
  context {
    description
  }
  challenges {
    id
    name
    context {
      description
    }
    tags {
      name
    }
    groups {
      name
    }
  }
  
  userGroups {
    name
    members {
      name
    }
  }
  
  users {
    id
    name
  }

}


src/models/entities/Challenge.ts
```diff
@@ -9,32 +9,32 @@ export class Challenge extends BaseEntity {
...
-  @OneToOne(type => Context,  { eager: true, cascade: true})
+  @Field(() => Context, {nullable: true, description: "The shared understanding for the challenge"})
+  @OneToOne(type => Context,  { eager: true, cascade: true })
   @JoinColumn()
   context?: Context;
 
   // Community
 
-  @Field(() => UserGroup)
+  @Field(() => UserGroup, {nullable: true, description: "The leads for the challenge. The focal point for the user group is the primary challenge lead."})
```
atstoyanov, line 23: We should discuss which quotes to use. Best practice is to use single quotes for strings. But in most case is personal preference. Also a rule should be configured to keep the quotes consistent in all files. Also this should be done in the other repositories (at least for the Client.web). Linting check can be part of the PR requirements.

atstojanov (changes requested):

src/models/entities/Challenge.ts
atstoyanov: I've added and issue to address this on global level #30

atstojanov (approved):

PR #68, opened 2020-09-15
Make graphql endpoint port configurable

Bogoev91:

Endpoints through which client and server communicate need to be configurable

It makes those endpoints/ports configurable through environment variables

### Additional context

Demo is not yet updated

### Governance 

- [x] Documentation is added


package.json
```diff
@@ -7,6 +7,7 @@
   "license": "ISC",
   "scripts": {
     "start": "nodemon -w src --ext ts --exec ts-node src/server.ts",
+    "start-local":"nodemon -w src --ext ts --exec ts-node -r dotenv/config src/server.ts",
```
atstoyanov, line 10: What is the point of start-local? If won't be used for development nodemon is not required. 
Also "start:local" is more suitable name for the scripts. This allow having different purpose of same script. Like (dev, production etc.).

src/server.ts
```diff
@@ -30,9 +30,9 @@ const main = async () => {
   const app = express();
   apolloServer.applyMiddleware({ app });
   app.listen(
```
atstoyanov, line 35: With this solution we rely on having the endpoint set. There should be a default value or configuration validation:
- add default post ( const serverPort = process.env.GRAPHQL_ENDPOINT_PORT || 4000)
- validate configuration (check ConfigurationValidator)

atstojanov (dismissed):
Include the new environment variable in the README.md

src/server.ts
valentinyanakiev: const ENDPOINT_PORT :)

package.json
valentinyanakiev: I agree here - remove this and leave it as it is.If we have a default port (see below) configuration for port 4000, this change here won't be necessary

valentinyanakiev (changes requested):
Some little tweaks required.

valentinyanakiev (approved):

valentinyanakiev (approved):
👍

valentinyanakiev (approved):
👍

PR #80, opened 2020-09-18
corrected version number

techsmyth:
Corrected version number of the server


atstojanov (approved):
I approve that, but we have to decide when the version will be updated and if it is possible to be updated automatically. npm-version could be used for versioning. It could be configured to execute test, bump version and do a commit with a tag after bumping the version. https://docs.npmjs.com/cli/version

PR #129, opened 2020-10-01
Replace all usage of Tag[] with Tagset - closes Server 127

techsmyth:
Removes the Tag entity, replacing everywhere it was used with a Tagset.
Renamed the previous ITaggable interface to be ITagsetable to reflect that it is about managing multiple tagsets. 
Added new interface class ITaggable to be able to interact with all types that have a single Tagset.

Rename mutation updateTagsOnTagset() to be replaceTagsOnTagset()
Removed mutation for creating a Tag

Updated load-sample-data script to use updated api, and also to create an ecoverse if one does not exist. 

Note: depends on PR related to server-119 as builds upon that code base.


src/schema/inputs/OrganisationInput.ts
```diff
@@ -24,8 +24,8 @@ export class BaseUpdateOrganisationInput {
   @Field(() => [UpdateNestedUserInput], { nullable: true })
   members?: UpdateNestedUserInput[];
 
-  @Field(() => [UpdateNestedTagInput], { nullable: true })
-  tags!: UpdateNestedTagInput[];
+  @Field(() => UpdateNestedTagsetInput, { nullable: true })
+  tagset!: UpdateNestedTagsetInput;
```
atstoyanov, line 28: tagset should be marked as nullable, not as required if the field is set to `nullable: true`

src/schema/inputs/ProjectInput.ts
```diff
@@ -40,8 +40,8 @@ export class UpdateProjectInput {
   @MaxLength(255)
   lifecyclePhase?: string;
 
-  @Field(() => [UpdateNestedTagInput], { nullable: true })
-  tags!: UpdateNestedTagInput[];
+  @Field(() => UpdateNestedTagsetInput, { nullable: true })
+  tagset!: UpdateNestedTagsetInput;
```
atstoyanov, line 44: tagset should be marked as nullable, not as required if the field is set to `nullable: true`

src/schema/inputs/ProjectInput.ts
```diff
@@ -16,8 +16,8 @@ export class ProjectInput {
   @MaxLength(255)
   lifecyclePhase?: string;
 
-  @Field(() => [TagInput], { nullable: true })
-  tags!: TagInput[];
+  @Field(() => TagsetInput, { nullable: true })
+  tagset!: TagsetInput;
```
atstoyanov, line 20: tagset should be marked as nullable, not as required if the field is set to `nullable: true`

atstojanov (approved):

PR #141, opened 2020-10-06
Mutation createTagsetOnProfile to return tagset - closes Coordination 136

techsmyth:
Mutation createTagsetOnProfile to return tagset 
Optimised addUserToGroup somewhat - more needed


src/schema/resolvers/UpdateMutations.ts
```diff
@@ -21,26 +22,38 @@ export class UpdateMutations {
...
     // Have both user + group so do the add
     group.addUserToGroup(user);
+    //const t3 = performance.now()
     await group.save();
 
+    //const delta1 = t0 - t1;
+    //const delta2 = t2 - t1;
+    //const delta3 = t3 - t2;
+    //const msg = `AddUserToGroup deltas: ${delta1}, ${delta2}, ${delta3}`;
+    //console.log(`AddUserToGroup deltas: ${delta1}, ${delta2}, ${delta3}`);
```
atstoyanov, line 56: Please remove commented code that is not needed or leave a comment why the code is left commented.

atstojanov (changes requested):

src/schema/resolvers/UpdateMutations.ts
techsmyth: Commented code is left in there to assist in server-143 issue - there is a performance issue with this mutation

atstojanov (approved):

PR #167, opened 2020-10-09
Muation: addGroupOnEcoverse - closes Server-166

techsmyth:
Migrates the addGroupOnEcoverse mutation to v0.2

### DIFF
--- a/src/domain/ecoverse/ecoverse.resolver.ts
+++ b/src/domain/ecoverse/ecoverse.resolver.ts
@@ -1,5 +1,6 @@
 import { Get, Inject, UseGuards } from '@nestjs/common';
 import { Query, Resolver } from '@nestjs/graphql';
+import { Args, Mutation } from '@nestjs/graphql/dist/decorators';
 import { GqlAuthGuard } from '../../utils/authentication/graphql.guard';
 import { Context } from '../context/context.entity';
 import { IContext } from '../context/context.interface';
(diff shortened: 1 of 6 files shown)


src/domain/ecoverse/ecoverse.resolver.ts
```diff
@@ -49,4 +50,15 @@ export class EcoverseResolver {
   async context(): Promise<IContext> {
     return this.ecoverseService.getContext();
   }
+
+  ///// Mutations /////
```
valentinyanakiev: I'd suggest splitting queries  / mutations not with a comment but queries to be in [entity].resolver.queries and mutations in [entity].resolver.mutations. It will end up being cleaner separation and will help us keep our classes short and neat.

src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -158,4 +158,14 @@ export class EcoverseService {
       throw e;
     }
   }
+
+  ///////////////////////////// Mutations ////////////////////////////////////
```
valentinyanakiev: I'd omit this comment. Although the method is used in a mutation, the service is providing business logic => this method can very well be consumed in a different way, e.g. rest service.

valentinyanakiev (changes requested):
A couple minor changes required - comments inline.

PR #182, opened 2020-10-12
updated data loading to work with current api; fixed a couple of bugs that were affecting it.

techsmyth:
removed async from Use…rGroupService methods that did not need it; 
fixed the returning of members query at the ecoverse level


valentinyanakiev (approved):
👍  Looks good

PR #188, opened 2020-10-12
Added support for memberof field on User

techsmyth:
Added support for runtime created memberof field on User
Uses lazy loading to iterate over the various groups to return the structure for the query


src/domain/user/user.dto.ts
```diff
@@ -1,7 +0,0 @@
-import { Field, InputType } from '@nestjs/graphql';
-
```
valentinyanakiev, line 2: UserDto might still be needed for other mutations.

src/utils/data-management/data-management.module.ts
```diff
@@ -31,6 +32,7 @@ import { DataManagementService } from './data-management.service';
     UserModule,
     UserGroupModule,
     TypeOrmModule.forFeature([Ecoverse]),
+    TypeOrmModule.forFeature([User]),
```
valentinyanakiev: The User Repo is added but not used in the DataManagementService.

valentinyanakiev (approved):
👍

PR #190, opened 2020-10-12
Adds back in the createChallenge mutation togther with associated ChallengeInput type - closes Server 189

techsmyth:
Tested it with the odyssey challenges + it loaded fine.


src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -212,6 +216,28 @@ export class EcoverseService {
     return group;
   }
 
+  async createChallenge(challengeData: ChallengeInput): Promise<IChallenge> {
```
valentinyanakiev, line 219: createChallenge should be in ChallengeService

valentinyanakiev (changes requested):
Please move createChallenge to ChallengeService

src/domain/ecoverse/ecoverse.service.ts
techsmyth: This would create a circular reference in the modules as the challenge also needs to be added into the list of challenges for the ecoverse.

valentinyanakiev: No entity is standalone, but the challenge business logic should be in ChallengeService. Circular modular / service reference can be resolved with forwardRef on module and service level (see latest pushed commits on server-159).

valentinyanakiev: In general every entity can be considered part of the Ecoverse, but that doesn't mean its business logic is part of the EcoverseService logic. It should be in its own service, following [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) principle.

src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -212,6 +216,28 @@ export class EcoverseService {
...
+        // Challenge already exists, just return. Option:merge?
+        return challenge;
+      }
+    }
+
+    // No existing challenge found, create and initialise a new one!
+    const challenge = Challenge.create(challengeData);
+    this.challengeService.initialiseMembers(challenge);
+    ecoverse.challenges.push(challenge);
+    await ecoverse.save();
```
valentinyanakiev: use ecoverse repository instead of active record.

valentinyanakiev (approved):
👍

PR #195, opened 2020-10-12
added back in mutation for creating a new user

techsmyth:

### DIFF
--- a/src/domain/ecoverse/ecoverse.service.ts
+++ b/src/domain/ecoverse/ecoverse.service.ts
@@ -15,10 +15,13 @@ import { TagsetService } from '../tagset/tagset.service';
 import { IUser } from '../user/user.interface';
 import { IChallenge } from '../challenge/challenge.interface';
 import { ITagset } from '../tagset/tagset.interface';
+import { UserInput } from '../user/user.dto';
+import { UserService } from '../user/user.service';
 
 @Injectable()
 export class EcoverseService {
   constructor(
+    private userService: UserService,
     private userGroupService: UserGroupService,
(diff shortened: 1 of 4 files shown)


src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -229,4 +232,21 @@ export class EcoverseService {
 
     return ctVerse;
   }
+
+  async createUser(userData: UserInput): Promise<IUser> {
```
valentinyanakiev, line 236: createUser should be in UserService, not EcoverseService

valentinyanakiev (changes requested):
Please move createUser to UserService.

src/domain/ecoverse/ecoverse.service.ts
techsmyth: this would create a circular reference as the user also needs to be added in to the members group - see pending issue. In general design has that you create the entity on the containing entite e.g. createGroup on org, challenge + ecoverse. A user is not standalone.

valentinyanakiev: No entity is standalone, but the user business logic should be in UserService. Circular modular / service reference can be resolved with forwardRef on module and service level (see latest pushed commits on server-159).

valentinyanakiev: In general every entity can be considered part of the Ecoverse, but that doesn't mean its business logic is part of the EcoverseService logic. It should be in its own service, following [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) principle.

PR #197, opened 2020-10-13
Added back in mutation for creating new user; added user to members group on creation

techsmyth:
Replaces PR #195 as got a conflict with server lock. 

Added in members group additioin in addition to original request.


valentinyanakiev (changes requested):
Use ecoverse repository instead of active record.

valentinyanakiev (approved):
👍

PR #202, opened 2020-10-13
Adds in createOrganisation mutation; adds in mutation samples

techsmyth:
The creation of the org is throwing an error - would like to have that as a separate issue on the board if you are ok with the code setup.

Closes: 192 & 200


valentinyanakiev (approved):
👍

PR #207, opened 2020-10-13
new mutation to assign the focal point for a group; 

techsmyth:
updated focalPoit member to have eager loading and removed from other side of the relation; 
added sample mutation file


valentinyanakiev (approved):
👍

graphql-samples/mutations/update/assign-group-focal-point
```diff
@@ -0,0 +1,13 @@
+mutation assignGroupFocalPoint($userID: Float!, $groupID: Float!) {
+  assignGroupFocalPoint(groupID: $groupID, userID: $userID) {
+    name,
+    id,
+    focalPoint
```
valentinyanakiev, line 5: You need to provide focalPoint fields, otherwise it changes (I ran the sample and it threw an exception)

PR #208, opened 2020-10-14
update user(string) query to also check by email - closes Server 172

techsmyth:
Updates user(string) query to check by email.
Better input argument checking.


valentinyanakiev (approved):
👍

PR #217, opened 2020-10-14
Simplify bootstrapping process; new queries for organisations

techsmyth:
Unfortunately a bigger PR - as had to sort out the bootstrapping to get the organisations working (related to host org + how that was being setup)

### DIFF
--- a/src/domain/ecoverse/ecoverse.service.ts
+++ b/src/domain/ecoverse/ecoverse.service.ts
@@ -103,7 +103,7 @@ export class EcoverseService {
       // this.eventDispatcher.dispatch(events.ecoverse.query, { ecoverse: ecoverse });
       // Convert groups array into IGroups array
       if (!ecoverse.groups) {
-        throw new Error('Unreachable');
+        throw new Error('Ecoverse groups must be defined');
       }
       return ecoverse.groups as IUserGroup[];
     } catch (e) {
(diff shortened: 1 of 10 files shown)


src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -176,6 +128,18 @@ export class EcoverseService {
     }
   }
 
+  async getOrganisations(): Promise<IOrganisation[]> {
+    try {
+      const ecoverse: IEcoverse = await this.getEcoverse();
+
+      if (!ecoverse.organisations) throw new Error('Unreachable');
```
atstoyanov: Why not returning empty array instead of throwing error? 
What does 'Unreachable' mean in this context? It sounds like network error.

src/domain/organisation/organisation.service.ts
```diff
@@ -35,6 +35,16 @@ export class OrganisationService {
     return organisation;
   }
 
+  async getOrganisationByID(organisationID: number): Promise<IOrganisation> {
+    //const t1 = performance.now()
+    const organisation = await Organisation.findOne({
+      where: [{ id: organisationID }],
+    });
+    if (!organisation)
+      throw new Error(`Unable to find challenge with ID: ${organisationID}`);
```
atstoyanov: Should be: "Unable to find **organisation** with ID: ${organisationID}`"

atstojanov (changes requested):

src/domain/ecoverse/ecoverse.service.ts
techsmyth: Organisations are not initialised in the constructor due to restrictions from TypeORM - but are in the intialiseMembers method. So they are guaranteed to be in place.

src/domain/organisation/organisation.service.ts
techsmyth: To be fixed

atstojanov (approved):

PR #220, opened 2020-10-14
New mutation to allow removing of users from the ecoverse - closes Server 137

techsmyth:


valentinyanakiev (approved):
IDs should be GraphQL ID Scalar Type. Non-blocking issue so I am adding it to the list of cleanup items.

PR #221, opened 2020-10-14
Updates to get the ecoverse structure + user loading working

techsmyth:
Fixed various bugs to be able to get the structure + user loading working.

To validate use the odyssey structure + user loading


valentinyanakiev (approved):
👍

PR #225, opened 2020-10-15
Added mutation and sample file for updating a user

techsmyth:
Added mutation and sample file for updating a user


valentinyanakiev (approved):
👍

PR #226, opened 2020-10-15
updated documentation related to populating the sample data in the server

techsmyth:
updated README


valentinyanakiev (changes requested):
1. This should have been part of the previous PR.
2. Documentation is incomplete. There are 3 operations supported with the data management and only one documented.

valentinyanakiev (approved):

PR #229, opened 2020-10-15
Update docker-compose to have db locally available; update sample mutation

techsmyth:
The docker compose script was not exposting the db port locally - which is the handiest way currently to get a db instance running + setup.
Updated the sample file for creating a group on an ecoverse.


valentinyanakiev (approved):
👍

PR #231, opened 2020-10-15
new mutation to add a user to a challenge

techsmyth:
One of the key logical actions is to add a user to a challenge i.e. to add the user to the members group for a challenge
New mutation is there
New sample file created


valentinyanakiev (approved):
👍

PR #236, opened 2020-10-16
Server 162. Closes #162

### LINKED ISSUE #162: Creation of a user in CT triggers also creation of an account in AAD
## Description
As a user of CT I want to have one place where I can create new users for the platform that is consistently represented in the platform services

## Acceptance criteria

- [x] User creation to check for existance of a user with the supplied email address in both CT users table and AAD
- [x] Create new CT user and AAD user if no such user account with the given email address exists
- [x] Create new CT user if the user does not exist in CT but does exist in AAD
- [x] If the user does exist in CT and not in AAD then throw an error
- [x] Return the existing user if the user already exists in both CT + AAD

valentinyanakiev:

Added Microsoft Graph module and logic for AAD user management.

### Additional context
1. In order to call downstream api resource (MS Graph API) we need to replace our Bearer Auth token with another access token using the On-behalf-of flow.
2. In order to obtain the token, client secret is required (passed as env variable **AAD_CLIENT_SECRET** in the application) and also scopes need to be provided - in our case we provide them with **AAD_API_SCOPE** (in .env.default).
3. The **AAD_CLIENT_SECRET** env variable is sensitive and is not added to the .env.default. Contact me (Valentin) to provide you one or create one yourself in AAD.
4. In order to persist the bearer token between requests in-memory cache is used.
5. In order to activate the flow, you need to use the ct admin user (admin@playgroundcherrytwist.onmicrosoft.com) or you need to provide yourself Directory.ReadWrite.All permissions.
6. Isomorphic-fetch is a hard requirement for the ms graph client library.
7. In order to activate OBO flow and use it in ms graph client interface is implemented (the interface is defined in the ms graph library).
8. Some obsolete code was removed.

### Governance 

- [x] Documentation is added


package.json
```diff
@@ -46,6 +52,7 @@
     "typeorm": "^0.2.28"
   },
   "devDependencies": {
+    "@microsoft/microsoft-graph-types": "^1.23.0",
```
techsmyth, line 53: why only a dev dependency?

README.md
```diff
@@ -86,6 +86,8 @@ Define AAD_TENANT, AAD_CLIENT environment variables - e.g. locally in .env envir
 ```bash
 AAD_TENANT=[tenant (directory) ID]
 AAD_CLIENT= [client (application) ID]
+AAD_API_SCOPE= [API Scopes Required for Downstream APIs, in our case Microsoft Graph API]
+AAD_CLIENT_SECRET=Client secret obtained from the AAD App Registration
```
techsmyth: Is this not environment specific i.e. that it is the client registration for the playground environment? Am not too worried about this leaking out as it is only our playground setup.

package.json
atstoyanov: types are development dependency, they are not needed for production. If you review the whole `package.json` you'll see that all types are under devDependecies, except `@types/passport-azure-ad` which have to be moved to devDependecies

package.json
```diff
@@ -24,19 +24,25 @@
...
+    "@microsoft/microsoft-graph-client": "^2.1.0",
     "@nestjs/common": "^7.4.4",
     "@nestjs/config": "^0.5.0",
     "@nestjs/core": "^7.4.4",
     "@nestjs/graphql": "^7.7.0",
     "@nestjs/passport": "^7.1.0",
     "@nestjs/platform-express": "^7.0.0",
     "@nestjs/typeorm": "^7.1.4",
+    "@types/node-fetch": "^2.5.7",
     "@types/passport-azure-ad": "^4.0.7",
```
atstoyanov: Move them into "devDependencies".

valentinyanakiev: Addressed in latest commit.

README.md
valentinyanakiev: Addressed in the latest commit.

src/domain/user/user.service.ts
```diff
@@ -53,6 +55,19 @@ export class UserService {
     return this.userRepository.findOne({ email: email });
   }
 
+  async userExistsAadCt(email: string): Promise<boolean> {
```
techsmyth: Looks like this function is not used - keep it or now? Removing it also removes knowledge of MSGraph from teh User service which is probably correct.

valentinyanakiev: Addressed in commit c590131c0265b53e16dcf1b2160f43fc4613f547

techsmyth (approved):

PR #246, opened 2020-10-19
new mutation to add a tag to a tagset; misc minor updates

techsmyth:
updated version number to 0.2.2; 
fixed profile to be always present on usergroup; 
added sample file for mutation and fixed a typo in another mutation sample file

### DIFF
--- a/src/domain/tagset/tagset.service.ts
+++ b/src/domain/tagset/tagset.service.ts
@@ -1,10 +1,17 @@
 import { Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
 import { ITagsetable } from 'src/interfaces/tagsetable.interface';
+import { Repository } from 'typeorm';
 import { RestrictedTagsetNames, Tagset } from './tagset.entity';
 import { ITagset } from './tagset.interface';
 
 @Injectable()
 export class TagsetService {
+  constructor(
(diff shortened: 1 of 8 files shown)


src/domain/tagset/tagset.service.ts
```diff
@@ -28,7 +35,28 @@ export class TagsetService {
...
+  }
+
+  async addTag(tagsetID: number, newTag: string): Promise<ITagset> {
+    const tagset = (await this.getTagset(tagsetID)) as Tagset;
+
+    if (!tagset) throw new Error(`Tagset with id(${tagsetID}) not found!`);
+    if (!tagset.tags)
+      throw new Error(`Tagset with id(${tagsetID}) not initialised!`);
+
+    // Check if the tag already exists or not
```
valentinyanakiev, line 50: Use [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead of iterating.

valentinyanakiev (changes requested):
One minor change requested.

src/domain/tagset/tagset.service.ts
valentinyanakiev: You are comparing different types, ignore my comment.

valentinyanakiev (approved):

PR #247, opened 2020-10-19
adds query to find groups with a given tag - closes Server 245

techsmyth:
addresses issues with default profile initialisation on usergroups
updates data-mgmt to use await for async initialise calls

### DIFF
--- a/src/domain/tagset/tagset.service.ts
+++ b/src/domain/tagset/tagset.service.ts
@@ -113,4 +113,11 @@ export class TagsetService {
     tagsetable.tagsets?.push(newTagset as ITagset);
     return newTagset;
   }
+
+  hasTag(tagset: ITagset, tagToCheck: string): boolean {
+    for (const tag of tagset.tags) {
+      if (tag === tagToCheck) return true;
+    }
+    return false;
+  }
 }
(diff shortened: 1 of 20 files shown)


src/domain/tagset/tagset.service.ts
```diff
@@ -113,4 +113,11 @@ export class TagsetService {
     tagsetable.tagsets?.push(newTagset as ITagset);
     return newTagset;
   }
+
+  hasTag(tagset: ITagset, tagToCheck: string): boolean {
+    for (const tag of tagset.tags) {
+      if (tag === tagToCheck) return true;
```
valentinyanakiev, line 147: tagset.tags is normal array. Use [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead of iterating over the array.

valentinyanakiev (changes requested):
Only one minor change needed.

src/domain/tagset/tagset.service.ts
valentinyanakiev: Comparing string to ITagset, ignore my comments. It's ok.

valentinyanakiev (approved):

PR #258, opened 2020-10-20
allow user profiles to be created when authenticaton is disabled

techsmyth:
additional logging during bootstrap re auth status; 
removed logging message
new AuthService class

Note: was not able to use the auth service in the guard - seems the config service was not set? Means a bit of duplicate code in GqlGuard.


src/utils/bootstrap/bootstrap.service.ts
```diff
@@ -24,17 +26,29 @@ export class BootstrapService {
       console.info('Bootstrapping Ecoverse...');
       await this.ensureEcoverseSingleton();
       await this.ensureAdminRole();
+      await this.validateAccountMgmtSetup();
     } catch (error) {
       console.log(error);
     }
   }
 
+  async validateAccountMgmtSetup() {
```
valentinyanakiev: validateAccountManagementSetup - avoid acronyms and abbreviations.

src/utils/account/account.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  async accountUsageEnabled(): Promise<boolean> {
+    // Currently the only factor for account usage is authentication enabled or not i.e. accounts are used if auth is enabled
+    return this.authenticationEnabled();
+  }
+
+  async accountExists(email: string): Promise<boolean> {
+    // If authentication is disabled then the logic should continue as if an account is present
+    if (!this.authenticationEnabled()) return true;
+    return await this.msGraphService.userExists(undefined, email);
```
valentinyanakiev, line 31: I doubt this will work. MS Graph Service relies on OBO flow, which relies on an access token.

src/utils/authentication/aad.strategy.ts
```diff
@@ -79,7 +79,7 @@ export class AzureADStrategy
     const upstreamAccessToken = await this.getCachedBearerToken();
     const response = await this.getDownstreamAccessToken(upstreamAccessToken);
     const downstreamAccessToken = response['access_token'] as string;
-    console.info(`Downstream access token: ${downstreamAccessToken}`);
```
valentinyanakiev: If you don't want to see info messages you can control this in the bootstrap method in main.ts. No need to remove them.

src/utils/authentication/graphql.guard.ts
```diff
@@ -14,12 +14,13 @@ export class GqlAuthGuard extends AuthGuard('azure-ad') {
   canActivate(context: ExecutionContext) {
     const ctx = GqlExecutionContext.create(context);
     const { req } = ctx.getContext();
-    console.log(req);
```
valentinyanakiev, line 17: Alter it to console.info, do not remove it.

valentinyanakiev (approved):
👍

PR #304, opened 2020-10-25
Added in Opportunity entity, 

techsmyth:
Also moved Projects to be under Opportunity; 
renamed lifecyclePhase to be state at all levels (Challenge, Opportunity, Project) & updates sample files
Added graphql sample files


valentinyanakiev (approved):
👍

PR #323, opened 2020-10-27
fixed boostrapping of profiles; moved creating of users to ecoverse fully

techsmyth:
Moved the createUserAccount and createUserProfile to be on ecoverse
updated createUserProfile to also add the user to the members group

Removed MSGraph dependency from User module

Tidied up methods on bootstrap to remove flags / parameters not needed and to only call createUserProfile

### DIFF
--- a/src/domain/ecoverse/ecoverse.service.ts
+++ b/src/domain/ecoverse/ecoverse.service.ts
@@ -292,30 +292,53 @@ export class EcoverseService {
 
   // Create the user and add the user into the members group
   async createUser(userData: UserInput): Promise<IUser> {
+    const user = await this.createUserProfile(userData);
+    if (this.accountService.authenticationEnabled()) {
+      await this.createUserAccount(user.id);
+    }
+    return user;
+  }
+
+  // Create the user and add the user into the members group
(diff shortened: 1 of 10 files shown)


src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -292,31 +292,54 @@ export class EcoverseService {
...
   async createUser(userData: UserInput): Promise<IUser> {
+    const user = await this.createUserProfile(userData);
+    if (this.accountService.authenticationEnabled()) {
+      await this.createUserAccount(user.id);
+    }
+    return user;
+  }
+
+  // Create the user and add the user into the members group
+  async createUserAccount(userID: number): Promise<boolean> {
```
valentinyanakiev: I don't believe this is Ecoverse-specific logic and I don't think this should be in the Ecoverse service.

techsmyth: Creating of a user is an ecoverse action, as it is admitting a user in - creating an account could be on a resolver on the account module?

valentinyanakiev (approved):

PR #324, opened 2020-10-28
added avatar field on Profile

techsmyth:
new avatar field on profile which is a uri
added sample graphql file


valentinyanakiev (approved):
👍

PR #327, opened 2020-10-28
Added support for accountUpn that is independent of email

techsmyth:
adds accountUpn field to the User entity
update the account creation logic to generate a UPN based on firstname.lastname@upn_domain
added an enviornment variable that specifies the upndomain to be used
updated readme to detail the new environment variable
updated version number to 0.3.2


atstoyanov:
Does it mean that the 'First' and 'Last' name fields must be mandatory. And what will do when there are two or more 'John Snow'?

techsmyth:
> Does it mean that the 'First' and 'Last' name fields must be mandatory. And what will do when there are two or more 'John Snow'?

Good point re first + last name being mandatory. 
Exception case when two people end up with the same upn but different emails does need to be dealt with.

valentinyanakiev (commented):
I am not sure whether the proposed changes are an improvement on the current logic or not. Removing the building of UPN could lead to error-prone configurations. We don't have an excessive explanation of how and where to get the UPN from and it is less than obvious. Also these changes mandate using of native AAD users as the whole flow is changed to revolve around UPN. The leading field for invited users is email, not UPN. This needs broader discussion.

valentinyanakiev (approved):
👍 

Only caveat is given / surname upn building.

PR #344, opened 2020-10-31
createUserAccount mutation bug fixed; ensure generated upn does not have spaces

techsmyth:

### DIFF
--- a/src/utils/account/account.service.ts
+++ b/src/utils/account/account.service.ts
@@ -53,7 +53,7 @@ export class AccountService {
     } catch (e) {
       const msg = `Unable to complete account creation for ${userData.email}: ${e}`;
       this.logger.error(msg);
-      throw new Error(msg);
+      throw e;
     }
     // Update the user to store the upn
     const user = await this.userService.getUserByEmail(userData.email);
@@ -87,6 +87,7 @@ export class AccountService {
     if (!upnDomain)
       throw new Error('Unable to identify the upn domain to be used');
 
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
     const normalizer = require('normalizer');
 
     // Note: requesting client has the option to explicilty specify the account UPN to use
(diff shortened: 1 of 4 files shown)


src/utils/account/account.service.ts
```diff
@@ -44,12 +44,17 @@ export class AccountService {
...
-      );
+    try {
+      const result = await this.msGraphService.createUser(userData, accountUpn);
+      if (!result)
+        throw new Error(
+          `Unable to complete account creation for ${userData.email} using UPN: ${accountUpn}`
+        );
+    } catch (e) {
+      const msg = `Unable to complete account creation for ${userData.email}: ${e}`;
+      this.logger.error(msg);
```
valentinyanakiev, line 55: If you throw an Error it will be logged, no need to log it separately.

src/utils/account/account.service.ts
```diff
@@ -81,6 +87,8 @@ export class AccountService {
     if (!upnDomain)
       throw new Error('Unable to identify the upn domain to be used');
 
+    const normalizer = require('normalizer');
```
valentinyanakiev, line 91: Require is not recommended in TS code. You either need to suppress the warning (not recommended) or import the type. I couldn't find @types/normalize so we either need to write our own type definition or suppress the eslint warning here (again, not recommended, but it will be faster).

valentinyanakiev (changes requested):
A couple of minor points need reviewing.

src/utils/account/account.service.ts
techsmyth: ok, will fix

src/utils/account/account.service.ts
techsmyth: for now we just supress the eslint warning. Need to focus elsewhere.

src/utils/account/account.service.ts
techsmyth: Wanted to add an additional error msg that account creation failed, calling an external service so want to catch and log here and then throw further. Updated to rethrow and not throw a new msg.

valentinyanakiev (approved):

PR #349, opened 2020-11-01
Added ActorGroup & Actor entities

techsmyth:
New entity: ActorGroup
New entity: Actor
New mutation: createActor, together with a sample mutation file
Challenge entity updated to hold a list of actor groups
Support for default actor groups on Challenge

Notes:
- tests still to be added, will need some pointers / help on that
- additional mutations will be in a separate issue

### DIFF
--- a/src/domain/actor-group/actor-group.entity.ts
+++ b/src/domain/actor-group/actor-group.entity.ts
@@ -1,0 +1,59 @@
+import { Field, ID, ObjectType } from '@nestjs/graphql';
+
+import {
+  BaseEntity,
+  Column,
+  Entity,
+  ManyToOne,
+  OneToMany,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { Actor } from '../actor/actor.entity';
+import { Opportunity } from '../opportunity/opportunity.entity';
+import { IActorGroup } from './actor-group.interface';
+
+export enum RestrictedActorGroupNames {
+  Collaborators = 'collaborators',
+}
+
+@Entity()
+@ObjectType()
+export class ActorGroup extends BaseEntity implements IActorGroup {
+  @Field(() => ID)
+  @PrimaryGeneratedColumn()
+  id!: number;
+
+  @Field(() => String)
+  @Column('varchar', { length: 100 })
+  name: string;
+
+  @Field(() => String, {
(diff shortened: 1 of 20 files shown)


src/domain/actor-group/actor-group.entity.ts
```diff
@@ -0,0 +1,59 @@
...
+  opportunity?: Opportunity;
+
+  @Field(() => [Actor], {
+    nullable: true,
+    description: 'The set of actors in this actor group',
+  })
+  @OneToMany(
+    () => Actor,
+    actor => actor.actorGroup,
+    { eager: true, cascade: true }
```
valentinyanakiev, line 50: Was this eager loading intentional? It will be slow.

src/domain/opportunity/opportunity.entity.ts
```diff
@@ -69,6 +73,18 @@ export class Opportunity extends BaseEntity
...
 
+  @Field(() => [ActorGroup], {
+    nullable: true,
+    description:
+      'The set of actor groups within the context of this Opportunity',
+  })
+  @OneToMany(
+    () => ActorGroup,
+    actorGroup => actorGroup.opportunity,
+    { eager: true, cascade: true }
```
valentinyanakiev, line 86: Again, was this eager loading intentional? There will be plenty of opportunities and plenty of actorGroups + actors. This will have similar impact on performance as user-userGroup eager loadings.

valentinyanakiev (commented):
Please view my comments about eager loading - they will have performance implications.

src/domain/actor-group/actor-group.entity.ts
techsmyth: Yes intentional, as this data set I do not expect to grow substantially. If it is a problem then we optimise, but first want to have the funcationality in + working.

src/domain/opportunity/opportunity.entity.ts
techsmyth: yes plenty of them, but all narrowly scoped e.g. 2 or 3 per challenge, and then maybe 3 actor groups per opportunity. First operating and can then see where to optimise.

valentinyanakiev (approved):
👍

PR #351, opened 2020-11-01
Added Aspect entity and usage of aspects by Opportunity

techsmyth:
New entity: Aspect
Opportunity updated to have an array of aspects
new mutation for adding an aspect to an Opportunity

### DIFF
--- a/src/domain/opportunity/opportunity.entity.ts
+++ b/src/domain/opportunity/opportunity.entity.ts
@@ -12,6 +12,7 @@ import {
   PrimaryGeneratedColumn,
 } from 'typeorm';
 import { IGroupable } from '../../interfaces/groupable.interface';
+import { Aspect } from '../aspect/aspect.entity';
 import { Challenge } from '../challenge/challenge.entity';
 import { DID } from '../did/did.entity';
 import { Profile } from '../profile/profile.entity';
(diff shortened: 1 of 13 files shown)


src/domain/opportunity/opportunity.entity.ts
```diff
@@ -69,6 +70,17 @@ export class Opportunity extends BaseEntity
...
   projects?: Project[];
 
+  @Field(() => [Aspect], {
+    nullable: true,
+    description: 'The set of solution aspects for this Opportunity',
+  })
+  @OneToMany(
+    () => Aspect,
+    aspect => aspect.opportunity,
+    { eager: true, cascade: true }
```
valentinyanakiev, line 80: Same comment as other PRs - do we want lazy loading?

valentinyanakiev (commented):
Overall OK, the only comment is about eager loading.

src/domain/opportunity/opportunity.entity.ts
techsmyth: same comment as on actor groups / actor - lets first get it working and then can see re not using eager loading.

valentinyanakiev (approved):

techsmyth:
Lets go with the eager loading for now, if it does have an impact then will optimise. For now key to get the domain model extended so we can discuss its usage with Momentum.

PR #352, opened 2020-11-01
Added support for templates, with each template for now only containing user definitions

techsmyth:
new entity: Template
new query: templates
new mutation: createTemplate
new sample file for creating a new template

The population of a template to be another issue.

### DIFF
--- a/src/domain/ecoverse/ecoverse.service.ts
+++ b/src/domain/ecoverse/ecoverse.service.ts
@@ -45,6 +45,7 @@ export class EcoverseService {
     private ecoverseRepository: Repository<Ecoverse>,
     @Inject(WINSTON_MODULE_NEST_PROVIDER) private readonly logger: Logger
   ) {}
+
   // Helper method to ensure all members that are arrays are initialised properly.
   // Note: has to be a seprate call due to restrictions from ORM.
   async initialiseMembers(ecoverse: IEcoverse): Promise<IEcoverse> {
@@ -86,15 +87,8 @@ export class EcoverseService {
   }
 
   async getName(): Promise<string> {
-    try {
-      const ecoverse = await this.getEcoverse();
-      // this.eventDispatcher.dispatch(events.ecoverse.query, { ecoverse: ecoverse });
-
-      return ecoverse.name;
-    } catch (e) {
-      // this.eventDispatcher.dispatch(events.logger.error, { message: 'Something went wrong in getName()!!!', exception: e });
-      throw e;
-    }
+    const ecoverse = await this.getEcoverse();
+    return ecoverse.name;
   }
 
   async getEcoverseId(): Promise<number> {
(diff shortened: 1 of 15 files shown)


src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -159,6 +163,17 @@ export class EcoverseService {
     }
   }
 
+  async getTemplates(): Promise<ITemplate[]> {
+    try {
+      const ecoverseId = await this.getEcoverseId();
+      const templates = await this.templateService.getTemplates(ecoverseId);
+      return templates;
+    } catch (e) {
+      // this.eventDispatcher.dispatch(events.logger.error, { message: 'Something went wrong in getMembers()!!!', exception: e });
```
valentinyanakiev: Delete this line.

src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -159,6 +163,17 @@ export class EcoverseService {
...
+      throw e;
```
valentinyanakiev: No point to rethrow, with the entry above commented / removed and no context added to the Error object, this is pointless. Unhandled exceptions are handled by Nest anyway.

src/domain/template/template.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  @Field(() => [User], {
+    nullable: true,
+    description:
+      'The set of user types that are available within this template',
+  })
+  @ManyToMany(
+    () => User,
+    user => user.templates,
+    { eager: true, cascade: true }
```
valentinyanakiev, line 38: Many-to-many and eager loading will be a performance disaster most likely.

src/domain/template/template.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+  )
+  @JoinTable({ name: 'user_group_members' })
```
valentinyanakiev: You can't use the same join table as the one defining the user-group relationship, this needs to be a separate join table for the many-to-many.

valentinyanakiev (changes requested):
Several changes required. Also, have a look at the build https://travis-ci.com/github/cherrytwist/Server/builds/197020078. The changes to the model require changes to the tests, otherwise, the whole pipeline will start failing.

src/domain/template/template.entity.ts
techsmyth: Only going to be one user near term, so will not have an impact imho...

src/domain/template/template.entity.ts
techsmyth: good catch - missed this one. Will fix...

src/domain/template/template.entity.ts
techsmyth: First lets get it working, will be very little in the template for now

src/domain/ecoverse/ecoverse.service.ts
techsmyth: Agreed, removed all the catching / rethrowing which was from a while back

techsmyth:
Think this is good to go - I cannot reproduce the test results locally, will see what happens with the new commit...

valentinyanakiev (approved):

PR #367, opened 2020-11-03
removed eager loading on groups from challenges

techsmyth:
To start dealing with the findings of the load testing results from Evgeni.

### DIFF
--- a/src/domain/challenge/challenge.service.ts
+++ b/src/domain/challenge/challenge.service.ts
@@ -68,7 +68,11 @@ export class ChallengeService {
     if (!challengeID)
       throw new Error(`Invalid challenge id passed in: ${challengeID}`);
     // Try to find the challenge
-    const challenge = await Challenge.findOne(challengeID);
+    const challenge = await this.challengeRepository.findOne({
+      where: { id: challengeID },
+      relations: ['groups'],
+    });
+    //const challenge = await Challenge.findOne(challengeID);
     if (!challenge) {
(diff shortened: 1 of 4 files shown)


src/domain/challenge/challenge.service.ts
```diff
@@ -68,7 +68,11 @@ export class ChallengeService {
...
       throw new Error(`Invalid challenge id passed in: ${challengeID}`);
     // Try to find the challenge
-    const challenge = await Challenge.findOne(challengeID);
+    const challenge = await this.challengeRepository.findOne({
+      where: { id: challengeID },
+      relations: ['groups'],
+    });
+    //const challenge = await Challenge.findOne(challengeID);
     if (!challenge) {
       const msg = `Unable to find challenge with ID: ${challengeID}`;
```
valentinyanakiev, line 73: Better to throw an error here, assuming you expect to have this information/message on the client.

src/domain/challenge/challenge.service.ts
```diff
@@ -78,11 +82,17 @@ export class ChallengeService {
       challenge,
       groupName
     );
-    await challenge.save();
+    await await this.challengeRepository.save(challenge);
```
valentinyanakiev: Await await?

valentinyanakiev (changes requested):
A couple of minor changes required.

valentinyanakiev (approved):

PR #371, opened 2020-11-03
recursive updates and creation on users; 

techsmyth:
removed usage of TagsInput which was a workaround


src/domain/tagset/tagset.resolver.ts
```diff
@@ -17,35 +15,35 @@ export class TagsetResolver {
     RestrictedGroupNames.EcoverseAdmins
   )
   @UseGuards(GqlAuthGuard)
-  @Mutation(() => Tagset, {
+  @Mutation(() => Boolean, {
```
atstoyanov, line 19: The result of a mutation must be an object that has a primary key. It is crucial to have the Identifier in the result, because the ApolloClient cache is using it to updates its internal cache.

src/domain/tagset/tagset.resolver.ts
```diff
@@ -17,35 +15,35 @@ export class TagsetResolver {
...
+    return true;
   }
 
   @Roles(
     RestrictedGroupNames.CommunityAdmins,
     RestrictedGroupNames.EcoverseAdmins
   )
   @UseGuards(GqlAuthGuard)
-  @Mutation(() => Tagset, {
+  @Mutation(() => Boolean, {
```
atstoyanov, line 37: The result of a mutation must be an object that has a primary key. It is crucial to have the Identifier in the result, because the ApolloClient cache is using it to updates its internal cache.

atstojanov (commented):
Mutations must return objects not scalar values.

src/domain/tagset/tagset.resolver.ts
techsmyth: ah, was not aware of this...

atstojanov (approved):

PR #392, opened 2020-11-04
memberof 

techsmyth:
return empty groups
only return groups on the challenge / organisaiton level that a user is a member of


valentinyanakiev (approved):
👍

PR #404, opened 2020-11-05
remove created user if account creation fails

techsmyth:
## Not yet tested as cannot go through the full creation loop


valentinyanakiev (approved):
👍

PR #409, opened 2020-11-06
put awaits in that were missing for async functions

techsmyth:
There are too many errors showing up in the perf tests, so have gone through code to verify that all async funcitons are being awaited


valentinyanakiev (approved):
👍

PR #413, opened 2020-11-06
throw an error if try to create a challenge with the same name as one already existing

techsmyth:


src/domain/ecoverse/ecoverse.service.ts
```diff
@@ -225,17 +225,18 @@ export class EcoverseService {
...
+    if (challenge) {
+      // already have a challenge with the given name, not allowed
+      throw new Error(
+        `Unable to create challenge: already have a challenge with the provided name (${challengeData.name})`
+      );
     }
+    // No existing challenge found, create and initialise a new one!
+    challenge = await this.challengeService.createChallenge(challengeData);
+
+    ecoverse.challenges.push(challenge as Challenge);
```
valentinyanakiev: That's an anti-pattern, never cast an interface to its implementation like that.

valentinyanakiev (changes requested):
Remove the cast from an interface to its implementation.

valentinyanakiev (approved):
👍

PR #419, opened 2020-11-06
Added query usersById

atstojanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/domain/IDList.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, InputType } from '@nestjs/graphql';
+
+@InputType()
+export class IdList {
+  @Field(() => [String], {
```
techsmyth: Why a separate class rather than a straight array of strings like on the mutation replaceTagsOnTagset?

atstoyanov: Because I wasn't aware how to do it this way :) Thanks for pointing to that solution.

src/domain/ecoverse/ecoverse.resolver.queries.ts
```diff
@@ -89,6 +89,25 @@ export class EcoverseResolverQueries {
...
 
+  @Roles(
+    RestrictedGroupNames.CommunityAdmins,
+    RestrictedGroupNames.EcoverseAdmins
+  )
+  @UseGuards(GqlAuthGuard)
+  //should be in user queries
+  @Query(() => [User], {
+    nullable: false,
+    description: 'A particular user, identified by the ID or by email',
```
valentinyanakiev: Change the description - this is misleading and is the actual API schema documentation that is visible to gql consumers.

valentinyanakiev (changes requested):
Fix the query description.

Enderedne (approved):

PR #423, opened 2020-11-07
Added context to logger calls

techsmyth:

### DIFF
--- a/src/utils/data-management/data-management.service.ts
+++ b/src/utils/data-management/data-management.service.ts
@@ -16,6 +16,7 @@ import { UserService } from '../../domain/user/user.service';
 import { Connection, Repository } from 'typeorm';
 import { BootstrapService } from '../bootstrap/bootstrap.service';
 import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import { LogContexts } from '../logging/logging-framework';
 
 @Injectable()
 export class DataManagementService {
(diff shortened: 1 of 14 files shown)


src/utils/logging/logging-framework.ts
```diff
@@ -0,0 +1,8 @@
+export enum LogContexts {
```
valentinyanakiev, line 1: Move this enums in the winston.config.ts - this is just a config utility.

src/utils/data-management/data-management.service.ts
```diff
@@ -240,7 +241,7 @@ export class DataManagementService {
       await this.connection.synchronize();
       this.addLogMsg(msgs, '.....dropped. Completed successfully.');
     } catch (error) {
-      this.logger.verbose(error.message);
+      this.logger.verbose(error.message, LogContexts.DATA_MGMT);
```
valentinyanakiev, line 244: Check error messages - I think there are 3 parameters for this.logger.error and the second one is not the context.

valentinyanakiev (changes requested):
Move the enum to winston.config.ts and verify this.logger.error(message, context) - I think it is this.logger.error(message,undefined,context)

src/utils/data-management/data-management.service.ts
techsmyth: yes but this is a verbose message

valentinyanakiev (approved):
👍

PR #427, opened 2020-11-07
added decorator for tracking api usage + performance

techsmyth:


src/utils/config/logging.config.ts
```diff
@@ -0,0 +1,18 @@
+import { registerAs } from '@nestjs/config/dist/utils/register-as.util';
+
+export default registerAs('logging', () => ({
+  loggingLevel:
+    process.env.LOGGING_LEVEL?.toLowerCase() ||
+    LOGGING_LEVEL.Error.toString().toLowerCase(),
+  profilingEnabled: process.env.LOGGING_PROFILING_ENABLED === 'true',
```
valentinyanakiev: It might be nice to have this .toLowerCase() in case somebody writes True.

valentinyanakiev (changes requested):
Only one small change required - true vs True - use lowercase.

valentinyanakiev (approved):

PR #430, opened 2020-11-08
Initial implementation of search on the ecoverse

techsmyth:
New query: search
New types for input parameters for search and results from search

Search parameters:
- terms, used to search in User.firstName and Group.name for now
- restricting which entity types are searched, 
- restricting which challenges to search (not implemented)
- expanding search to incude tagset names (not implemented)

Results contain:
- a score 
- the located entity, either a user or a group

Initial implementation but within a framework that is targetted at being enough to support a more sophisticated implementation.


src/utils/search/search.resolver.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Profiling } from '../logging/logging.profiling.decorator';
+import { SearchInput } from './search-input.dto';
+import { SearchResult } from './search-result.dto';
+import { RestrictedGroupNames } from '../../domain/user-group/user-group.entity';
+
+@Resolver()
+export class SearchResolver {
+  constructor(private searchService: SearchService) {}
+
+  @Roles(RestrictedGroupNames.Members)
```
valentinyanakiev, line 16: Currently, we don't have a way to set a 'minimum' role so you need to add community-admins and ecoverse-admins here as well.

valentinyanakiev (approved):
👍

PR #432, opened 2020-11-08
Logic to prevent users from being directly removed from members group on ecoverse

techsmyth:
Allowing this could leave the ecoverse in an inconsistent state


atstoyanov:
this PR contains unmerged data for Search functionality.

valentinyanakiev (approved):
👍

PR #440, opened 2020-11-09
added scoring to search results; 

techsmyth:
ensure only one instance of each user and group is returned
if a user / group has multiple matches then the score of that result is incremented


valentinyanakiev (approved):
👍

PR #445, opened 2020-11-09
Relation entity and associated mutations + sample file

techsmyth:

### DIFF
--- a/src/domain/opportunity/opportunity.entity.ts
+++ b/src/domain/opportunity/opportunity.entity.ts
@@ -119,7 +119,6 @@ export class Opportunity extends BaseEntity
   )
   groups?: UserGroup[];
 
-
   @OneToOne(() => DID, { eager: true, cascade: true })
   @JoinColumn()
   DID!: DID;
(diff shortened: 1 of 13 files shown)


src/domain/opportunity/opportunity.entity.ts
```diff
@@ -100,6 +101,17 @@ export class Opportunity extends BaseEntity
...
   aspects?: Aspect[];
 
+  @Field(() => [Relation], {
+    nullable: true,
+    description: 'The set of relations for this Opportunity',
+  })
+  @OneToMany(
+    () => Relation,
+    relation => relation.opportunity,
+    { eager: true, cascade: true }
```
valentinyanakiev, line 111: Again eager in one-to-many?

valentinyanakiev (commented):
The graph of dependencies on opportunities was long and with the new eager-loaded one-to-many will be even longer - this will be a performance hit.

src/domain/opportunity/opportunity.entity.ts
techsmyth: Yes but this is at the end of a containment tree i.e. relations are simple entities that are contained by opportunities - so yes think this is the best option...

techsmyth: and as before, first lets see if this actually causes performance issues and then see where to break the dependency chain for eager loading...

valentinyanakiev (approved):
👍

src/domain/opportunity/opportunity.entity.ts
valentinyanakiev: I am approving it but even if it is at the end of a long-ish tree, all eager relations are loaded and it will impact all the items up to the top.

PR #455, opened 2020-11-09
added mutations around creating / updating / removing projects; 

techsmyth:
query for getting a single project by ID; 
aspects added to a project
sample files for the mutations

### DIFF
--- a/src/domain/opportunity/opportunity.service.ts
+++ b/src/domain/opportunity/opportunity.service.ts
@@ -11,6 +11,9 @@ import { AspectInput } from '../aspect/aspect.dto';
 import { IAspect } from '../aspect/aspect.interface';
 import { AspectService } from '../aspect/aspect.service';
 import { ProfileService } from '../profile/profile.service';
+import { ProjectInput } from '../project/project.dto';
+import { IProject } from '../project/project.interface';
+import { ProjectService } from '../project/project.service';
 import { RelationInput } from '../relation/relation.dto';
 import { IRelation } from '../relation/relation.interface';
 import { RelationService } from '../relation/relation.service';
@@ -30,6 +33,7 @@ export class OpportunityService {
     private userService: UserService,
     private aspectService: AspectService,
     private profileService: ProfileService,
+    private projectService: ProjectService,
     private relationService: RelationService,
     @InjectRepository(Opportunity)
     private opportunityRepository: Repository<Opportunity>,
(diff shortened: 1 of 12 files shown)


src/domain/project/project.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { InputType, Field } from '@nestjs/graphql';
+import { MaxLength } from 'class-validator';
+
+@InputType()
+export class ProjectInput {
```
Enderedne, line 5: Not sure it's supposed to be here but we're missing the **textId** and **shortName**? Also are we going to introduce duration for a project in this version?

src/domain/opportunity/opportunity.service.ts
```diff
@@ -224,6 +228,32 @@ export class OpportunityService {
...
+    opportunityId: number,
+    projectData: ProjectInput
+  ): Promise<IProject> {
+    const opportunity = await this.getOpportunityByID(opportunityId);
+    if (!opportunity)
+      throw new Error(`Unalbe to locate opportunity with id: ${opportunityId}`);
+
+    // Check that do not already have an Project with the same name
+    const title = projectData.name;
+    const existingProject = opportunity.projects?.find(
```
Enderedne, line 248: As far as the mockups go projects have Short Name (2 words max)? Should there be validation for its uniqueness as well?

src/domain/project/project.resolver.ts
```diff
@@ -1,4 +1,81 @@
...
+    RestrictedGroupNames.CommunityAdmins,
+    RestrictedGroupNames.EcoverseAdmins
+  )
+  @UseGuards(GqlAuthGuard)
+  @Query(() => Project, {
+    nullable: false,
+    description: 'A particular Project, identified by the ID',
+  })
+  @Profiling.api
+  async opportunity(@Args('ID') id: number): Promise<IProject> {
```
Enderedne: the method is named opportunity. is that going to impact the endpoint?

src/domain/project/project.interface.ts
```diff
@@ -1,4 +1,5 @@
 import { IAgreement } from '../agreement/agreement.interface';
+import { IAspect } from '../aspect/aspect.interface';
 import { ITagset } from '../tagset/tagset.interface';
 
 export interface IProject {
```
Enderedne, line 5: This might be the place where **textId** is missing. Will we be able to query groups and users for a project?

Enderedne (changes requested):

src/domain/opportunity/opportunity.service.ts
techsmyth: Yes, will raise a separate issue

src/domain/project/project.resolver.ts
techsmyth: Fixed...

Enderedne (approved):

PR #461, opened 2020-11-10
Update 25-server-deployment.yaml

GhostOnTheFiber:

Update the server vars so that they are in line with the config map


valentinyanakiev (approved):
👍

PR #463, opened 2020-11-10
updated authorisation bootstrap to also be able to pass in users first + last names so that UPNs are created correctly

techsmyth:


antst (approved):
Seems to be perfectly correct

Enderedne (approved):

PR #485, opened 2020-11-12
trimmed all incoming user creation key fields…

techsmyth:
also trim their usage in call to msgraph to avoid email addresses already in with spaces


valentinyanakiev (approved):
👍

PR #497, opened 2020-11-14
Remove user account support; new mutation for removing challenge lead; additional checks on incming reqs

techsmyth:


valentinyanakiev (approved):
👍

PR #498, opened 2020-11-15
added projects, opportunities queries; fixed check for challenge name duplicates

techsmyth:
Also incrased version number


valentinyanakiev (approved):
👍

PR #503, opened 2020-11-16
updated field lengths on aspect; version number

techsmyth:


valentinyanakiev (commented):
Changes are fine but without migrations we can't apply this to production data. TypeORM does drop & create when a field size is changed.

Comoque1:
The fields are dropped and recreated, which leads to data lose
image

image

image

valentinyanakiev (approved):

PR #530, opened 2020-11-18
new mutation to add a reference to a context; updated version number;…

techsmyth:
updated version number;
 avoided creating ref on profile if already one with that name


valentinyanakiev (approved):
👍

PR #552, opened 2020-11-26
moved addUserToOpportunity to Challenge level and extended business logic

techsmyth:
An error is thrown if the user is not already a challenge member

Note:
- there is clearly room to optimise this particular mutation, but it is one that should not be happening very often.


valentinyanakiev (approved):
:+1:

PR #558, opened 2020-11-27
remove duplicates from the list of matched terms per result entry

techsmyth:


src/utils/search/search.service.ts
```diff
@@ -110,10 +110,17 @@ export class SearchService {
     let results: ISearchResultEntry[] = [];
     results = await this.buildSearchResults(userResults);
     results.push(...(await this.buildSearchResults(groupResults)));
-
+    this.ensureUniqueTermsPerResult(results);
     return results;
   }
 
+  ensureUniqueTermsPerResult(results: ISearchResultEntry[]) {
```
valentinyanakiev, line 117: This might not work if there are many results. array.forEach is parallel in JS and if you assign a value to a variable this might be a problem.

valentinyanakiev (changes requested):
Change forEach with for. It won't be a big performance hit and the risk is not worth it.

OlegSubik (approved):

valentinyanakiev (approved):
:+1:

PR #584, opened 2020-12-02
Server 513. Closes #513.

### LINKED ISSUE #513: Throw specific error for Token expired
## Description
As a client I need to know when a token  is expired

## Acceptance criteria

- [x] New error type for Auth Token expired
- [x] Log a warning if the env var AAD_NO_PII is set to true as this is blocked by a M$ issue and it is always set to false

valentinyanakiev:

Returns TokenException with CherrytwistErrorStatus.TOKEN_EXPIRED when JWT token has expired.

### Additional context

Works only with AUTH_AAD_LOGGING_PII=true. Microsoft has decided that verbose validation logging needs PII. Go figure.
The culprit is on line 330 in the passport-azure-ad/lib/bearerstrategy.js (you can find it in the node_modules after npm install).

    if (err) {
      if (err.message && !self._options.loggingNoPII)
        return self.failWithLog(err.message);
      else
        return self.failWithLog('In Strategy.prototype.jwtVerify: cannot verify token');
    }

There is a bunch of code similar to this one in the library.

### Governance 

### DIFF
--- a/src/utils/bootstrap/bootstrap.service.ts
+++ b/src/utils/bootstrap/bootstrap.service.ts
@@ -84,6 +84,15 @@ export class BootstrapService {
         );
       const entries = Object.entries(config);
       for (const [key, value] of entries) {
+        if (
+          key === 'loggingNoPII' &&
+          value === false &&
+          process.env.AUTH_AAD_LOGGING_PII === 'false'
+        )
+          this.logger.warn(
+            'Overriding AUTH_AAD_LOGGING_PII to true due to Microsoft passportJs bearer strategy bug https://github.com/AzureAD/passport-azure-ad/issues/521.',
+            LogContext.BOOTSTRAP
+          );
         this.logConfigLevel(key, value, '');
       }
     }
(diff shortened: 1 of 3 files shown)


techsmyth (approved):
Well found...

src/utils/bootstrap/bootstrap.service.ts
```diff
@@ -84,6 +84,15 @@ export class BootstrapService {
         );
       const entries = Object.entries(config);
       for (const [key, value] of entries) {
+        if (
+          key === 'loggingNoPII' &&
+          value === false &&
+          process.env.AUTH_AAD_LOGGING_PII === 'false'
```
techsmyth, line 90: is it not that the case we are overriding is if the env var is set to true? Also looks like code would still set the value...

valentinyanakiev: The env var is true when we are logging PII, and the key is true when no PII is logged. It's inversed logic.

valentinyanakiev: loggingNoPII has the opposite value of  process.env.AUTH_AAD_LOGGING_PII - that's what we check. Whether the hardcoded value is different than the env variable.

PR #635, opened 2020-12-17
Added checks whether challenge with same textID and opportunity with same name exist upon creation.

valentinyanakiev:
Added checks whether challenge with same textID and opportunity with same name exist upon creation.

**Note: Validation logic can be extracted to a separate service / library. **


src/domain/opportunity/opportunity.service.ts
```diff
@@ -235,14 +249,16 @@ export class OpportunityService {
         `Required field textID provided not in the correct format: ${textID}`,
         LogContext.CHALLENGES
       ); // Ensure field is lower case
-    opportunityData.textID = textID.toLowerCase();
 
-    // reate and initialise a new Opportunity using the first returned array item
-    const opportunity = Opportunity.create(opportunityData);
-    await this.initialiseMembers(opportunity);
-    await this.opportunityRepository.save(opportunity);
+    const opportunity = await this.opportunityRepository.find({
```
techsmyth: It is possible to have opportunities with the same name within different challenges i.e. the name uniqueness applies within the context of the containing challenge

techsmyth (approved):

PR #639, opened 2020-12-18
Reset Ecoverse Data Management call using migrations.

valentinyanakiev:


techsmyth (commented):
Running reset-ecoverse multiple times causes issues
- first time get no response

techsmyth (approved):

PR #645, opened 2020-12-22
Fix issue where members can't see opportunities they created

valentinyanakiev:


Comoque1:
In the [attached file](https://github.com/cherrytwist/Server/files/5730206/Privileges.xlsx), with "forbidden" is flagged all entities and sub-entities, to which member user doesn't have rights to access.




@valentinyanakiev, please review and we can discuss it later.
Search is throwing error for member user:
image

Comoque1:
The error message for member user when trying to access forbidden resources must be fixed.
The current one is: `User 'ecoverse-member@cherrytwist.com' doesn't have any roles in this ecoverse.`

Comoque1:
There are 2 failing tests:
1. Member user must not be able to update profile, but currently is able
2. Member user must be able to create relations, but currently is not

image

Comoque1 (approved):

PR #652, opened 2020-12-29
BCH Refactoring - Write Short Units of Code

valentinyanakiev:
Draft PR to measure against BCH compliance, DO NOT MERGE


src/utils/search/search.service.ts
```diff
@@ -9,6 +9,7 @@ import { SearchResultEntry } from './search-result-entry.dto';
 import { ISearchResultEntry } from './search-result-entry.interface';
 import { LogContext } from '../logging/logging.contexts';
 import { ValidationException } from '../error-handling/exceptions/validation.exception';
+import { group } from 'console';
```
techsmyth: wrong import?

techsmyth (approved):
Looks good - one small update.

src/utils/search/search.service.ts
valentinyanakiev: Not sure where that came from, will remove it.

PR #653, opened 2020-12-30
Split AAD Auth strategies.

valentinyanakiev:
Split AAD Bearer, On-behalf-of and Resource Owner Password Credentials strategies, their configurations and usage. Removed circular dependency between User, Authentication and MS Graph modules.


techsmyth (approved):
Nice update!

PR #654, opened 2020-12-30
Removed unnecessary type casts. Refactored RemoveByID methods to use Repository.Delete (it uses ID as an argument).

valentinyanakiev:


techsmyth (approved):
Nice set of cleanup, plenty of legacy there from me...

PR #655, opened 2020-12-30
Changed container to run with production settings

valentinyanakiev:
- Default node_nev = production
- The application runs from dist/main

Submitting as draft as this requires testing / discussion re defaults


techsmyth:
Added in the volumes support as per my comment...

techsmyth (approved):
Working as far as I can tell.

Issue #571, opened 2020-12-01
Single implementation of textID across entities


techsmyth:
## Description
As a developer I want a single implementation of textID
## Acceptance criteria

- [ ] One implementation of textID across entities, with same validatinon rules

PR #682, opened 2021-01-12
Text fields are either varchar(255) or text.

atstojanov:
Consistent nullable fields between graphql and database models.

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/domain/challenge/challenge.entity.ts
+++ b/src/domain/challenge/challenge.entity.ts
@@ -37,14 +37,14 @@ export class Challenge extends BaseEntity implements IChallenge, IGroupable {
     nullable: false,
     description: 'The name of the challenge',
   })
-  @Column('varchar', { length: 100 })
+  @Column('varchar')
   name: string;
 
   @Field(() => String, {
     nullable: false,
     description: 'A short text identifier for this challenge',
   })
-  @Column('varchar', { length: 15 })
+  @Column('varchar')
   textID: string;
 
   @Field(() => Context, {
@@ -82,7 +82,7 @@ export class Challenge extends BaseEntity implements IChallenge, IGroupable {
       'The maturity phase of the challenge i.e. new, being refined, ongoing etc',
   })
   @Column({ nullable: true })
-  state: string;
+  state?: string;
 
   @Field(() => Tagset, {
     nullable: true,
--- a/src/domain/challenge/challenge.entity.ts
+++ b/src/domain/challenge/challenge.entity.ts
@@ -37,14 +37,14 @@ export class Challenge extends BaseEntity implements IChallenge, IGroupable {
     nullable: false,
     description: 'The name of the challenge',
   })
-  @Column('varchar')
+  @Column()
   name: string;
 
   @Field(() => String, {
     nullable: false,
     description: 'A short text identifier for this challenge',
   })
-  @Column('varchar')
+  @Column()
   textID: string;
 
   @Field(() => Context, {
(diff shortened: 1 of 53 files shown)


src/domain/project/project.interface.ts
```diff
@@ -7,7 +7,7 @@ export interface IProject {
   textID: string;
   name: string;
   description?: string;
-  state: string;
+  state?: string;
```
techsmyth: state should be a required field, certainly moving forward - so suggest not to make it optional now

src/domain/profile/profile.dto.ts
```diff
@@ -1,16 +1,13 @@
-import { InputType, Field } from '@nestjs/graphql';
-import { MaxLength } from 'class-validator';
```
techsmyth, line 2: am guessing we want to leave these then in given the discussion this morning?

src/domain/project/project.entity.ts
```diff
@@ -43,7 +43,7 @@ export class Project extends BaseEntity implements IProject {
       'The maturity phase of the project i.e. new, being refined, committed, in-progress, closed etc',
   })
   @Column({ nullable: true })
-  state: string;
+  state?: string;
```
techsmyth: ditto - keep mandatory

src/domain/user/user.dto.ts
```diff
@@ -1,50 +1,51 @@
...
+  @MaxLength(255)
   lastName?: string;
 
   @Field({
     nullable: true,
     description: 'Email address is required for creating a new user',
   })
-  @MaxLength(120)
+  @MaxLength(255)
+  @IsEmail()
```
techsmyth, line 28: nice!

src/domain/challenge/challenge.interface.ts
```diff
@@ -8,7 +8,7 @@ export interface IChallenge {
   id: number;
   name: string;
   textID: string;
-  state: string;
+  state?: string | null;
```
techsmyth: keep mandatory

src/domain/challenge/challenge.entity.ts
```diff
@@ -82,7 +82,7 @@ export class Challenge extends BaseEntity implements IChallenge, IGroupable {
       'The maturity phase of the challenge i.e. new, being refined, ongoing etc',
   })
   @Column({ nullable: true })
-  state: string;
+  state?: string;
```
techsmyth: keep mandatory

src/domain/user/user.dto.ts
techsmyth: is it possible to create our own for e.g. textID?

techsmyth (changes requested):
Nice update!

The state field does need to be mandatory as this will be used for lifecycles.

Also, need to take care that business rules related to lengths of eg tagline, textID are present on the dto objects.

src/domain/profile/profile.dto.ts
atstoyanov: Only for the description. The avatar is url. The url by specification can be up to 2MB. Which is not very likely to happen.

atstoyanov: But we can add constrain IsUri

src/domain/user/user.dto.ts
atstoyanov: it is done for #571

techsmyth (approved):

valentinyanakiev (approved):

PR #687, opened 2021-01-12
add check that user groups cannot be created with empty strings

techsmyth:


techsmyth:
Note: now a related issue that the client does not display the returned error nicely, only in default box bottom right

valentinyanakiev (approved):

PR #693, opened 2021-01-13
Validation using class-validator.

valentinyanakiev:

The PR adds global InputType validation.

### Additional context

There are some dead parts of the code / samples (how to chain handlers, custom validation decorator) left out for reference.


src/domain/reference/reference.dto.ts
```diff
@@ -1,17 +1,19 @@
...
 export class ReferenceInput {
   @Field({ nullable: true })
-  @MaxLength(30)
+  @MaxLength(MID_TEXT_LENGTH)
   name!: string;
 
   @Field({ nullable: true })
-  @MaxLength(300)
+  @IsDataURI()
   uri!: string;
```
atstoyanov: This field contains URLs should be used `@IsUrl` instead.

Optionally if we want to store data in URI form we may need to find a way to apply two validators with `OR` operation. The default one is `AND`.

atstojanov (changes requested):

atstojanov (approved):

techsmyth (commented):
Think we need to bring the set of lenghts from 7 ==> 4

So would then have: TINY, SMALL, MEDIUM, LARGE

Suggest to chat through...

PR #706, opened 2021-01-14
updating of organisation also updates profile

techsmyth:


src/domain/organisation/organisation.service.ts
```diff
@@ -58,6 +58,35 @@ export class OrganisationService {
...
+    if (!existingOrganisation)
+      throw new EntityNotFoundException(
+        `Oganisation with given ID (${orgID}) not found!`,
+        LogContext.CHALLENGES
+      );
+
+    // Merge in the data
+    if (organisationData.name) {
+      existingOrganisation.name = organisationData.name;
+      await this.organisationRepository.save(existingOrganisation);
```
valentinyanakiev, line 75: await this.organisationRepository.save(existingOrganisation) --> returns Pomise<Organisation> with the updated Org. If you switch the two blocks (Merge the data and Check the tagsets) you can simply return this result --> return await this.organisationRepository.save(existingOrganisation) instead of making a second query to the DB for the same result.

valentinyanakiev (changes requested):

valentinyanakiev:
All integration / e2e tests pass otherwise

src/domain/organisation/organisation.service.ts
techsmyth: Good point re saving returning the updated saved entity - but do not think that we can do what you suggest. If swap the blocks then would be saving the non-updated profile which would overwrite the update just made....

valentinyanakiev (approved):

PR #719, opened 2021-01-18
updated validation to deal with empty strings

techsmyth:


src/domain/reference/reference.service.ts
```diff
@@ -86,7 +86,7 @@ export class ReferenceService {
         references.push(
           new Reference(
             referenceDTO.name,
-            referenceDTO.uri,
+            referenceDTO.uri || '',
```
valentinyanakiev: On second thought, that's not OK. If a client of the API sends a request to update the Reference, it's OK to have an undefined URI, in case the field is not updated. In that case it will be overwritten with empty one here.

src/domain/reference/reference.service.ts
```diff
@@ -33,7 +33,7 @@ export class ReferenceService {
   async createReference(referenceInput: ReferenceInput): Promise<IReference> {
     const reference = new Reference(
       referenceInput.name,
-      referenceInput.uri,
+      referenceInput.uri || '',
```
valentinyanakiev, line 20: That's OK.

src/domain/reference/reference.dto.ts
```diff
@@ -13,9 +13,10 @@ export class ReferenceInput {
   name!: string;
 
   @Field({ nullable: true })
-  @IsUrl()
+  @IsOptional()
+  //@IsUrl()
   @MaxLength(MID_TEXT_LENGTH)
-  uri!: string;
+  uri?: string;
```
valentinyanakiev, line 19: We need to discuss this field validation.

src/domain/reference/reference.service.ts
techsmyth: a more basic issue is that we are matching on name, not using the id of the reference :(

valentinyanakiev (approved):
URL / URI needs to be looked at later. Rest looks OK and tests are passing.

PR #751, opened 2021-01-27
added logic to block renaming to a restricted group name

techsmyth:


valentinyanakiev (changes requested):
With this logic, the test 'should throw error for updating restricted group name' test/functional/integration/group/groups.it-spec.ts fails so there is a regression where restrictedGroupName can be updated.

Comoque1:
Checking the "Search" failing tests

Comoque1 (approved):

Comoque1 (approved):

PR #764, opened 2021-02-05
Server 754. Major change re OIDC Bearer token + Account Management.

valentinyanakiev:

- Generic OIDC Bearer Strategy replacing AAD Bearer implementation
- Auth logic split between OIDC Bearer Strategy and Auth Service
- Account Management logic abstracted with separate AAD Account Management (Dynamic) Module plugged in
- Test suite updated accordingly.

### Additional context

OBO flow is still pending.


.env.oidc.default
```diff
@@ -0,0 +1,5 @@
+OAUTH2_CLIENT_PROVIDER_OIDC_ISSUER=https://login.microsoftonline.com/524e761c-d162-4fdf-ab43-2855246d986c/v2.0
+OAUTH2_CLIENT_REGISTRATION_LOGIN_CLIENT_ID=505041fc-fca2-4a74-88ee-6d50a6417e38
+OAUTH2_CLIENT_REGISTRATION_LOGIN_CLIENT_SECRET=3t~CH3Tkz-Fx7CVA9VjXjCW_CK_8kS8fD.
```
techsmyth, line 3: Presumably this is for our Playground env?

src/config/oidc.config.ts
```diff
@@ -0,0 +1,10 @@
+import { registerAs } from '@nestjs/config/dist/utils/register-as.util';
```
techsmyth, line 1: this file needs a discussion

src/main.ts
```diff
@@ -35,6 +37,18 @@ const bootstrap = async () => {
     })
   );
 
+  app.use(
+    session({
+      secret: 'keyboard cat',
```
techsmyth: what is this for?

src/utils/aad/aad.account-management.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const AAD_ACCOUNT_MANAGEMENT_MODULE_OPTIONS = 'aadModuleOptions';
+export const AAD_ACCOUNT_MANAGEMENT_MODULE_PROVIDER = 'aadModule';
```
techsmyth: All these constants are used within one file so probably best to have them in that file as not used elsewhere? Also names (i.e. values) do need an update?

src/utils/account/account.module.ts
```diff
@@ -1,11 +1,10 @@
 import { Module } from '@nestjs/common';
 import { AccountService } from './account.service';
-import { MsGraphModule } from '@utils/ms-graph/ms-graph.module';
 import { UserModule } from '@domain/user/user.module';
 import { AccountResolver } from './account.resolver';
 
 @Module({
-  imports: [MsGraphModule, UserModule],
+  imports: [UserModule],
```
techsmyth, line 7: For later: ideally Account should not have a dependency on User service? Basically two pieces of logic that use the userservice need to be moved outside. Actually the updates re register / apply might also make the code that uses userservice redundant...

src/app.module.ts
```diff
@@ -98,15 +104,27 @@ import { ValidationPipe } from '@utils/validation/validation.pipe';
...
       useClass: WinstonConfigService,
     }),
     SearchModule,
     KonfigModule,
+    AadAccountManagementModule.forRootAsync({
+      imports: [ConfigModule],
+      inject: [ConfigService],
+      useFactory: async (configService: ConfigService) => ({
+        clientID: configService.get<AuthConfig>('aad_ropc')?.clientID as string,
+        clientSecret: '',
```
techsmyth, line 121: presumably user has to fill it in? Looks strange to have an empty string...

techsmyth (commented):
Nice update! A few clarifying comments that need a chat through - but nothing major so if you respond to points then think for me this can go in...

.env.oidc.default
valentinyanakiev: At the moment not used at all - I am re-using the old AAD variables. This is a leftover and needs to be cleaned up.

src/config/oidc.config.ts
valentinyanakiev: It needs refactoring for sure, env variables need to be cleaned up and config files with them. (aad.client.config.ts, aad.config.ts, aad.obo.config.ts, aad.ropc.config.ts and oidc.config.ts). Will raise a card on the board for that.

src/main.ts
valentinyanakiev: Leftover - session key...

src/utils/aad/aad.account-management.constants.ts
valentinyanakiev: Completely disagree - that's standard notation and it's better to have 10 files with 5 lines each with clear distinction than 1 file with 10 purposes and 50 lines. Names are tokens and can be anything - agreed it will be cleaner to change the strings but to be honest a guid will do the same (and quite frankly might be better, as this tokens need to be unique, otherwise injecting them will fail).

src/utils/account/account.module.ts
valentinyanakiev: Agreed - that needs a discussion. Out-of-scope for now thou - otherwise this PR will be 5 epics into 1.

src/config/oidc.config.ts
atstoyanov: it can be imported shorter:
```javascript
import { registerAs } from '@nestjs/config';
```

src/app.module.ts
valentinyanakiev: This needs to be cleaned up overall, together with the rest of the env variables... The clientSecret is optional - it's mandatory only for some authentication flows (e.g. client credentials grant).

src/main.ts
atstoyanov: It is a secret to sign the sessions cookie, should be changed to something more secure like configurable hash. Have to be preserved.

valentinyanakiev: Fixed in ba4381f9fc391608ca423d5692abd6b9ed64c4fe

src/main.ts
```diff
@@ -35,6 +37,18 @@ const bootstrap = async () => {
...
+      resave: false,
+      saveUninitialized: true,
+      cookie: { secure: true },
```
atstoyanov: passport session needs store that can be preserved between service restart. By default the InMemory store is not a good idea. All user sessions will be lost.

src/app.module.ts
valentinyanakiev: https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/server/765

src/utils/aad/aad.account-management.constants.ts
valentinyanakiev: https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/server/765

src/config/oidc.config.ts
valentinyanakiev: https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/server/765

.env.oidc.default
valentinyanakiev: https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/server/765

src/main.ts
valentinyanakiev: @atstojanov this is redundant and outdated, I removed it. Sessions are mandatory for different auth flows.

src/utils/aad/aad.account-management.constants.ts
valentinyanakiev: Improved / fixed in ccce2a9f05949ea416259c7af1ca183ba955375f

techsmyth (approved):

PR #772, opened 2021-02-11
Applications to join

atstojanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


techsmyth:
Just tried to run the migration on a clean db + it failed with the error below:
image


Steps:
drop the schema + create new db
checkout develop + run migrations
checkout this branch + run migrations - it fails as per above msg

src/domain/application/application.entity.ts
```diff
@@ -0,0 +1,78 @@
...
+  @Field(() => User)
+  @ManyToOne(
+    () => User,
+    user => user.focalPoints,
+    { eager: true, cascade: true, onDelete: 'CASCADE' }
+  )
+  user!: User;
+
+  @Field(() => [Question])
+  @ManyToMany(
```
techsmyth, line 53: Questions are not going to be shared i.e. each question entity is specific to a particular applicaiton.

So this can be oneToMany. Ditto for all ther manyToMany as far as I can tell i.e. applications are also not re-used across entities.

src/domain/application/application.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    private ecoverseRepository: Repository<Ecoverse>,
+    @InjectRepository(Challenge)
+    private challengeReposity: Repository<Challenge>,
+    @InjectRepository(Opportunity)
+    private opportunityReposity: Repository<Opportunity>,
+    private userService: UserService,
+    @Inject(WINSTON_MODULE_NEST_PROVIDER) private readonly logger: LoggerService
+  ) {}
+
+  async createApplication(
```
techsmyth, line 29: This is now over in the factory correct?

src/domain/challenge/challenge.resolver.mutations.ts
```diff
@@ -162,4 +164,21 @@ export class ChallengeResolverMutations {
       organisationID
     );
   }
+
+  @Roles(
```
techsmyth: Needs to be callable by a user who does not have these roles...to chat through

src/domain/ecoverse/ecoverse.resolver.mutations.ts
```diff
@@ -164,4 +166,20 @@ export class EcoverseResolverMutations {
 
     return organisation;
   }
+
+  @Roles(
```
techsmyth: ditto - users need to be able to create these without having an elevated roles.

techsmyth (changes requested):
Nice full PR.

A couple of points need discussing...

src/domain/application/application.entity.ts
atstoyanov: The Questions are using the generic entity `nvp` so it is not possible to use OneToMany. If we specify OneToMany/ManyToOne the nvp entity will have ApplicationId as foreign key, and every future entities having such relation will amend that table. To resolve that we are defining the ralation like ManyToMany, but we are going to use it like `1:n`

src/domain/challenge/challenge.resolver.mutations.ts
atstoyanov: We need to introduce new role - `Authenticated` which means that the user must be authenticated, but it is not necessary to be member of any group.

atstoyanov: Or we have this? Using `Guard` without `Roles`?

techsmyth: would prefer not to have another group that is equivalent to being a User - by implication is there is a User role, which is anyone that is authenticated. The guard may need to be smart to be able to understand this.

@Valentin: your thoughts?

valentinyanakiev: @techsmyth @atstojanov The guard may not be that smart at the moment, but if that's the way we want to go, we can switch the behavior in the user : role matching and add a check for a user with no roles. We can default the behavior to anonymous if the user is authenticated and not in any roles.

techsmyth: key is that there is a difference between an non-authenticated + authenticated user. So it may be that there is a "known" term authenticated that the guard understands - or we create another guard that simply checks for authenticated?

valentinyanakiev: We can consider adding another role (restricted group) but definitely not another guard for that.

techsmyth: not another restricted group please :)

how about another guard isAuthenticated?

techsmyth (approved):

PR #785, opened 2021-02-16
Separation of users from ecoverse members; autocreate profile if valid token

techsmyth:
Note: still a bug whereby a user with no roles can carry out all mutations!


src/utils/auth/auth.service.ts
```diff
@@ -12,12 +13,24 @@ export class AuthService {
...
+    let knownUser = await this.userService.getUserWithGroups(token.email);
 
-    if (!knownUser)
-      throw new AuthenticationException(
-        `No user with email ${token.email} found!`
-      );
+    if (!knownUser) {
+      // User does not yet have a profile on the platform, create one
+      const userData = new UserInput();
+      userData.email = token.email;
```
valentinyanakiev, line 21: This shouldn't be here - the business logic building the new object should be in the factory creating the object, validating the DTO should be using decorators in the DTO (Input) class. You can also decorate the fields in the entity class with default values - that will make init consistent across the application.

valentinyanakiev (changes requested):
Overall great to have that update - it's definitely a big step in the right direction.
I would like to see streamlined data validation and building going forward - hence the comment in AuthService.

From the test suites, the following need to be reviewed (as they are failing):

- test/non-functional/auth/ecoverse-member-user.it-spec.ts
- test/functional/e2e/user-management/user-to-group.e2e-spec.ts
- test/functional/e2e/user-management/remove-user.e2e-spec.ts

Largely we receive code 400 instead of 200 for the requests in these suites, but there are a couple where the logic has changed as well.

src/utils/auth/auth.service.ts
techsmyth: Agreed - added separate issue (#786) to update this code to throw a specific exception that can be picked up by the client.

atstojanov (approved):

PR #800, opened 2021-03-02
Draft : Reverted to Microsoft AAD-strategy

valentinyanakiev:
Default is AAD, if not authenticated there falls back trying to authenticate with Simple Auth Provider.

Draft as the code can be reviewed but I have checked only baseline scenarios.


techsmyth (approved):
Looks good. Removed the jwt decode package that was not longer needed.

PR #801, opened 2021-03-03
Split authorisation from authentication; refactored 

techsmyth:
Introduced a new DTO that is stored in the request context instead of the direct CT User object
- this has now additional info such as email etc

Split out authorisation from authentication.

Introduced explicit AuthorisationRoles enum

Moved decorators to be in the relevant module.


src/domain/user/user.resolver.mutations.ts
```diff
@@ -1,25 +1,47 @@
...
+    description:
+      'Creates a new user profile on behalf of an admin or the user account owner.',
+  })
+  @Profiling.api
+  async createUser(
+    @AccountMap() accountMapping: AccountMapping,
+    @Args('userData') userData: UserInput
+  ): Promise<IUser> {
+    if (accountMapping.newUser()) {
+      this.userService.validateAuthenticatedUserSelfAccessOrFail(
```
valentinyanakiev: In order to remove circular dependency User <--> Auth Modules you have broken domain boundaries. You have, essentially, moved the Authorization logic to the User Service which is a no-go, in my opinion. Having circular dependency between User and Auth is actually a common occurrence as Auth modules in nestJS world have the responsibility to retrieve the user profile and pass it to the request object (profile, not account), which is in turn handled by the AuthGuard. Moving the Auth logic in UserService is creating a bigger problem, not really solving the problem but rather trying to hide it.

src/utils/authorization/graphql.guard.ts
```diff
@@ -50,23 +50,39 @@ export class GqlAuthGuard extends AuthGuard(['simple-auth-jwt', 'bearer']) {
...
       ({ name }) =>
-        name === RestrictedGroupNames.GlobalAdmins || this.roles.includes(name)
+        name === AuthorizationRoles.GlobalAdmins || this.roles.includes(name)
     );
   }
 
-  handleRequest(err: any, user: any, info: any, _context: any, _status?: any) {
+  handleRequest(
+    err: any,
+    accountInfo: any,
```
valentinyanakiev: This is a user object or userProfile. It is not the account.

src/utils/authorization/graphql.guard.ts
```diff
@@ -50,23 +50,39 @@ export class GqlAuthGuard extends AuthGuard(['simple-auth-jwt', 'bearer']) {
...
   }
 
-  matchRoles(userGroups: IUserGroup[]): boolean {
+  matchRoles(accountMapping: AccountMapping): boolean {
     if (this.roles.length == 0) return true;
-    return userGroups.some(
+    const ctUser = accountMapping.user;
+    if (!ctUser) {
+      if (this.roles.includes(AuthorizationRoles.NewUser)) {
+        return true;
```
valentinyanakiev: This essentially gives global-admin permissions for a new user

src/utils/authorization/authorization.roles.ts
```diff
@@ -0,0 +1,7 @@
+export enum AuthorizationRoles {
```
valentinyanakiev, line 1: This should probably be in a separate entity, but anyway - better than RestrictedGroupNames for sure.

src/utils/auth/jwt.strategy.ts
```diff
@@ -18,13 +19,7 @@ export class JwtStrategy extends PassportStrategy(Strategy, 'simple-auth-jwt') {
   }
 
   async validate(payload: JwtPayload) {
-    const user = await this.authService.getUserFromJwtPayload(payload);
-    if (!user) {
-      throw new AuthenticationException(
-        'Failed to find user profile matching the email in jwt!'
```
valentinyanakiev, line 24: This line is the line we agreed on with Angel to be passed onto the client to be able to handle the new-user flow. Changing the flow from throwing an exception to returning undefined breaks this flow.

src/utils/auth/auth.service.ts
```diff
@@ -1,33 +1,13 @@
...
-import { AuthenticationException } from '@utils/error-handling/exceptions';
-import jwt_decode from 'jwt-decode';
-import { IUser } from '@domain/user/user.interface';
+import { AccountMapping } from './account.mapping';
 
 @Injectable()
 export class AuthService {
   constructor(private readonly userService: UserService) {}
 
-  async getUserFromToken(encodedToken: any): Promise<[IUser, string]> {
```
valentinyanakiev, line 11: At the moment the user comes in a different format from different requests. You are unifying this without taking into account that the payload is deserialized differently. This will not work in all cases.

src/domain/user/user.service.ts
```diff
@@ -179,6 +176,19 @@ export class UserService {
     return (await this.userRepository.find()) || [];
   }
 
+  validateAuthenticatedUserSelfAccessOrFail(
```
valentinyanakiev: This doesn't belong in the UserService.

valentinyanakiev (changes requested):
In my honest opinion this will:

1. Replace some issues/smells with others, potentially bigger ones.
2. Break multiple code paths.
3. Open (huge) security holes.
4. Will not improve the codebase.
5. Break standard notation / terms of the Passport flow, making it hard to maintain / understand by 3rd parties.

valentinyanakiev (approved):

PR #811, opened 2021-03-08
major cleanup of server documentation; docker composition updated

techsmyth:
server documentation split up and revamped:
- multiple sub files
- updated development instructions to use docker composition

reworking of docker composition to 
- isolate from demo
- added simple auth into composition;


.env.docker
```diff
@@ -18,4 +18,9 @@ AUTH_AAD_CLIENT_APP_ID=feb6d8b1-8cc2-4e7d-a419-ad7b544b0832
 AUTH_AAD_CHERRYTWIST_API_SCOPE=api://505041fc-fca2-4a74-88ee-6d50a6417e38/.default
 AUTH_AAD_CLIENT_LOGIN_REDIRECT_URI=http://localhost:3000
 
+AUTH_SIMPLE_AUTH_ISSUER=localhost:3003
+AUTH_SIMPLE_AUTH_TOKEN_ENDPOINT=localhost:3003/auth/login
+AUTH_SIMPLE_AUTH_CLIENT_SECRET=!#@$#SDFRSF
```
valentinyanakiev, line 23: AUTH_SIMPLE_AUTH_CLIENT_SECRET and SECRET are the same thing

docker-compose.yml
```diff
@@ -57,4 +61,26 @@ services:
...
+    environment:
+      - DATABASE_HOST
+      - MYSQL_ROOT_PASSWORD
+      - MYSQL_DATABASE=ct_auth_db
+      - WAIT_HOSTS
+      - WAIT_HOSTS_TIMEOUT
+      - WAIT_SLEEP_INTERVAL
+      - WAIT_HOST_CONNECT_TIMEOUT
+      - SERVICE_ENDPOINT_PORT
+      - SECRET
```
valentinyanakiev: SECRET= ...
Don't add an extra env variable in the config which is exactly the same (SECRET <=> AUTH_SIMPLE_AUTH_CLIENT_SECRET) just to pass it directly to the docker-compose.

docs/Configuration.md
```diff
@@ -0,0 +1,119 @@
...
+```conf
+AUTH_ENABLED=true. Specifies whether authentication should be enabled on the CT Web Client and CT Server.
+```
+
+### **AAD Authentication Provider**
+
+The following environment variables are used to configure AAD Authentication:
+
+```conf
+AUTH_AAD_TENANT=[tenant (directory) ID]
```
valentinyanakiev, line 51: placeholder comment -> we need to review this configuration...

valentinyanakiev (changes requested):
Infinitely better structure. So so so much better and easier to navigate...

Two minor comments - one more technical, one more regarding content:
- try to limit the number of variables and not duplicate them
- you keep telling the story that we have a microservices architecture and it has the same story on the landing page. This is completely untrue - and having the concept of CT Server right next to 'microservices' architecture will be a thorn in the eyes of any technical person and that will be immediately challenged. We have service-oriented architecture but we are worlds apart from the dogmatic microservices approach... Also, these colors are killing me - we really need to land a palette. It's really, _really_ offputting, especially when it's the first thing you see in CT.  Look how easy on the eye are the https://sdgs.un.org/goals colors. Also our own colors in CT are quite good - but the presentations don't reflect that.

docs/Configuration.md
techsmyth: Config needs a wider review

docker-compose.yml
techsmyth: Updated...

.env.docker
techsmyth: updated

valentinyanakiev (approved):

PR #821, opened 2021-03-11
added community entity; graphql api clean up

techsmyth:
Added Community Entity to Ecoverse, Challenge + Opportunity. 
Tidied up module loading: AppModule startup has much less modules; Barrel files do not include services.
A lot of code removed. 
Community hierarchy explicit via community parent relationship.
Community has a type to differentiate where it is used.
Organisations now live independent of Ecoverse entity, similar to Users - and then can be added. Implied moving Organisation related Queries to the Organisation module. 
Graphql samples have been updated.

Mutations:
- addUserToEcoverse/Opportunity/Challenge ==> addUserToCommunity
- createGroupToEcoverse/Opportunity/Challenge ==> createGroupToCommunity
- createApplicationOnEcoverse/Opportunity/Challenge ==> createApplicationOnCommunity
- updateOrganisation/Opportunity/Challenge ==> now take a single DTO for the update, whereby the ID for the entity is now a string
- createOpportunityOnChallenge ==> createOpportunity, and now takes a single argument
- adding / removing Challenge Leads now work with strings

Query updates:
- added new ecoverse query
- removed host, name, context, tagset ==> all now via Ecoverse
- challenge() / organisation() / opportunity() now take strings instead of numbers

Field updates:
- Ecoverse: removed organisations; they now live at global level 
- User.memberof : returns the communities instead of groups / challenges / opportunities

Removing entities:
- ORM definitions updated to pass "on delete: CASCADE" in places whereby there is a parent-child relationship (1-1 or 1-many)
- Manually still remove the relations which ideally we can remove later.

Implied changes
- Admin UI
- Search / anywhere that uses memberof
- Test harnesses heavily impacted 
- client.lib + populator: updated

Light testing has been carried out, will be regressions.

Performance is a concern...

Later: consider the relationship of Community <==> Organisations. For left them as independent entities.


src/common/interfaces/communityable.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { Community } from '@domain/community/community';
+
+export interface ICommunityable {
```
valentinyanakiev, line 3: That is an odd name

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -472,37 +361,34 @@ export class ChallengeService {
...
-
-    const existingApplication = challenge.applications?.find(
-      x => x.user.id === applicationData.userId
+  async removeChallengeLead(
+    challengeID: number,
+    organisationID: number
+  ): Promise<boolean> {
+    let challenge;
+    // eslint-disable-next-line prefer-const
+    [challenge, {}] = await this.getChallengeAndOrganisation(
```
valentinyanakiev: This is an odd approach, have a method that returns two objects just to take one of them. This approach is more for external methods where you don't need one / more than one of the return objects, not for our own services.

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -67,51 +51,32 @@ export class ChallengeService {
...
-      `Adding userGroup (${groupName}) to challenge (${challengeID})`,
-      LogContext.CHALLENGES
-    );
-
-    // Try to find the challenge
-    const challenge = await this.getChallengeOrFail(challengeID, {
-      relations: ['groups'],
+  // Loads the challenges into the challenge entity if not already present
+  async loadCommunity(challengeId: number): Promise<ICommunity> {
+    const challengeWithCommunity = await this.getChallengeOrFail(challengeId, {
```
valentinyanakiev: why not just challenge?

src/domain/challenge/challenge/challenge.entity.ts
```diff
@@ -52,30 +49,31 @@ export class Challenge extends BaseEntity implements IChallenge, IGroupable {
     nullable: true,
     description: 'The shared understanding for the challenge',
   })
-  @OneToOne(() => Context, { eager: true, cascade: true })
+  @OneToOne(() => Context, { eager: false, cascade: true })
```
valentinyanakiev: It's OK to be eager loaded if it's one-to-one - won't have a huge performance hit.

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -67,51 +51,32 @@ export class ChallengeService {
     if (!challenge.context) {
       challenge.context = new Context();
     }
+
+    if (!challenge.community) {
+      challenge.community = new Community(challenge.name, [
+        RestrictedGroupNames.Members,
```
valentinyanakiev: Are we keeping both roles and restrictedGroupNames?

src/domain/community/community/community.service.ts
```diff
@@ -0,0 +1,247 @@
...
+      community,
+      groupName
+    );
+    await this.communityRepository.save(community);
+
+    return group;
+  }
+
+  // Loads the group into the Community entity if not already present
+  async loadGroups(community: Community): Promise<IUserGroup[]> {
```
valentinyanakiev: ICommunity instead of Community

src/domain/community/community/community.service.ts
```diff
@@ -0,0 +1,247 @@
...
+    );
+    if (!Community)
+      throw new EntityNotFoundException(
+        `Unable to find Community with ID: ${communityID}`,
+        LogContext.COMMUNITY
+      );
+    return Community;
+  }
+
+  async removeCommunity(CommunityID: number): Promise<boolean> {
```
valentinyanakiev: camelCase

src/domain/community/community/community.service.ts
```diff
@@ -0,0 +1,247 @@
...
+      for (const group of community.groups) {
+        await this.userGroupService.removeUserGroup(group.id);
+      }
+    }
+
+    await this.communityRepository.remove(community as Community);
+    return true;
+  }
+
+  async isUserMember(userID: number, CommunityID: number): Promise<boolean> {
```
valentinyanakiev: camelCase

src/domain/community/community/community.service.ts
```diff
@@ -0,0 +1,247 @@
...
+    const application = await this.applicationFactoryService.createApplication(
+      applicationData
+    );
+
+    community.applications?.push(application);
+    await this.communityRepository.save(community);
+    return application;
+  }
+
+  async getApplications(Community: Community) {
```
valentinyanakiev: ICommunity

valentinyanakiev (commented):
1. Very inconsistent naming - some places are camelCase, some PascalCase. Same variable in different classes follows different notation.
2. Names can be improved
3. Input types for methods are inconsistent - in some places are used interfaces, in some places entities (where interfaces can be used). Stick to interfaces where possible
4. Eager loading is OK (maybe even preferred) in one-to-one relationships

Haven't tested it functionally but overall more lines deleted than added which is always a good thing.

src/domain/challenge/challenge/challenge.service.ts
techsmyth: Updated code

src/domain/challenge/challenge/challenge.entity.ts
techsmyth: Made Community all eager loaded - loading stops by groups.

src/domain/community/community/community.service.ts
techsmyth: done

src/domain/community/community/community.service.ts
techsmyth: done

src/domain/community/community/community.service.ts
techsmyth: done

Comoque1:
Creating a challenge is returning error, but the entity is created:
```
mutation CreateChallenge($challengeData: ChallengeInput!) {
  createChallenge(challengeData: $challengeData) {
    name,
    id
  }
}
```
```
{
  "challengeData":
  {
    "name": "test",
    "textID": "test"
  }
}
```
image
image

## NOTE: 
The behavior above is when the DB is in clean state and start the service

Comoque1:
Creating opportunity on challenge is returning error:` "message": "ER_DUP_ENTRY: Duplicate entry '20' for key 'community.REL_8e8283bdacc9e770918fe68933'"`

## Observation:
The first created entity receive `communityId`.
All the rest after intending are with `communityId: null`

image
image

Comoque1:
Would it be possible to limit the number of nested parent queries?

image

```
query {
  groups {
    parent {
      __typename
      ... on Community {
        name
        type
        groups {
          parent {
            __typename
            ... on Community {
              name
              groups {
                parent {
                  __typename
                  ... on Community {
                    name
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}

```

techsmyth:
> Would it be possible to limit the number of nested parent queries?
> 
> image
> 
> ```
> query {
>   groups {
>     parent {
>       __typename
>       ... on Community {
>         name
>         type
>         groups {
>           parent {
>             __typename
>             ... on Community {
>               name
>               groups {
>                 parent {
>                   __typename
>                   ... on Community {
>                     name
>                   }
>                 }
>               }
>             }
>           }
>         }
>       }
>     }
>   }
> }
> ```

Good suggestion. Can you please raise this as a separate card on the board? It is independent of the Community refactoring.

src/app.module.ts
```diff
@@ -68,12 +55,13 @@ import { ApplicationFactoryModule } from '@domain/community/application/applicat
...
     }),
     TypeOrmModule.forRootAsync({
+      name: 'default',
       imports: [ConfigModule],
       inject: [ConfigService],
       useFactory: async (configService: ConfigService) => ({
         type: 'mysql',
         insecureAuth: true,
-        synchronize: false,
+        synchronize: true, //warning: to be reverted when go back to migrations!
```
valentinyanakiev: needs to be false, migrations are updated

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -67,55 +54,38 @@ export class ChallengeService {
     if (!challenge.context) {
       challenge.context = new Context();
     }
+
+    if (!challenge.community) {
+      challenge.community = new Community(
+        challenge.name,
+        CommunityType.CHALLENGE,
+        [AuthorizationRoles.Members]
```
valentinyanakiev, line 60: There is an implicit mapping Roles <--> RestrictedGroupNames. I'd rather have Roles in the Community entity, logically that's what we have, not RestrictedGroupNames.

src/domain/challenge/opportunity/opportunity.entity.ts
```diff
@@ -116,19 +115,6 @@ export class Opportunity extends BaseEntity
...
-  )
-  @JoinTable({
-    name: 'opportunity_application',
-  })
-  applications?: Application[];
-
-  // The restricted group names at the Opportunity level
-  restrictedGroupNames: string[];
   // The restricted actor group names at the Opportunity level
   restrictedActorGroupNames: string[];
```
valentinyanakiev, line 127: Roles instead of RestrictedGroupNames?

src/domain/community/community/community.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+  )
+  opportunity?: Opportunity;
+
+  // The parent community can have many child communities; the relationship is controlled by the child.
+  @ManyToOne(() => Community, { eager: false, cascade: false })
+  parentCommunity?: Community;
+
+  // The restricted group names at the Community level
+  @Column('simple-array')
+  restrictedGroupNames: string[];
```
valentinyanakiev, line 79: roles?

src/domain/community/community/community.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+    restrictedGroupNames: string[]
+  ) {
+    super();
+    this.name = name;
+    this.restrictedGroupNames = restrictedGroupNames;
+    this.type = communityType;
+  }
+}
+
+export enum CommunityType {
```
valentinyanakiev: move the enum to common -> enums

src/domain/community/community/community.service.ts
```diff
@@ -0,0 +1,317 @@
...
+    const community = (await this.getCommunityOrFail(id, {
+      relations: ['applications'],
+    })) as Community;
+
+    const existingApplication = community.applications?.find(
+      x => x.user.id === applicationData.userId
+    );
+
+    if (existingApplication) {
+      throw new ApolloError(
```
valentinyanakiev: Wrap this in a custom exception. We don't use ApolloErrors directly - we have some custom metadata in our exceptions.

valentinyanakiev (changes requested):
A few minor changes required and one more major design decision we need to make - I believe roles and restrictedGroupNames are badly intertwined and the overall feel is a code smell - imho they need to be decoupled.

src/domain/community/community/community.entity.ts
techsmyth: Fixed in https://github.com/cherrytwist/Server/pull/821/commits/65fe1f0bdf1729691b2b15de0e04b3e9e6a36c13

src/domain/community/community/community.service.ts
techsmyth: Fixed in https://github.com/cherrytwist/Server/pull/821/commits/65fe1f0bdf1729691b2b15de0e04b3e9e6a36c13

src/domain/challenge/challenge/challenge.service.ts
techsmyth: Pick up as a separate topic

src/app.module.ts
techsmyth: Done by Evgeni

src/domain/challenge/opportunity/opportunity.entity.ts
techsmyth: To pick up later

src/domain/community/community/community.entity.ts
techsmyth: To pickup later

atstoyanov:
Migrations aren't working.

```sql
'CREATE TABLE `cherrytwist`.`query-result-cache` (`id` int NOT NULL AUTO_INCREMENT, `identifier` varchar(255) NULL, `time` bigint NOT NULL, `duration` int NOT NULL, `query` text NOT NULL, `result` text NOT NULL, PRIMARY KEY (`id`)) ENGINE=InnoDB'
```
The table name should not include the schema `cherrytwist.query-result-cache` -> `query-result-cache`

Why migrations were deleted and recreated from scratch?
Why there are 2 migrations which at first glance seems to make same thing?

src/app.module.ts
```diff
@@ -68,6 +55,7 @@ import { ApplicationFactoryModule } from '@domain/community/application/applicat
       ],
     }),
     TypeOrmModule.forRootAsync({
+      name: 'default',
```
valentinyanakiev, line 58: it's default by default, why set it explicitly?

src/common/exceptions/application.invalid.state.exception.ts
```diff
@@ -0,0 +1,9 @@
+import { LogContext } from '@common/enums';
+import { CherrytwistErrorStatus } from '../enums/cherrytwist.error.status';
+import { BaseException } from './base.exception';
+
+export class ApplicationInvalidStateException extends BaseException {
+  constructor(error: string, context: LogContext) {
+    super(error, context, CherrytwistErrorStatus.APPLICATION_INVALID_STATE);
```
valentinyanakiev: Every exception is an application invalid state, what does this mean? Invalid in what sense?

src/domain/challenge/challenge/challenge.resolver.queries.ts
```diff
@@ -0,0 +1,27 @@
...
+@Resolver()
+export class ChallengeResolverQueries {
+  constructor(private challengeService: ChallengeService) {}
+
+  @Query(() => [Challenge], {
+    nullable: false,
+    description: 'The Challenges on this platform',
+  })
+  @Profiling.api
+  async challenges(): Promise<IChallenge[]> {
```
valentinyanakiev: If we are going towards multitenancy does this make sense? Shouldn't it be challenges in the context of an ecoverse?

src/domain/community/application/application.entity.ts
```diff
@@ -40,34 +38,23 @@ export class Application extends BaseEntity {
...
-  )
-  challenge?: Challenge[];
-
-  @ManyToMany(
-    () => Opportunity,
-    opportunity => opportunity.applications
+  @ManyToOne(
+    () => Community,
+    community => community.applications,
+    { eager: false, cascade: false, onDelete: 'CASCADE' }
```
valentinyanakiev, line 57: That's extemely inconsistent, somewhere cascade:true, somewhere false, somewhere onDelete: 'CASCADE', somewhere not. What is the reason? If it is implemented like that, the services will be inconsistent as well and overall the approach to deleting will be extremely inconsistent.

src/domain/community/organisation/organisation.resolver.queries.ts
```diff
@@ -0,0 +1,28 @@
...
+@Resolver()
+export class OrganisationResolverQueries {
+  constructor(private organisationService: OrganisationService) {}
+
+  @Query(() => [Organisation], {
+    nullable: false,
+    description: 'The Organisations on this platform',
+  })
+  @Profiling.api
+  async organisations(): Promise<IOrganisation[]> {
```
valentinyanakiev, line 16: Same as challenges - why are organizations not in the context of an ecoverse?

src/domain/community/user-group/user-group.entity.ts
```diff
@@ -16,7 +14,7 @@ import {
 } from 'typeorm';
 import { IUserGroup } from './user-group.interface';
 import { Profile } from '@domain/community/profile/profile.entity';
-import { Opportunity } from '@domain/challenge/opportunity/opportunity.entity';
+import { Community } from '../community';
```
valentinyanakiev, line 17: @domain...

valentinyanakiev (commented):
Functionally great ideas and improvements but largely inconsistent implementation:
1. Imports are all over the place. Some are absolute, some are relative, even when all changes are new.
2. Cascading changes on entities (especially Delete) is inconsistent everywhere which will lead to largely inconsistent CRUD (or just D) operations in the services. This will create problems.
3. (That's not an issue with this particular PR, just a continuation of a negative trend) CRUD operations continue to be largely copy/paste, instead of taking the time to resolve the issue structurally. Means more tests, more code paths, more bugs, higher maintenance. OK for prototyping but not beyond that. We should prioritise fixing this.
4. The way entities are removed is a smell. We continue removing the nested entities before we delete the containing one - this should be taken care of by the ORM really if used properly. Again not an issue coming from this PR per se, but one that needs to be prioritized (like point 3). If we had generic CRUD this code would be much shorter and easier to test.
5. Shouldn't organizations and challenges be in the context of an ecoverse if the update touches multitenancy?

src/common/exceptions/application.invalid.state.exception.ts
techsmyth: will update to be INVALID_STATE_TRANSITION_EXCEPTION

atstoyanov:
Will be useful to define the `community.type` as `CommunityType` and make the enum available to the GraphQL similar to the `ApplicationStatus`. This will make the API richer and easier to understand. Currently the `type` is just a string which is not very easy to understand what value to expect.

src/domain/community/community/community.entity.ts
```diff
@@ -0,0 +1,91 @@
...
+    { eager: false, cascade: false, onDelete: 'CASCADE' }
+  )
+  challenge?: Challenge;
+
+  @OneToOne(
+    () => Opportunity,
+    opportunity => opportunity.community,
+    { eager: false, cascade: false, onDelete: 'CASCADE' }
+  )
+  opportunity?: Opportunity;
```
atstoyanov: onDelete: 'CASCADE' must be only on the one side of a relation, otherwise will lead to cyclic reference.

src/domain/challenge/challenge/challenge.entity.ts
```diff
@@ -56,10 +52,21 @@ export class Challenge extends BaseEntity implements IChallenge, IGroupable {
...
+    nullable: true,
+    description: 'The community for the challenge',
+  })
+  @OneToOne(
+    () => Community,
+    community => community.challenge,
+    { eager: false, cascade: true, onDelete: 'CASCADE' }
+  )
+  @JoinColumn()
+  community?: Community;
```
atstoyanov, line 65: onDelete: 'CASCADE' must be only on the one side of a relation, otherwise will lead to cyclic reference. 
Deleting Challenge must delete community. Deleting a community should set the challenge communityID to null. or something similar.

src/domain/community/community/community.entity.ts
techsmyth: Fixed in https://github.com/cherrytwist/server/pull/821/commits/689c369cafab7d4253aa0b358830932b593ed16c

src/domain/challenge/challenge/challenge.entity.ts
techsmyth: Fixed in https://github.com/cherrytwist/server/pull/821/commits/689c369cafab7d4253aa0b358830932b593ed16c

Note that onDelete behavior needs a careful discussion - a lot of discussion on forums about this.

techsmyth:
> Will be useful to define the `community.type` as `CommunityType` and make the enum available to the GraphQL similar to the `ApplicationStatus`. This will make the API richer and easier to understand. Currently the `type` is just a string which is not very easy to understand what value to expect.

Not yet 100% sure we want to keep this field - wanted to discuss this after the client is updated to see if you actually need it client side. So for now suggest to leave as is + do a review once the client is updated to see how this is actually used.

valentinyanakiev (approved):

PR #830, opened 2021-03-19
Server 826

valentinyanakiev:
IPFS basic implementation spike. WIP - graphql file upload pending , DO NOT MERGE!.


src/services/ipfs/ipfs.resolver.ts
```diff
@@ -0,0 +1,19 @@
...
+import { IpfsService } from './ipfs.service';
+
+@Resolver()
+export class IpfsResolver {
+  constructor(@Inject(IpfsService) private ipfsService: IpfsService) {}
+
+  @Mutation(() => String, {
+    description: 'Uploads a file to IPFS endpoint',
+  })
+  async addFileToIpfs(
```
atstoyanov, line 13: In my opinion should be more generic. Like UploadFile/Image. Where is stored should be matter of configuration.

PR #851, opened 2021-03-26
Improved exception handling / feedback

valentinyanakiev:
The code is a bit speculative based on what is actually returned by JWT and not what I expected to see (an actual error).
Also there was an issue when there was no user profile the account was compared to the profile - the check whether profile is returned should be before that.

Please review.


techsmyth (approved):
This should do it - basicly first check user was authenticated.

PR #857, opened 2021-03-30
Add icon for the azure provider

atstoyanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

- I've read the contributing document https://github.com/cherrytwist/.github/blob/master/CONTRIBUTING.md.
- I've read and understand the Code of Conduct https://github.com/cherrytwist/.github/blob/master/CODE_OF_CONDUCT.md.
     https://github.com/cherrytwist/Coordination/blob/master/LICENSE.

### DIFF
--- a/src/services/configuration/config/config.service.ts
+++ b/src/services/configuration/config/config.service.ts
@@ -28,7 +28,7 @@ export class KonfigService {
       {
         name: 'Azure Active Directory',
         label: 'Log in with Azure AD',
-        icon: '',
+        icon: '/images/azure.svg',
         enabled: true,
         config: await this.getAadConfig(),
       },


src/services/configuration/config/config.service.ts
```diff
@@ -28,7 +28,7 @@ export class KonfigService {
       {
         name: 'Azure Active Directory',
         label: 'Log in with Azure AD',
-        icon: '',
+        icon: '/images/azure.svg',
```
techsmyth, line 31: this should be a full URL, not relative - it can be an image we have pinned on ipfs for example

techsmyth (changes requested):
The URL should be a full URL, not relative as we should not have this kind of image within our client.

PR #869, opened 2021-04-03
Application Lifecycles using XState

techsmyth:
Integration and usage of XState for lifecycle management, including moving to a more reactive / event driven api.

Explicit management of application lifecycles with states
Adds Lifecycle entity to domain model, with the current state and next states
adds new mutation: eventOnApplication 
Integration of commands, wrapped in an OptionsProvider, that allows the hooks back into the services to e.g. add the user to the community once the application is approved. 

To do: tests around application approvals will need updating?


src/domain/common/lifecycle/lifecycle.dto.transition.ts
```diff
@@ -0,0 +1,14 @@
+import { InputType, Field } from '@nestjs/graphql';
+import { IsOptional, MaxLength } from 'class-validator';
+import { SMALL_TEXT_LENGTH } from '@src/common/constants';
+
+@InputType()
+export class LifecycleEventInput {
```
valentinyanakiev: lifecycle.dto.transition --> LifecycleEventInput - was that intentional?

src/domain/common/lifecycle/lifecycle.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+  state: string;
+
+  // Stores the xstate engine definition
+  @Column('text', { nullable: true })
+  machine: string;
+
+  constructor(machine: any) {
+    super();
+    this.machine = machine;
+    this.state = '';
```
valentinyanakiev: It might be better to keep the state field undefined if it wasn't initialized.

src/domain/common/lifecycle/lifecycle.service.ts
```diff
@@ -0,0 +1,125 @@
...
+        `Unable to update state: provided event (${eventName}) not in valid set of next events: ${nextStates}`,
+        LogContext.LIFECYCLE
+      );
+    }
+
+    const machineService = interpret(machineWithContext).start(restoredState);
+
+    machineService.send({
+      type: eventName,
+      parentID: machineDef.context.parentID,
```
valentinyanakiev: For discussion - are we mandating the parentID?

src/migrations/1618911977800-lifecycle-application.ts
```diff
@@ -0,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
valentinyanakiev: I suggest we merge all 4 migrations into one.

valentinyanakiev (changes requested):
Merge migrations into 1 and a couple of minor discussion points.

src/domain/common/lifecycle/lifecycle.dto.transition.ts
techsmyth: Good catch

src/domain/common/lifecycle/lifecycle.entity.ts
techsmyth: Agreed

src/domain/common/lifecycle/lifecycle.service.ts
techsmyth: Yes lets discuss

src/migrations/1618911977800-lifecycle-application.ts
techsmyth: Agreed, done

valentinyanakiev (approved):

PR #873, opened 2021-04-06
I - create/Update mutations to take a single DTO input

techsmyth:
This PR also moves several ID references to be strings, not all but a start.

References in particular need to be tested - the behavious there has changed in that the client needs to send up a list of create and a list of updates.

api also cleaned up in places.


valentinyanakiev (dismissed):
We are not on the same page about this - the goal is not to move the IDs to the InputTypes (DTOs), the goal is to have a single payload object per GraphQL mutation - and the majority of mutations updated fail in that aspect.
For reference: https://docs.github.com/en/graphql/reference/mutations
All mutations have single payload / DTO / input - that's what we are striving to achieve. That way we have unique set of objects that can be used for validation as well.

techsmyth:
> We are not on the same page about this - the goal is not to move the IDs to the InputTypes (DTOs), the goal is to have a single payload object per GraphQL mutation - and the majority of mutations updated fail in that aspect.
> For reference: https://docs.github.com/en/graphql/reference/mutations
> All mutations have single payload / DTO / input - that's what we are striving to achieve. That way we have unique set of objects that can be used for validation as well.

as discussed, was still workign through the remaining mutations!

Comoque1:
Querying specific group, doesn't return results
```
query {
  ecoverse {
    group(ID: "1") {
      id
      name
      focalPoint {
        name
      }
      members {
        name
        id
      }
    }
  }
}

```
image
image

techsmyth:
> Querying specific group, doesn't return results
> 
> ```
> query {
>   ecoverse {
>     group(ID: "1") {
>       id
>       name
>       focalPoint {
>         name
>       }
>       members {
>         name
>         id
>       }
>     }
>   }
> }
> ```
> 
> image
> image

this is fixed with the commit 17:54

Comoque1:
Performing `challenge.context.updateReference` is not updating the reference data

```
mutation UpdateChallenge($challengeData: UpdateChallengeInput!) {
  updateChallenge(challengeData: $challengeData) {
    name,
    id
    context{id references{id description}}
  }
}
```
```
{
  "challengeData":
		{
  	  "ID": "1",
      "name": "k",
      "context": {"ID": "2",       
        "updateReferences": {"ID": 1, "description": "test"}
      }
   }
}
```

image
image

valentinyanakiev (commented):
Several comments:

1. The idea of the DTOs is to have 1 specific DTO per mutation as they are used for 2 reasons:
- optimized payload - that's OK with a generic DTO
- data validation - the contract for the DTO payload parameters might be different and there might be different middlewares handling different object type validation / transformation / mapping. Also documentation might need to be specific for different mutation / mutation input types (https://docs.github.com/en/graphql/reference/input-objects#removeenterprisesupportentitlementinput)
Currently, all remove operations share the same DTO object and while that looks like a good design (DRY) it will likely lead to rework / inconsistent validation approaches.
2. addUserToCommunity, addUserToGroup, assignGroupFocalPoint, removeUserFromCommunity, removeUserFromGroup share the same input type (UpdateMembershipInput). These might be candidates for an interface.
3. addChallengeLead, removeChallengeLead, removeGroupFocalPoint, approveApplication - all need DTOs
4. All xxxOnyyy mutations should be moved to interfaces - at the moment the API signatures look exactly the same.

atstoyanov:
> * addUserToCommunity, addUserToGroup, assignGroupFocalPoint, removeUserFromCommunity, removeUserFromGroup share the same input type (UpdateMembershipInput). These might be candidates for an interface.

Creating interface will result in the same field names. I prefer to have types like:

``` javascript
AddUserToCommunityInput: {
   userId // The id of the user which will be added to the community.
   communityId // The id of the community to which the user will be added.
}

AddUserToGroupInput: {
   userId,    // The id of the user which will be added to the group.
   groupId  // The id of the group to which the user will be added.
}
```
So when somebody is using the API know exactly what information is required to provide.

Also the schema could have description for each field. (aka Documentation)
image


``` javascript
export type UpdateMembershipInput = {
  childID: Scalars['Float'];
  parentID: Scalars['Float'];
};
```
Many IDs were changed to `String`. These two are still `Number`.

valentinyanakiev (approved):

PR #878, opened 2021-04-07
Upload avatars + Docker Compose

valentinyanakiev:
Upload avatars logic + validation.

Added new common folder - utils. File and image utils are inside it.
Removed generic upload file mutation and the ipfs resolver - keeping the service methods intact, but the public API has been removed.
Moved the bulk of the logic in the ProfileService instead of the resolver. 

To test it (easiest) --> Change the docker-compose server port mapping to 4000:4000 and run the client, log in, update your profile and add an avatar. If all is correctly implemented the avatar should be visible.


.env.demo.auth.provider.default
```diff
@@ -1,3 +1,3 @@
-AUTH_DEMO_AUTH_ISSUER=http://localhost:3002
-AUTH_DEMO_AUTH_TOKEN_ENDPOINT=http://localhost:3002/auth/login
-AUTH_DEMO_AUTH_CLIENT_SECRET=!#@$#SDFRSF
\ No newline at end of file
+AUTH_DEMO_AUTH_ISSUER=http://localhost:3003
+AUTH_DEMO_AUTH_TOKEN_ENDPOINT=http://localhost:3003/auth/login
+AUTH_DEMO_AUTH_CLIENT_SECRET=12345678aAbB!
```
atstoyanov: Are this the correct ones?
The AuthProvider default port is 3002, and the secret is the previous one.

docker-compose.yml
```diff
@@ -63,9 +66,23 @@ services:
...
 
+  ct_ipfs_server:
+    container_name: ct_ipfs_server
+    image: ipfs/go-ipfs
+    networks:
+      - ct_server_dev
+    restart: always
+    ports:
+      - 4001:4001/tcp
+      - 4002:4002/udp
```
atstoyanov, line 77: `4002` is mapped to the `ct-server`, isn't it?

docker-compose.yml
```diff
@@ -77,7 +94,7 @@ services:
       - WAIT_SLEEP_INTERVAL
       - WAIT_HOST_CONNECT_TIMEOUT
       - SERVICE_ENDPOINT_PORT
-      - SECRET=AUTH_DEMO_AUTH_CLIENT_SECRET
+      - SECRET=12345678aAbB!
```
atstoyanov: Why it is repeated?

src/domain/community/profile/profile.service.ts
```diff
@@ -171,4 +177,53 @@ export class ProfileService {
...
+
+    if (!readStream)
+      throw new ValidationException(
+        'Readstream should be defined!',
+        LogContext.COMMUNITY
+      );
+
+    const buffer = await streamToBuffer(readStream);
+
+    if (!validateImageDimensions(buffer, 390, 410))
```
atstoyanov: Too narrow. The user should send an image with almost exact dimensions. It will be better if the user is allowed to upload image with dimension between 200x200 and 400x400. The image will be shown in a square around 80x80 so 200x200 is quite enough.

src/services/ipfs/ipfs.service.ts
```diff
@@ -16,4 +18,20 @@ export class IpfsService {
...
   }
+
+  public async uploadFileFromStream(stream: ReadStream): Promise<string> {
+    const buffer = await streamToBuffer(stream);
+    return await this.uploadFileFromBuffer(buffer);
+  }
+
+  public async uploadFileFromBuffer(buffer: Buffer): Promise<string> {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const IpfsHttpClient = require('ipfs-http-client');
```
atstoyanov: Is it possible to import it like typescript module?
Track ipfs/js-ipfs#2945

atstoyanov (changes requested):

.env.demo.auth.provider.default
valentinyanakiev: There are 4 different values - in demo project, 2 in server (1 in docker-compose, 1 in .env files) and 1 in the demo-auth provider. Which one should I keep?

docker-compose.yml
valentinyanakiev: Repeated how?

src/domain/community/profile/profile.service.ts
valentinyanakiev: That was in the AC we all agree on during the refinement...

atstoyanov (approved):

PR #886, opened 2021-04-10
II - Mutations api refresh; DTO's unique 

techsmyth:


src/domain/community/organisation/organisation.resolver.mutations.ts
```diff
@@ -68,4 +68,15 @@ export class OrganisationResolverMutations {
     );
     return org;
   }
+
+  @Roles(AuthorizationRoles.EcoverseAdmins)
```
valentinyanakiev, line 72: Discussion: EcoverseAdmin should be able to delete only orgs in his eco verse context.

src/domain/community/organisation/organisation.service.ts
```diff
@@ -44,7 +46,10 @@ export class OrganisationService {
     return organisation;
   }
 
-  async initialiseMembers(organisation: IOrganisation): Promise<IOrganisation> {
+  async initialiseMembers(
+    organisation: IOrganisation,
```
valentinyanakiev: Shouldn't CreateOrganisationInput as DTO contain all data needed to create an organisation? What does the first parameter bring?

src/domain/community/profile/profile.resolver.mutations.ts
```diff
@@ -1,79 +1,71 @@
...
 
   @Roles(AuthorizationRoles.EcoverseAdmins, AuthorizationRoles.CommunityAdmins)
   @SelfManagement()
-  @Mutation(() => Profile)
+  @Mutation(() => Profile, {
+    description: 'Uploads and sets an avatar image for the specified Profile.',
+  })
   async uploadAvatar(
-    @Args('profileID') profileID: number,
+    @Args('uploadData') uploadData: UploadProfileAvatarInput,
```
valentinyanakiev, line 68: I think I misled you here - we can put the GraphQLUpload inside the UploadProfileAvatarInput

src/domain/community/user-group/user-group.dto.remove.member.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, InputType } from '@nestjs/graphql';
+
+@InputType()
+export class RemoveUserGroupMemberInput {
+  @Field({ nullable: false })
+  parentID!: number;
```
valentinyanakiev: This is really super generic - needs to be discussed

src/domain/community/user/user.service.ts
```diff
@@ -28,46 +32,39 @@ export class UserService {
...
     );
 
-    return populatedUser;
+    return savedUser;
   }
 
   // Helper method to ensure all members that are arrays are initialised properly.
   // Note: has to be a seprate call due to restrictions from ORM.
-  async initialiseMembers(user: IUser): Promise<IUser> {
+  async initialiseMembers(
```
valentinyanakiev: I think it will be cleaner if you have one method where you create the user from the UserData, not create the user from the repo method outside the method and pass it next to the UserData

valentinyanakiev (commented):
Some minor comments up for discussion but overall looks like a huge update in the right direction :)

src/domain/community/organisation/organisation.resolver.mutations.ts
techsmyth: Interesting one, and has the same implication for Users - it may be that Community Admin is not a subset of EcoverseAdmin. Authorisation in general needs a discussion re multi-ecoverse support.

src/domain/community/organisation/organisation.service.ts
techsmyth: Actually realised we do not need this split with initialiseMembers any more...merged into one function

src/domain/community/profile/profile.resolver.mutations.ts
techsmyth: Had tried but was having an issue - can you give it a go?

src/domain/community/user-group/user-group.dto.remove.member.ts
techsmyth: Ok

src/domain/community/user/user.service.ts
techsmyth: Agreed - see comment above about this split not being needed.

Comoque1:
Group is not created, when create a new challenge.
image

valentinyanakiev (approved):

PR #889, opened 2021-04-10
III - improved multi-ecoverse support

techsmyth:
Prepares code for multiple ecoverse support:
- reduces all getDefaultEcoverse to only be in root query or in bootstrap
- updates opportunities to do recursive lookup instead of db lookup
- updates adding users to roles to be in context of an ecoverse

### DIFF
--- a/src/core/bootstrap/bootstrap.service.ts
+++ b/src/core/bootstrap/bootstrap.service.ts
@@ -183,7 +183,7 @@ export class BootstrapService {
 
   @Profiling.api
   async createGroupProfiles(groupName: string, usersData: any[]) {
-    const defaultEcoverse = await this.ecoverseService.getFirstEcoverseOrFail();
+    const defaultEcoverse = await this.ecoverseService.getDefaultEcoverseOrFail();
     try {
       for (const userData of usersData) {
         const userInput = new CreateUserInput();
--- a/src/core/bootstrap/bootstrap.service.ts
+++ b/src/core/bootstrap/bootstrap.service.ts
@@ -246,7 +246,7 @@ export class BootstrapService {
     }
   }
 
-  async ensureEcoverseSingleton(): Promise<IEcoverse> {
+  async ensureEcoverseSingleton() {
     this.logger.verbose?.(
       '=== Ensuring single ecoverse is present ===',
       LogContext.BOOTSTRAP
@@ -259,34 +259,33 @@ export class BootstrapService {
       this.logger.verbose?.('...No ecoverse present...', LogContext.BOOTSTRAP);
       this.logger.verbose?.('........creating...', LogContext.BOOTSTRAP);
       // Create a new ecoverse
-      const ecoverse = new Ecoverse();
-      await this.ecoverseService.initialiseMembers(ecoverse);
-      // Save is needed so that the ecoverse is there for other methods
-      await this.ecoverseRepository.save(ecoverse);
+      const ecoverse = await this.ecoverseService.createEcoverse({
+        textID: 'Eco1',
+        name: 'Empty ecoverse',
+      });
 
       this.logger.verbose?.('........populating...', LogContext.BOOTSTRAP);
       await this.populateEmptyEcoverse(ecoverse);
       await this.ecoverseRepository.save(ecoverse);
-      return ecoverse;
+      return ecoverseArray[0];
     }
     if (ecoverseCount == 1) {
       this.logger.verbose?.(
         '...single ecoverse - verified',
         LogContext.BOOTSTRAP
       );
-      return ecoverseArray[0];
     }
-
-    throw new ValidationException(
-      'Cannot have more than one ecoverse',
-      LogContext.BOOTSTRAP
-    );
+    if (ecoverseCount > 1) {
+      this.logger.warn?.(
+        `...multiple ecoverses detected: ${ecoverseCount}`,
+        LogContext.BOOTSTRAP
+      );
+    }
   }
 
   // Populate an empty ecoverse
   async populateEmptyEcoverse(ecoverse: IEcoverse): Promise<IEcoverse> {
     // Set the default values
-    ecoverse.name = 'Empty ecoverse';
     if (!ecoverse.context)
       throw new EntityNotInitializedException(
         'Non-initialised ecoverse',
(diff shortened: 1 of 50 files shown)


src/core/bootstrap/bootstrap.service.ts
```diff
@@ -184,6 +183,7 @@ export class BootstrapService {
 
   @Profiling.api
   async createGroupProfiles(groupName: string, usersData: any[]) {
+    const defaultEcoverse = await this.ecoverseService.getFirstEcoverseOrFail();
```
valentinyanakiev: getDefaultEcoverse might be a better name

src/core/validation/constraints/unique.text.id.ts
```diff
@@ -25,7 +27,13 @@ export enum TextIdType {
 export class IsUniqueTextIdConstraint implements ValidatorConstraintInterface {
   async validate(textId: any, args: ValidationArguments): Promise<boolean> {
     const [target] = args.constraints;
-    if (target === TextIdType.challenge) {
```
valentinyanakiev, line 28: This is OK and will work but will look better with a switch as it's likely that this check is extended in the future with more type checks.

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -75,6 +86,93 @@ export class ChallengeService {
...
   }
 
+  async updateChallenge(
+    challengeData: UpdateChallengeInput
+  ): Promise<IChallenge> {
+    const challenge = await this.getChallengeOrFail(challengeData.ID);
+
+    const newName = challengeData.name;
+    if (newName) {
+      if (!(newName === challenge.name)) {
```
valentinyanakiev: Style-wise this is an odd check - might be more readable with with !== instead of !(... === ...)

src/domain/challenge/ecoverse/ecoverse.dto.create.ts
```diff
@@ -0,0 +1,45 @@
...
+//   TextIdType,
+// } from '@src/core/validation/constraints/unique.text.id';
+
+@InputType()
+export class CreateEcoverseInput {
+  @Field({
+    nullable: false,
+    description: 'The unique text ID for the ecoverse',
+  })
+  // @IsUniqueTextId(TextIdType.ecoverse, {
```
valentinyanakiev, line 16: Why is this commented out?

src/domain/challenge/ecoverse/ecoverse.resolver.mutations.ts
```diff
@@ -6,18 +6,33 @@ import { Resolver, Args, Mutation } from '@nestjs/graphql';
...
+  UpdateEcoverseInput,
+} from '@domain/challenge/ecoverse';
 
 @Resolver()
 export class EcoverseResolverMutations {
   constructor(
     @Inject(EcoverseService) private ecoverseService: EcoverseService
   ) {}
 
+  @Roles(AuthorizationRoles.EcoverseAdmins)
```
valentinyanakiev: I don't think EcoverseAdmin should be able to create an Ecoverse - his scope is inside the Ecovse, not on the 'Multiverse' scope - only Global Admin should have authority there in my opinion.

src/domain/challenge/ecoverse/ecoverse.service.ts
```diff
@@ -113,37 +139,33 @@ export class EcoverseService {
...
-      .getOne(); // TODO [ATS] Replace with getOneOrFail when it is released. https://github.com/typeorm/typeorm/blob/06903d1c914e8082620dbf16551caa302862d328/src/query-builder/SelectQueryBuilder.ts#L1112
-
-    if (!ecoverse) {
-      throw new ValidationException(
-        'Ecoverse is missing!',
-        LogContext.BOOTSTRAP
+  async getOpportunities(ecoverse: IEcoverse): Promise<IOpportunity[]> {
+    const opportunities: IOpportunity[] = [];
+    const challenges = await this.getChallenges(ecoverse);
+    for (let i = 0; i < challenges.length; i++) {
```
valentinyanakiev: for(const entity of entities)... --> like everywhere else in the codebase

src/domain/challenge/ecoverse/ecoverse.service.ts
```diff
@@ -272,4 +290,26 @@ export class EcoverseService {
 
     return ecoverse;
   }
+
+  async getFirstEcoverseOrFail(
```
valentinyanakiev: getDefaultEcoverseOrFail

valentinyanakiev (changes requested):
Some minor comments:
- getDefaultEcoverseOrFail is standard notation, getFirstEcoverseOrFail is not better
- some code-style (minor) changes will improve the code
- there are some code pieces left commented out - why?
- mutations on Ecoverse level should need global admin privileges

src/core/bootstrap/bootstrap.service.ts
techsmyth: It was getDefaultEcoverse - but on reflection we are actually taking the first ecoverse, so I updated it. Can chat it through...

src/core/validation/constraints/unique.text.id.ts
techsmyth: This decorator was giving problems again - same as when adding it to Organisation. See comment in api doc about switching to using a dedicated type for this...so think this class will go.

src/domain/challenge/challenge/challenge.service.ts
techsmyth: Will fix.

src/domain/challenge/ecoverse/ecoverse.dto.create.ts
techsmyth: See above comment - this decorator was breaking the generation / loading. And think will replace with custom Scalar so did not fight it.

src/domain/challenge/ecoverse/ecoverse.resolver.mutations.ts
techsmyth: Agree.

src/domain/challenge/ecoverse/ecoverse.service.ts
techsmyth: Agreed. Still default to the old style at times :)

src/domain/challenge/ecoverse/ecoverse.service.ts
techsmyth: Ditto to first comment above...

valentinyanakiev (approved):

PR #904, opened 2021-04-15
extended self mgmt to also be able to check the incoming userData ID matches the ID of the logged in user

techsmyth:
First issue that triggered the bug: self mgmt had an undefined as the incoming updateUser DTO no longer supplied the email address (as not needed). This lead to an undefined field which caused the seen error.

Bigger issue: self-management was very implicit and hard-coded to a field on the incoming DTO. The recent DTO updates exposed this fragility.

The logic has been reworked to:
- store the relevant incoming DTO in all cases when the SelfManagement authorisation decorator is used
- check using logic per mutation how the data in the stored DTO matches that for the incoming request
Obviously a bad smell to have the logic like this, but at least it is now an explicit and extendable bad smell while we sort out a wider authorisation model update. 

As a bonus also allow now users to self-manage their own profileData via updateProfile.

Likely extension: allowing users to update references and tagsets on their own profile.

### DIFF
--- a/src/core/authorization/graphql.guard.ts
+++ b/src/core/authorization/graphql.guard.ts
@@ -39,6 +39,14 @@ export class GqlAuthGuard extends AuthGuard(['azure-ad', 'demo-auth-jwt']) {
     this._selfManagement = v;
   }
 
+  private _userID!: number;
+  public get userID(): number {
+    return this._userID;
+  }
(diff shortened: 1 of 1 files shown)


valentinyanakiev (approved):
Approving with the caveat that we are moving to rights-based access control from RBAC.

PR #915, opened 2021-04-22
Added lifecycle support to challenge, opportunity, project

techsmyth:
Challenges, Opportunities, Projects now have a Lifecycle.

Challenge, Opportunity can be instantiated with two different lifecycles: default, EXTENDED.


src/common/enums/challenge.lifecycle.templates.ts
```diff
@@ -0,0 +1,3 @@
+export enum ChallengeLifecycleTemplates {
+  EXTENDED = 'EXTENDED',
```
valentinyanakiev, line 2: I am assuming extended will be common for all / mostly all templates - it might make sense to have lifecycle.templates.ts with LifecycleTemplates enum (essentially a 'common' for LifecycleTemplates) and have EXTENDED there.

src/domain/challenge/challenge/challenge.resolver.mutations.ts
```diff
@@ -90,4 +94,18 @@ export class ChallengeResolverMutations {
...
+  @Roles(AuthorizationRoles.EcoverseAdmins, AuthorizationRoles.GlobalAdmins)
+  @UseGuards(GqlAuthGuard)
+  @Mutation(() => Challenge, {
+    description: 'Trigger an event on the Challenge.',
+  })
+  async eventOnChallenge(
+    @Args('challengeEventData')
+    challengeEventData: ChallengeLifecycleEventInput
+  ): Promise<IChallenge> {
+    return await this.challengeLifecycleOptionsProvider.eventOnChallenge(
```
valentinyanakiev, line 107: It might make sense to move eventOnChallenge (as a pattern) on the level of the root service (challengeService in this case).

src/domain/challenge/opportunity/opportunity.lifecycle.options.provider.ts
```diff
@@ -0,0 +1,62 @@
...
+@Injectable()
+export class OpportunityLifecycleOptionsProvider {
+  constructor(
+    private lifecycleService: LifecycleService,
+    private opportunityService: OpportunityService,
+    @Inject(WINSTON_MODULE_NEST_PROVIDER) private readonly logger: LoggerService
+  ) {}
+
+  async eventOnOpportunity(
+    opportunityEventData: OpportunityLifecycleEventInput
```
valentinyanakiev: For discussion - I feel this (and again as a pattern) should be OpportunityEventInput. The event is on opportunity and is being used by the lifecycle, but logically it's an opportunity event, not opportunityLifecycle event. The lifecycle is essentially a stateMachine consuming events of the underlying entities. Moving that makes explicit that that event is also on the level of the entity and not the lifecycle (in my opinion).

valentinyanakiev (changes requested):
1. There are 89 failing tests (Evgeni is working on them)
2. There are 2-3 patterns up for discussion (comments inline, didn't leave them on every occurrence as it makes no sense).

The rest looks OK.

Comoque1:
CreateApplication mutation is failing

image

src/common/enums/challenge.lifecycle.templates.ts
techsmyth: Actually I expect we could have many templates - not just a few. So we need later to have some kind of registration / factory setup for lifecycle definitions. To be brainstormed....there are two for now to illustrate the concept.

src/domain/challenge/challenge/challenge.resolver.mutations.ts
techsmyth: Agreed. Cleaner - then the options provider is only for the options.

src/domain/challenge/opportunity/opportunity.lifecycle.options.provider.ts
techsmyth: Agreed, ties in with the actual mutation name then.

techsmyth:
> CreateApplication mutation is failing
> 
> image

bug on createApplicaiton is fixed with latest commit

valentinyanakiev (approved):

PR #943, opened 2021-04-30
updated self management to work with additional mutations

techsmyth:

### DIFF
--- a/src/core/authorization/graphql.guard.ts
+++ b/src/core/authorization/graphql.guard.ts
@@ -80,7 +80,6 @@ export class GqlAuthGuard extends AuthGuard(['azure-ad', 'demo-auth-jwt']) {
 
       // Failsafe: if decorator SelfManagement was used then a DTO must have been set
       if (!this.mutationDTO) {
-        console.log(`${this.mutationDTO}`);
         throw new ForbiddenException(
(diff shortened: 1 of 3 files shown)


src/core/authorization/graphql.guard.ts
```diff
@@ -71,12 +71,21 @@ export class GqlAuthGuard extends AuthGuard(['azure-ad', 'demo-auth-jwt']) {
...
+      if (args.referenceInput) this.mutationDTO = args.referenceInput;
+      // Mutation: createTagsetOnProfile
+      if (args.tagsetData) this.mutationDTO = args.tagsetData;
+      // Mutation: deleteReference
+      if (args.deleteData) this.mutationDTO = args.deleteData;
+
       // Failsafe: if decorator SelfManagement was used then a DTO must have been set
-      if (!this.mutationDTO)
+      if (!this.mutationDTO) {
+        console.log(`${this.mutationDTO}`);
```
valentinyanakiev: remove console.log

valentinyanakiev (changes requested):
Remove the console.log. The rest is a terrible extension of the hack but let's fix it altogether with the auth...

valentinyanakiev (approved):

PR #965, opened 2021-05-13
Domain model refactoring

techsmyth:
Major restructuring of domain model relationships.

New entities:
- Collaboration, to hold Projects and Relations
- EcosystemModel, to hold a representation of an ecosystem model. For now this holds only actors / actor groups but later should be expanded to include value types, value exchanges etc.

Challenges are now hierarchical, in that a Challenge can have child challenges. The Opportunity entity has been removed. Also the Ecoverse entity now has a direct Challenge member that provides most of its functionality.

Context entity has been expanded, and now contains both an EcosystemModel as well as Aspects. 

Multiple ecoverse support: this PR also does another big set of groundwork towards multiple ecoverse support. In particular, entities that are within the context of a particular ecoverse, they now have an additional field to capture the ID of the containing ecoverse. This is used to get for example all Projects contained within a single Ecoverse. 
- all ecoverse queries are updated to be able to query within the context of an ecoverse. 

UserGroup service has been simplified, removing logic that was no longer needed.

"activity" metrics are now directly available via a new field on ecoverse + Challenge
- this is still in progress but should be able to speed up the client loading as server does not need to pull all the data into memory.

Note: given the severity of the restructuring, the database definition has again been reset. 

Note: both client-lib and populator have branches / draft PRs that work against this PR

To do:
- isUniqueTextID decorator is not working
- tests, again...
- domain model review...
- docs


atstoyanov:
Running migrations are failing: `Error: ER_TABLE_EXISTS_ERROR: Table 'query-result-cache' already exists`

Comoque1:
"Project" cannot be created.

Repro steps:
1. Create Challenge
2. Create ChildChallenge
3. CreateProject - request fails
`"message": "ER_NO_DEFAULT_FOR_FIELD: Field 'ecoverseID' doesn't have a default value"`

image

Comoque1:
Reference cannot be created as part of createChallenge gql request.
image

techsmyth:
@Comoque1 re bugs raised:
- just fixed the references on challenge creation: 82f8cd4604e3048cb66a17d85e71721163d1d84b
- the project creation bug: please note this and pick it up when we merge in the domain-model-plus branch

atstoyanov:
When trying to create new Project  `ER_NO_DEFAULT_FOR_FIELD: Field 'ecoverseID' doesn't have a default value`.

atstoyanov:
> When trying to create new Project `ER_NO_DEFAULT_FOR_FIELD: Field 'ecoverseID' doesn't have a default value`.

Maybe the problem is in my machine, because I'm passing as `parentId` the `opportunityId` (`challengeID`) instead of `collaborationID`.

So the `parentId` is not informative enough for the developer to understand what value must be passed. Either change the field to `collaborationId` (best variant) or add description what is expected (worst).

src/app.module.ts
```diff
@@ -39,7 +39,7 @@ import { MembershipModule } from './services/membership/membership.module';
       useFactory: async (configService: ConfigService) => ({
         type: 'mysql',
         insecureAuth: true,
-        synchronize: false,
```
valentinyanakiev: should be false

atstoyanov:
TypeOrm Error: `Relation "ecoverse" was not found; please check if it is correct and really exists in your entity.`

src/domain/collaboration/project/project.resolver.fields.ts
```diff
@@ -1,20 +1,20 @@
...
+import { ILifecycle } from '@domain/common/lifecycle';
+import { IProject, Project } from '@domain/collaboration/project';
 
-@Resolver(() => Project)
+@Resolver(() => IProject)
 export class ProjectResolverFields {
   constructor(private projectService: ProjectService) {}
 
-  @ResolveField('lifecycle', () => Lifecycle, {
+  @ResolveField('lifecycle2', () => ILifecycle, {
```
atstoyanov: Why `lifecycle2` after every other entity has property `lifecycle`?

src/common/enums/logging.context.ts
```diff
@@ -1,12 +1,14 @@
 export enum LogContext {
+  CHALLENGES = 'challenges',
+  CONTEXT = 'context',
   COMMUNITY = 'community',
+  COLLABORATION = 'collaboration',
```
valentinyanakiev, line 5: collaboration should be opportunity

atstoyanov:
`Activity` is missing from `Opportunity`

src/domain/common/lifecycle/lifecycle.interface.ts
```diff
@@ -1,5 +1,14 @@
-export interface ILifecycle {
-  id: number;
+import { IBaseCherrytwist } from '@domain/common/base-entity';
+import { Field, ObjectType } from '@nestjs/graphql';
+import JSON from 'graphql-type-json';
+
+@ObjectType('Lifecycle2')
```
valentinyanakiev: Rename 'Lifecycle2'

valentinyanakiev (changes requested):
Still a few minor tweaks / changes required, comments inline.

src/domain/collaboration/project/project.resolver.fields.ts
techsmyth: Agreed

src/common/enums/logging.context.ts
techsmyth: disagree, opportunity is part of collaboration

valentinyanakiev (approved):

PR #971, opened 2021-05-14
fixed bug with console logging; updated bootstrap to log config

techsmyth:
The logging config is now also logged on platform boostrap


valentinyanakiev (approved):
Enums need attention - tracked in https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/server/977.

Otherwise looks good - we can add config validation at a later stage but this is out-of-scope for this PR.

PR #973, opened 2021-05-15
cleaned up branch for membership

techsmyth:
provides a new top level query to allow finding the memberships for a given user.


src/services/membership/membership.dto.result.ecoverse.entry.ts
```diff
@@ -0,0 +1,27 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+import { MembershipResultEntry } from './membership.dto.result.entry';
+
+@ObjectType()
+export class MembershipEcoverseResultEntry {
```
valentinyanakiev, line 5: EcoverseMembershipPayload might be a better name for the given class in my opinion.

src/services/membership/membership.dto.result.ecoverse.entry.ts
```diff
@@ -0,0 +1,27 @@
...
+  @Field(() => String, {
+    nullable: true,
+    description: 'The name of the Ecoverse.',
+  })
+  name = '';
+
+  @Field(() => String, {
```
valentinyanakiev, line 12: The bulk of the IDs we have (excluding TextIDs) are @Field(() => ID) mapped to number in JS.

src/services/membership/membership.dto.result.entry.ts
```diff
@@ -0,0 +1,19 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+
+@ObjectType()
+export class MembershipResultEntry {
```
valentinyanakiev, line 4: MembershipPayload?

src/services/membership/membership.service.ts
```diff
@@ -0,0 +1,108 @@
...
+  async membership(membershipData: MembershipInput): Promise<Membership> {
+    const membership = new Membership();
+    const user = await this.userService.getUserWithAgent(membershipData.userID);
+    const credentials = user.agent?.credentials;
+    if (!credentials) {
+      return membership;
+    }
+    const storedChallenges: IChallenge[] = [];
+    const storedUserGroups: IUserGroup[] = [];
+    for (const credential of credentials) {
```
valentinyanakiev, line 36: Design-wise, we should consider adding a Membership name (I don't have a good name for it) that is implemented by the services and has a method that receives a resourceID. A lot of the code below will be simplified and instead of getting an organization and then resolving its membership, we can resolve the membership by passing the resourceID.

valentinyanakiev (commented):
Overall feels mostly OK, but there are a couple of discussion points.

src/services/membership/membership.dto.result.ecoverse.entry.ts
techsmyth: left this as generic string for now as need to pick up IDs systematically with global ids epic...

src/services/membership/membership.service.ts
techsmyth: agreed that this can be more efficient. 
already realised when doing it that we need to carefully look again at the set of credentials e.g. community member is not right, it should be ecoverse member + challenge member. 
Also ties into the notion of authorisation scopes that we talked about last week.

So not pick up here, but have added a note to https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/cherrytwist/234 so that we pick it up with that phase 2 authorisation work.

src/services/membership/membership.dto.result.ecoverse.entry.ts
techsmyth: Took this pattern from the search service - happy to have another name but lets land it quick.

src/services/membership/membership.dto.result.entry.ts
techsmyth: ditto to first comment above.

src/services/membership/membership.service.ts
valentinyanakiev: I see nothing in this epic reflecting my suggestion. Agreed that we might postpone some fixes to be done in the grand schema of things but we should at the very least capture our thoughts. Let's chat this through.

valentinyanakiev (approved):

PR #976, opened 2021-05-15
Config docs

valentinyanakiev:
- Updated config docs - inline in the cherrytwist.yaml. 
- Removed config explanations from README.MD and the Configuration.MD
- Made small adjustments to the config file

Couldn't make an inline reference from the yaml to the MDs the way the MDs have between each other so kept more details in the AAD section - please review whether it makes sense.


techsmyth (approved):
Great to have this documented, makes the platform much easier to on-board with.

PR #984, opened 2021-05-19
Domain model plus refactoring

techsmyth:
In adding back the Opportunity entity, but derived from a BaseChallenge entity, it became apparent that we could substantially clean up the code base.

The api should be very similar, primary change would be that some number fields became strings due to sharing a common definition, only good for consistency in the api.

Note: given the scope of the changes there does need to be a good regression test done.

GraphQL / TypeORM logic split
=======================
Entities are now only used for TypeORM and business logic. 
All graphql api definitions are now in the interface classes. 

Code changes
==========
New BaseCherrytwistEntity to hold common field definitions for TypeORM 
- this holds all columns that are common across all the entities, including typeorm fields like version etc.

New IBaseCherrytwist to hold common field definitions for GraphQL

New interface types:
- Searchable to return entities that are searchable
- Groupable to return the parent of a user group
Note: these new interface types were forced as creating union types was not working with abstract classes, but is anyway a cleaner design.

Opportunity entity
=============
Challenges can now have child Opportunities.

Added in mutations to create, update, delete opportunities. 
Updated ecoverse field opportunities to return the opportunities one level down.

DTO Inheritance
============
Update / delete DTOs to be derived from a base class


src/app.module.ts
```diff
@@ -39,7 +39,7 @@ import { MembershipModule } from './services/membership/membership.module';
       useFactory: async (configService: ConfigService) => ({
         type: 'mysql',
         insecureAuth: true,
-        synchronize: false,
+        synchronize: true,
```
valentinyanakiev, line 42: Migration + synchronize: false

src/domain/common/base-entity/base.cherrytwist.interface.ts
```diff
@@ -0,0 +1,14 @@
...
+@ObjectType('IBaseCherrytwist')
+export abstract class IBaseCherrytwist {
+  @Field(() => ID, {
+    nullable: false,
+    description: 'The ID of the entity',
+  })
+  id!: number;
+
+  constructor() {
+    this.id = -1;
```
valentinyanakiev, line 12: This is probably redundant but it will be replaced by guid / uuid anyway...

src/domain/common/lifecycle/lifecycle.interface.ts
```diff
@@ -1,5 +1,14 @@
-export interface ILifecycle {
-  id: number;
+import { IBaseCherrytwist } from '@domain/common/base-entity';
+import { Field, ObjectType } from '@nestjs/graphql';
+import JSON from 'graphql-type-json';
+
+@ObjectType('Lifecycle2')
```
valentinyanakiev, line 5: Why Lifecycle2?

valentinyanakiev (approved):

PR #985, opened 2021-05-19
Added Ory Kratos + Oathkeeper

valentinyanakiev:


techsmyth (approved):
Looks great. Nicely done.

One comment is that it looks like the ip addresses in the ory config files are hard coded - am guessing we need to be able to modify them?

PR #1005, opened 2021-05-26
Authorization engine + context based authorization

techsmyth:
Allowing context based authorization.

Each entity has a set of authorization rules that grant privileges to certain credentials
Each agent has credentials.

A particular action looks at the combination of rules, credentials held by the agent and the privilege required to determine if the action is allowed or not. 

Key design choices:
- all authorization related checks happen in the graphql layer
- all authorization related logic is in new authorization services per entity

Separate cascading of authorization rules from entity creation. 

New Service: AuthenticationEngineService

New decorator that works on fields and allows to do the privilege checking based on credentials (nice!)

Mutations that need to check entity based authorization rules updated to use Authorization engine. 

To do: check that authorizations are properly cascaded on all created entities.


src/common/enums/authorization.credential.ts
```diff
@@ -6,13 +6,15 @@ export enum AuthorizationCredential {
...
   OrganisationMember = 'organisation-member', // Able to be a part of an organisation
   UserGroupMember = 'user-group-member', // Able to be a part of an user group
-  //UserUpdate = 'user-update', // able to update a user
-  // ChallengeAdmin = 'challenge-admin', // able to manage all aspects of a particular Ecoverse
-  // ChallengeCommunityAdmin = 'challenge-community-admin', // Able to manage membership of a particular community
-  // ChallengeContextAdmin = 'challenge-context-admin', // Able to update the context information for a Challenge
+  ChallengeAdmin = 'challenge-admin', // able to manage all aspects of a particular Ecoverse
+  ChallengeMember = 'challenge-member', // able to manage all aspects of a particular Ecoverse
+  CommunityMember = 'community-member', // Able to be a part of a community
+  UserSelfManagement = 'user-self', // able to update a user
```
valentinyanakiev, line 17: Is it SelfManagement or SelfRegistration?

src/core/authorization/rules/authorization.rule.self.registration.ts
```diff
@@ -0,0 +1,33 @@
...
+import { UserInfo } from '@core/authentication/user-info';
+import { IAuthorizationRule } from '@core/authorization/rules';
+
+export class AuthorizationRuleSelfRegistration implements IAuthorizationRule {
+  userEmail?: string;
+  operation!: string;
+  priority: number;
+
+  constructor(fieldName: string, args: any, priority?: number) {
+    this.operation = fieldName;
```
valentinyanakiev: The operation doesn't seem to be used and seems to be redundant.

src/domain/community/user/user.service.authorization.ts
```diff
@@ -0,0 +1,98 @@
...
+    private userService: UserService,
+    @InjectRepository(User)
+    private userRepository: Repository<User>
+  ) {}
+
+  async applyAuthorizationRules(user: IUser): Promise<IUser> {
+    user.authorizationRules = this.createAuthorizationRules(user.id);
+
+    // cascade
+    const profile = this.userService.getProfile(user);
```
valentinyanakiev, line 39: for discussion: cascading looks like a lot of copy / pasted code - here base functionality / interface would make sense.

valentinyanakiev (commented):
Overall feels almost there. It might be worth it to wrap it up and merge it - doesn't feel far away from what it should be in production. A few points for discussion:

- Cascading rules can be refactored a bit - a lot is repeated, same permissions etc.
- A decorator will really be quite helpful and will simplify the code greatly.
- For later (just a note) - it seems a common approach to have the credentials stringified, we'll need to find the right storage for it.

src/common/enums/authorization.credential.ts
techsmyth: Fixed in an earlier commit

src/core/authorization/rules/authorization.rule.self.registration.ts
techsmyth: good catch, removed

src/domain/community/user/user.service.authorization.ts
techsmyth: as discussed, there is now an inherit method for authorizations

valentinyanakiev (commented):
Overall looks better than the last iteration and a good start. One thing that just hit me is that we might want to have the privileges / credentials configured in YAML/TOML/JSON and not in the database - but that would mean implementing another mini-framework and is more a mid - / long - term comment. DO NOT MERGE yet as this combined with Ory will be impossible to test for Evgeni. We need to brainstorm a test strategy to efficiently get this through the pipeline.

Comoque1:
Invalid error message for community-admin inending to create project:
` "message": "Authorization: unable to grant update access: create project: dxxxx"`
Should return project nameID, instead opportunity nameID


image

techsmyth:
> Invalid error message for community-admin inending to create project:
> ` "message": "Authorization: unable to grant update access: create project: dxxxx"`
> Should return project nameID, instead opportunity nameID
> 
> image

Just pushed up a fix for this particular error message: 505bfb9d3c198b16f87c47e75aa86bb0c115739d

But this is a wider one, all the error messages that are returned upon authorization failure likely need to be reviewed.

Comoque1:
> > Invalid error message for community-admin inending to create project:
> > ` "message": "Authorization: unable to grant update access: create project: dxxxx"`
> > Should return project nameID, instead opportunity nameID
> > image
> 
> Just pushed up a fix for this particular error message: [505bfb9](https://github.com/cherrytwist/server/commit/505bfb9d3c198b16f87c47e75aa86bb0c115739d)
> 
> But this is a wider one, all the error messages that are returned upon authorization failure likely need to be reviewed.

Agree.
Regarding this one, now is better :)
image

Comoque1:
Global admin user doesn't have privileges to update/delete aspect

image
image

techsmyth:
> Global admin user doesn't have privileges to update/delete aspect
> 
> image
> image

I am presuming this was an aspect created on a Project? If so then the issue should now be fixed - just pushed up a commit. Note: it will be fixed for new created aspects, not existing ones.

Comoque1:
So, just reproduced one of the errors:

1. Run tests for the following suite `aspect-on-opportunity.it-spec.ts`
2. Error is thrown for this function
```
let aspectDataPerOpportunity = async (): Promise<String> => {
  const responseQuery = await getAspectPerOpportunity(opportunityId);
  console.log(responseQuery.body);
  let response = responseQuery.body.data.ecoverse.opportunity.context.aspects[0];
  return response;
};
```
image
image

Comoque1:
Non-ecoverse member can create application for global admin.

techsmyth:
> Non-ecoverse member can create application for global admin.

I had a PR up for that, which I just closed and merged into this branch:
https://github.com/cherrytwist/server/pull/1005/commits/ce85ef9dd553bb46fbea6a7b3adc881444a957cf

valentinyanakiev (approved):

PR #1051, opened 2021-06-09
Fixed authorization on references on context

techsmyth:
Also updated business logic to throw an error if a reference with an existing name is used


valentinyanakiev (approved):
I approve this one but it is branched of a branch that is not merged yet...

Comoque1:
Tested OK with the branch.
Now:
 - reference created as part of Challenge creation has `authorisationId`
 - meaningful error is thrown, when intend to create references with same name under the same context
image
image

Creating reference with existing name on different context is successfully created
image
image

PR #1057, opened 2021-06-10
Authorization using verified credentials; SSI integration based on Jolocom SDK

techsmyth:
Integration of Jolocom SDK (sqlite for now, shortly mysql)
Creation of DID identity (jun) for all agents. 
Adds Verified Credentials (VCs) support to Agent
Extended AuthorizationDefinition to be able to specify VCs
Extends AuthorizationEngine to be able to verify based on Verified Credentials.
New mutation for assigning a VC from a Challenge to User
Extended XState integration to be able to work with VC based authorization
Extended sample XState workflow definition for Challenge (default) to use a guard based on VC authorization
...

Notes:
- this is an experimental feature, that is disabled by default
- two updates needed from Jolocom before can fully leverage: support mysql via typeorm (imminent), migrations support from sdk. Latter is to be raised.


src/app.module.ts
```diff
@@ -54,6 +55,34 @@ import { MembershipModule } from '@src/services/domain/membership/membership.mod
...
+      inject: [ConfigService],
+      useFactory: async (configService: ConfigService) => ({
+        type: 'sqlite', //todo: switch to mysql when issue is addressed.
+        insecureAuth: true,
+        synchronize: true /* note: only for demo */,
+        cache: true,
+        entities: [
+          'node_modules/@jolocom/sdk-storage-typeorm/js/src/entities/*.js',
+        ],
+        // host: configService.get(ConfigurationTypes.Identity)?.ssi.jolocom.database
```
valentinyanakiev, line 71: why do you keep all those commented lines?

src/domain/agent/agent/agent.resolver.fields.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+  @Profiling.api
+  async verifiedCredentials(
+    @Parent() agent: Agent
+  ): Promise<VerifiedCredential[]> {
+    const ssiEnabled = this.configService.get(ConfigurationTypes.Identity).ssi
+      .enabled;
+    if (ssiEnabled) {
+      return await this.agentService.getVerifiedCredentials(agent);
+    }
```
valentinyanakiev, line 38: It's better to throw an exception than returning empty array.

src/migrations/1623310575144-agent-auth.ts
```diff
@@ -0,0 +1,35 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class agentAuth1623310575144 implements MigrationInterface {
```
valentinyanakiev: Do we need two migrations for one story?

valentinyanakiev (changes requested):
A few minor changes are required after a quick scan - comments inline.

src/app.module.ts
techsmyth: am waiting for a fix from jolocom; once that is in then want to enable mysql storage.

src/domain/agent/agent/agent.resolver.fields.ts
techsmyth: 0373ff34429161e383c18c9ee5111b08a1a8f27d

src/migrations/1623310575144-agent-auth.ts
techsmyth: 0373ff34429161e383c18c9ee5111b08a1a8f27d

valentinyanakiev (approved):

PR #1065, opened 2021-06-11
fixed bug in api strategy on non-authenticated user; tidied up logging

techsmyth:


src/core/authentication/authentication.service.ts
```diff
@@ -36,12 +36,14 @@ export class AuthenticationService {
...
         LogContext.AUTH
       );
-      // Allow the user to create a credential for themselves for the context of this request
-      const createUserCredential = new Credential(
-        AuthorizationCredential.UserSelfManagement,
-        agentInfo.email
-      );
-      agentInfo.credentials.push(createUserCredential);
+      // Allow the user to create a credential for themselves for the context of this request if a valid email is being used
+      if (email.length > 0) {
```
valentinyanakiev, line 45: I think we need a check whether the email is trutsy at the beginning - the logic should be decoupled and you are making assumptions here that something is defined whereas we can have n amount of different strategies, you can't rely that something is coming from a specific place in a specific format.

src/core/authorization/graphql.guard.ts
```diff
@@ -19,14 +19,15 @@ export class GraphqlGuard extends AuthGuard([
   'oathkeeper-jwt',
   'oathkeeper-api-token',
 ]) {
-  agentInfo?: AgentInfo;
+  instanceId: string;
```
valentinyanakiev, line 22: Wasn't this just for debugging / tracing the issue?

valentinyanakiev (changes requested):
The instance id might be useful but I believe checking whether the mail is defined in the Authentication Service should be a given.

valentinyanakiev (approved):

PR #1090, opened 2021-06-16
User group assigned fixes; cascaded fixes

techsmyth:
UserGroup assignment, for both groups on Community and Organisations, are now working, 

Creating of an ecoverse no longer automatically creates a host org in the ecoverse service.


Comoque1:
The tests are passing, except the scenario: `create group with empty name`

image
image

techsmyth:
> The tests are passing, except the scenario: `create group with empty name`
> 
> image
> image

Good catch. Added a minimum lenghth of 2 for user group names. 

Re "restricted group name tests", I presume these are no longer needed?

atstoyanov (approved):

PR #1117, opened 2021-06-22
cascading of removals, especially re authorization

techsmyth:
Moved all base entity types into one folder
Moved all functionality related to manipulating authorizationDefinitions to a service in that entity's folder
Moved the relationship between organisation / challenge and organisation / ecoverse to be based on credentials - so no longer direct references. 

Database schema is reset for hopefully the last time...

Note: still some entities being left over but should be a lot less...


Comoque1:
With these changes, there is significant improvements in terms of less data leftovers.
There is an issue introduced:
querying user from through application, throws an error: `"message": "Cannot return null for non-nullable field Application.user."`
image
The user exists.

techsmyth:
> With these changes, there is significant improvements in terms of less data leftovers.
> There is an issue introduced:
> querying user from through application, throws an error: `"message": "Cannot return null for non-nullable field Application.user."`
> image
> The user exists.

Hi Evgeni, afraid needed to go deep on the code again with circular dependencies. Issue with user application field should now be fixed.

Comoque1:
Unable to delete Organisation after the latest  changes
image
image

Comoque1:
Tests passing
image

Comoque1 (approved):

PR #1176, opened 2021-07-02
made user phone number and email to be dependent on having read access to the user profile

techsmyth:
For gdpr reasons the email + phone number fields on a user should not be visible to global registered users.

Added ability to reset the authorization definition on an ecoverse and have it cascade through all contained entities; needed to deal with any bugs that show up in ecoverse authorization definitions.

Removed dependency that should not have been there between authorization server to challenge entity.


Comoque1:
Challenge-admin - improved one case:
image

image

image

image

image

image

image

image

image

image

image

Comoque1 (approved):

PR #1259, opened 2021-07-28
JWKS exposed publicly (only for dev purposes!!!)

valentinyanakiev:
Both compose services + compose services & server should work with the client out-of-the-box now


Comoque1 (approved):
Works fine now.

Issue #1088, opened 2021-06-15
BUG: typeorm failing at version 0.2.33 (+ 0.2.34)


techsmyth:
**Describe the bug**
Increasing the typeorm version to 0.2.34 results in the typeorm generation failing:

![image.png](https://images.zenhubusercontent.com/5ecb9817aeb589610134bd43/4d57d8e4-86cb-467b-8d93-c1ebbdfc31d3)

For now reverted to 0.2.32 as this version is working. Any higher version is failing.

techsmyth:
No longer an issue: have now updated to v0.2.35

PR #1261, opened 2021-07-29
Updated nest to v8 and (almost) all packages

valentinyanakiev:
Apollo express and Ory Client are pending. Ory requires update on the server and Apollo requires careful planning as subscriptions, upload, graphql playground are handled differently in v3 than v2.


package.json
```diff
@@ -36,91 +36,73 @@
     "migration:show": "npm run typeorm migration:show"
   },
   "dependencies": {
-    "@jolocom/sdk": "^1.2.0",
```
techsmyth, line 39: Please leave the jolocom dependencies in

package.json
```diff
@@ -36,91 +36,73 @@
...
     "passport-jwt": "^4.0.0",
     "reflect-metadata": "^0.1.13",
     "replace-special-characters": "^1.2.5",
     "rimraf": "^3.0.2",
-    "rxjs": "^6.6.7",
+    "rxjs": "^7.3.0",
     "sqlite3": "^5.0.2",
-    "typeorm": "0.2.32",
-    "uuid": "^8.3.2",
+    "typeorm": "^0.2.35",
```
techsmyth, line 74: Interesting if this is now working, would imply that the bug was triggered by one of the dependencies that were removed...

techsmyth (approved):
Looks like a good tidy up. Have not tried running it yet.

Key to capture the other notes re Apollo version / Ory update elsewhere.

package.json
valentinyanakiev: I don't think we should until they fix their issues and update their packages. They haven't updated multiple packages in ages (e.g. typeorm dependency 0.2.24 is causing issues with our upgrade, and current typeorm version is 0.2.35, which is conflicting to theirs). Running dependency analyzer shows we are not actively using it either so it's just a nuisance - I suggest we re-add it once it is ready to be used.

package.json
valentinyanakiev: Which bug and how can I check? I recall there was a bug but I haven't tested this update against it...

package.json
techsmyth: ok, lets not hold up this PR then - can you please raise another issue with the problems you ran into when using their components with latest version of typeorm?

package.json
techsmyth: https://github.com/alkem-io/server/issues/1088

valentinyanakiev: Haven't seen that error running the application, merging the PR

Comoque1:
Tests are passing against this branch, so suggest to merge it.

PR #1266, opened 2021-08-03
Updated Dockerfile to install latest npm version

valentinyanakiev:


Dockerfile
```diff
@@ -16,6 +16,7 @@ ARG ENV_ARG=production
 # where available (npm@5+)
 COPY package*.json ./
 
+RUN npm i -g npm@latest
```
techsmyth, line 19: should we not be hard coding a particular version for the generation of the dockerfile?

techsmyth (approved):
See comment re specific version...

Dockerfile
valentinyanakiev: I'd prefer npm version to be the latest, to be honest, it's the same in travis.yml and I also don't install it locally with specific version but with the latest, I suspect (mostly) everybody does the same.

PR #1284, opened 2021-08-06
Canvas support; member on EcosystemModel

techsmyth:
Add a new entity: Canvas 
- for the storage of visualisation canvases

Extended EcosystemModel entity to have a member that is a Canvas
- just one for now, which actually is what I think makes sense in terms of Canvas for an Ecosystem Model

Added mutation to allow directly updating the EcosystemModel
- this entity will likely have a lot of direct updates

Added sample mutation for updating an EcosystemModel

Add migration to update the data store


pimotte:
Just to check, I thought the intention was to just store the raw json as a string, not a url? I'm good with doing something like url to ipfs, but I though we had decided just stuffing it in the database at first.

Besides that, dry-reviewed, LGTM.

techsmyth:
Yes this is intended to store the raw JSON; might then be better to mark the 'value' field as being JSON instead of string. 

But should either way be good for you to try out the end-end flow I think...

graphql-samples/mutations/update/update-ecosystem-model
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+Variables:
+{
+  "ecosystemModelData": {
+    "ID": "uuid",
+    "description": "some description",
+    "canvas": {
+      "value": "http://someUri"
```
pimotte, line 14: This was why I was confused about the intention

pimotte:
- [ ] TODO: Ensure that existing opportunities can get a canvas through update (or the canvas is added in migration?)

valentinyanakiev (approved):

PR #1294, opened 2021-08-12
separate out DTOs used for creating challenges in ecoverse + challenge

techsmyth:
Separate DTOs for the creation of a challenge in an ecvoerse and in a challenge. 

Update DTO field names to make it clearer.

Updated graphql samples.

### DIFF
--- a/src/domain/challenge/challenge/challenge.resolver.mutations.ts
+++ b/src/domain/challenge/challenge/challenge.resolver.mutations.ts
@@ -26,7 +26,7 @@ import { IChallenge } from './challenge.interface';
 import { IUser } from '@domain/community/user/user.interface';
 import { AssignChallengeAdminInput } from './dto/challenge.dto.assign.admin';
 import { RemoveChallengeAdminInput } from './dto/challenge.dto.remove.admin';
-import { CreateChallengeInChallengeInput } from './dto/challenge.dto.create.in.challenge';
+import { CreateChallengeOnChallengeInput } from './dto/challenge.dto.create.in.challenge';
 
 @Resolver()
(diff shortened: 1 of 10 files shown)


src/domain/challenge/challenge/challenge.resolver.mutations.ts
```diff
@@ -45,10 +45,10 @@ export class ChallengeResolverMutations {
   @Profiling.api
   async createChildChallenge(
     @CurrentUser() agentInfo: AgentInfo,
-    @Args('challengeData') challengeData: CreateChallengeInput
+    @Args('challengeData') challengeData: CreateChallengeInChallengeInput
```
valentinyanakiev: Won't it be better to name it CreateSubChallengeInput or CreateChildChallengeInput?

src/domain/challenge/ecoverse/ecoverse.resolver.mutations.ts
```diff
@@ -119,18 +119,20 @@ export class EcoverseResolverMutations {
   @Profiling.api
   async createChallenge(
     @CurrentUser() agentInfo: AgentInfo,
-    @Args('challengeData') challengeData: CreateChallengeInput
+    @Args('challengeData') challengeData: CreateChallengeInEcoverseInput
```
valentinyanakiev: All the rest of the mutations are xxxOnxxx, and here is xxxInxxx, it is not consistent with the rest of the API

valentinyanakiev (changes requested):
A couple of small naming changes required in my opinion.

valentinyanakiev (approved):

PR #1298, opened 2021-08-13
Search enhancements; admin visibility on users

techsmyth:
Updates
- Allow ecoverse admins / challenge admins to have full read access to users in the relevant community
- Search: expanded set of fields on context that are searched
- Search: also now search on Opportunities

Bug fixes
- Only accept UUIDs for creation of challenges / opportunities within Challenges

### DIFF
--- a/src/domain/challenge/ecoverse/ecoverse.service.ts
+++ b/src/domain/challenge/ecoverse/ecoverse.service.ts
@@ -39,7 +39,7 @@ import { IUser } from '@domain/community/user/user.interface';
 import { RemoveEcoverseAdminInput } from './dto/ecoverse.dto.remove.admin';
 import { UserService } from '@domain/community/user/user.service';
 import { UpdateEcoverseInput } from './dto/ecoverse.dto.update';
-import { CreateChallengeInEcoverseInput } from '../challenge/dto/challenge.dto.create.in.ecoverse';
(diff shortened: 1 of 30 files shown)


src/domain/challenge/ecoverse/ecoverse.service.ts
```diff
@@ -295,10 +295,10 @@ export class EcoverseService {
     );
   }
 
-  async createChallenge(
-    challengeData: CreateChallengeInput
+  async createChallengeInEcoverse(
```
valentinyanakiev, line 298: createChallengeOnEcoverse, to be consistent with the rest

src/domain/community/user/user.service.authorization.ts
```diff
@@ -35,9 +35,9 @@ export class UserAuthorizationService {
       user.authorization
     );
 
-    user.authorization = this.appendCredentialRules(
+    user.authorization = await this.appendCredentialRules(
```
valentinyanakiev, line 38: This makes the method inconsistent with other implementations, where it is synchronous, e.g. organisation.service.authorization.ts. Either make it synchronous everywhere or asynchronous everywhere. Otherwise this inconsistency will bring problems.

valentinyanakiev (changes requested):
Please review comments inline.

valentinyanakiev (approved):

PR #1300, opened 2021-08-13
IPFS Fix

valentinyanakiev:
When working on the 'web port' (3000), having an image uploaded to ipfs should resolve it via traefik route through http://localhost:3000/ipfs/[CID].

However, routing to that endpoint has a redirect to XXX.ipfs.localhost:3000 that is not managed by the same IPFS route and ends up trying to be routed to the web client. The redirect to .ipfs.domain is not observed on any prod environment.

I have fixed it by changing the IPFS route to match port 8080 and not 3000. This will imply changes on demo and client as well. Alternative approach would be to try match and redirect .ipfs.localhost as well to the correct endpoint and port.

### DIFF
--- a/.build/traefik/http.yml
+++ b/.build/traefik/http.yml
@@ -122,10 +122,10 @@ http:
         - 'web'
 
     ipfs:
-      rule: 'PathPrefix(`/ipfs`)'
+      rule: 'PathPrefix(`/ipfs`) || HostRegexp(`{^[a-zA-Z0-9_.-]*$}.ipfs.localhost`)'
       service: 'ipfs-http-adapter'
       entryPoints:
-        - 'ipfs-http-adapter'
+        - 'web'
 
     ipfs-api:
       rule: 'PathPrefix(`/`)'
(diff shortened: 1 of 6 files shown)


valentinyanakiev:
> When working on the 'web port' (3000), having an image uploaded to ipfs should resolve it via traefik route through [http://localhost:3000/ipfs/[CID]](http://localhost:3000/ipfs/%5BCID%5D).
> 
> However, routing to that endpoint has a redirect to XXX.ipfs.localhost:3000 that is not managed by the same IPFS route and ends up trying to be routed to the web client. The redirect to .ipfs.domain is not observed on any prod environment.
> 
> I have fixed it by changing the IPFS route to match port 8080 and not 3000. This will imply changes on demo and client as well. Alternative approach would be to try match and redirect .ipfs.localhost as well to the correct endpoint and port.

Update: Implemented the alternative, much cleaner fix with less impact on other repositories.

techsmyth (approved):
Working nicely!

Do we need to apply the same set of fixes also through to client + demo repos?

PR #1309, opened 2021-08-17
BUG: Subscriptions are not disposed / disconnected correctly

Enderedne:
### Drop problematic subscription mechanism
**Reasoning**: A subscription created with gql in apollo opens a socket, BUT when the subscription is disposed the socket is not terminated. This leads to open sessions even though no one is listening. On re-subscribing the existing session along with the new one will push updates to the same PubSub instance creating duplicate messages. 

- subscribe to all matrix events from the start
- use the PubSub implementation as the sole subscription control method

### Additional context

Automatically mark messages as read, because the functionality is not exposed to the client. Once notifications are in place this needs to be revisited. Marked with a TODO.

Add any other context

### Governance 

- [ ] Documentation is added


valentinyanakiev (approved):
:+1:

PR #1311, opened 2021-08-17
Added ordering of ecoverses

valentinyanakiev:
- Public ecoverses first
- Then combined number of challenges + opportunities

I am not reading sub-challenges so far.


techsmyth (approved):
Cannot test but code looks good. Can optimize later if needed but doubt this sorting will have a huge overhead near term.

valentinyanakiev:
> Cannot test but code looks good. Can optimize later if needed but doubt this sorting will have a huge overhead near term.

It's actually pretty easy to test via client or:

query
{
  ecoverses
  {
    nameID
  }
}

If you have ecoverses created private / public / different amount of challenges + opportunities, the returned namesIDs will be ordered

PR #1316, opened 2021-08-19
Release 20.08.2021

valentinyanakiev:
- Fixed bug where communication sockets were not disposed
- Fixed bug where organisations searched by tagset threw an exception
- Added ordering of the Ecoverses returned by the `ecoverse` query
- Exposed mailslurper through traefik
- Exposed parent Challenge on Opportunity


Comoque1 (approved):

.build/traefik/http.yml
```diff
@@ -1,9 +1,14 @@
 http:
   services:
-    ipfs-http-adapter:
```
techsmyth, line 3: Is the ipfs adapter still exposed?

techsmyth (approved):
Nice set of updates. Only thing that looked odd is the ipfs http adapter?

.build/traefik/http.yml
valentinyanakiev: It is and it must be exposed in order to make IPFS work with Alkemio Server outside of the docker-compose network.

techsmyth: This PR makes it look like the ipfs adapter is being removed

PR #1322, opened 2021-08-24
added org owner credential with associated mutations

techsmyth:


src/domain/community/organisation/organisation.service.ts
```diff
@@ -378,4 +414,51 @@ export class OrganisationService {
...
+    const organisationID = removeData.organisationID;
+    const organisation = await this.getOrganisationOrFail(organisationID);
+    const agent = await this.userService.getAgent(removeData.userID);
+
+    if (checkAtLeastOneOwner) {
+      const orgOwners = await this.userService.usersWithCredentials({
+        type: AuthorizationCredential.OrganisationOwner,
+        resourceID: organisationID,
+      });
+      if (orgOwners.length < 2)
```
valentinyanakiev: This is a bit of misleading feedback. If the org was not initialized correctly or e.g. somebody runs a script and alters the data, and there are no Organisation owners, you will get feedback that you can't remove the last one, even if there aren't any. I think if orgOwners.length === 0 you need to throw a different exception giving feedback that the org credentials are in an inconsistent state.

valentinyanakiev (changes requested):

src/domain/community/organisation/organisation.service.ts
techsmyth: Fair point, have updated logic check to check if exactly 1 owner. That way orgs that do not yet have an owner are handled as before.

valentinyanakiev (approved):

PR #1327, opened 2021-08-24
Extend membership entities so they can look normalized for the apollo  cache

atstoyanov:
The apollo client's cache stores entites normalized.
In the case of UserMembership our hierarchy structure can't be stored as normalized structure using only the ID as a key.

Lets say we have next two results for two different users:
```json
{
   "membershipUser":{
      "ecoverses":[
         {
            "id":"e02c0498-4116-4ac4-8b5a-aa992c0fc824",
            "challenges":[
               {
                  "id":"76fd5c6e-bf4c-458b-a1a6-6e7e63999b5b"
               },
               {
                  "id":"17bb5476-6cff-41e5-bdb6-cdaab2fbaf2e"
               }
            ]
         }
      ]
   }
}
```

```json
{
   "membershipUser":{
      "ecoverses":[
         {
            "id":"e02c0498-4116-4ac4-8b5a-aa992c0fc824",
            "challenges":[
               {
                  "id":"76fd5c6e-bf4c-458b-a1a6-6e7e63999b5b"
               }
            ]
         }
      ]
   }
}
```
The Ecoverse in both results have same ID so in the cache single Record will be created.
When the first data is stored in the cache the two records for the Ecoverse.Challenges will be stored in the case with respective IDs. After the second data is received the first Ecoverse record will be overwritten with the new challenges, so one challenge is lost. And because same ecoverse is return as result from 2 different queries they will have same results.

The other problem here is that the membershipUser does not have ID and cache can't merge it easily.

### First Solution - implemented

First we need the membershipUser to have ID. Best candidate is the UserId.
Second we need a way to distinguish the ecoverses result (MembershipUserResultEntryEcoverse) between two users. So complex cache key can be used. UserID+EcoverseID. Unfortunately the cache can't use value from the parent to create a complex key, but can work with child object keys. The easiest way was to add the parentID to the Ecoverse result. Now we have a complex key: Id+ parentID.

### Second solution
the result can be normalized.

membershipUser can be flatten to:
```
{
  ecoverses: MembershipResultEntry;
  challenges: MembershipResultEntry;
  opportunities: MembershipResultEntry;
}

```
and MembershipResultEntry will need a `parentId` or `ecoverseId` so we know which entity to which ecoverse  belongs to.

NB! We may have issue with nested challenges and fallowing the parent-child chain.

### DIFF
--- a/src/services/domain/membership/membership.dto.organisation.result.ts
+++ b/src/services/domain/membership/membership.dto.organisation.result.ts
@@ -1,8 +1,14 @@
+import { UUID } from '@domain/common/scalars';
 import { Field, ObjectType } from '@nestjs/graphql';
 import { MembershipResultEntry } from './membership.dto.result.entry';
 
 @ObjectType()
 export class OrganisationMembership {
+  @Field(() => UUID, {
+    nullable: false,
+  })
+  id!: string;
+
   @Field(() => [MembershipResultEntry], {
     description: 'Details of Ecoverses the Organisation is hosting.',
   })
--- a/src/services/domain/membership/membership.dto.user.result.entry.ecoverse.ts
+++ b/src/services/domain/membership/membership.dto.user.result.entry.ecoverse.ts
@@ -8,11 +8,6 @@ export class MembershipUserResultEntryEcoverse extends MembershipResultEntry {
   })
   ecoverseID: string;
 
-  @Field(() => String, {
-    description: 'The Parent User ID',
-  })
-  userID: string;
-
   @Field(() => [MembershipResultEntry], {
     description: 'Details of the Challenges the user is a member of',
   })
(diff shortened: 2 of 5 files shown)


techsmyth (commented):
Nice analysis, clear now what the problem is, and first option at first glance looks the better one.

Seeing this analysis made me think that we are actually doing this fundamentally wrong in the api: we should not be re-using the user or ecoverse ID for the membership result IDs. Each membership result should actually have its own unique ID (potentially derived from the user / ecoverse ID)  and then have specific fields for the id of the entity that each result is for (e.g. userID / ecoverseID).

Thoughts?

techsmyth (approved):
Working nicely for me now; glad this cache issue is clear.

PR #1340, opened 2021-08-30
fixed loading of tags by reloading of ecoverses after sorting

techsmyth:
Ideally would have used a .map implementation but that was not working out easily for me. 

Clearly this puts an extra load for each ecoverse, but this probably needs to be done in all cases where we directly query the db and not use the ORM layer to load.

### DIFF
--- a/src/domain/challenge/ecoverse/ecoverse.service.ts
+++ b/src/domain/challenge/ecoverse/ecoverse.service.ts
@@ -147,14 +147,37 @@ export class EcoverseService {
   }
 
   async getEcoverses(): Promise<IEcoverse[]> {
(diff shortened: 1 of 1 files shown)


atstoyanov (approved):

valentinyanakiev (changes requested):
Same result with less db roundtrips (only one) can be achieved by adding .leftJoinAndSelect(...)

techsmyth:
I know that I could fix it that way by adding a left join and select. However I wanted to ensure that the loading goes through the typeorm layer - so that do *not* return data that is loaded directly via queries.

valentinyanakiev (approved):

PR #1360, opened 2021-09-09
allow RegisteredUsers to have READ access to Organisation members / groups, and also to communities on non-private ecoverse/challenge/opportunites

techsmyth:
The authorization policy does need to be reset on Organizations and on Ecoverses after deploying this update.


valentinyanakiev (commented):
Code looks OK but how is this handled via the API? Creating a new registered user, not a member to any ecoverse / challenge / opp, can't see the relevant data in the application (e.g. community).

valentinyanakiev (approved):

PR #1362, opened 2021-09-10
fix nameID comparision to ignore case differences

techsmyth:
Came across an issue when testing the org admin page update that the nameID comparison was case sensitive, which it should not be.


valentinyanakiev (commented):
For discussion - this is actively changing inputted data from an ecoverse admin / organization admin - e.g. you input Digi-Campus and you end up with digi-campus. I don't believe that is the correct approach - I think if we want only lowercase nameIDs we should validate the createOrganizationDto and also validate the react textbox whether it's lowercase or not and simply only accept lowercase.

valentinyanakiev (approved):

PR #1381, opened 2021-09-15
add support for OpportunityAdmin credential

techsmyth:
New credential: OpportunityAdmin
New mutations: Assing/Remove OpportunityAdmin


src/domain/collaboration/opportunity/opportunity.resolver.mutations.ts
```diff
@@ -169,4 +172,46 @@ export class OpportunityResolverMutations {
...
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => IUser, {
+    description: 'Assigns a User as an Opportunity Admin.',
+  })
+  @Profiling.api
+  async assignUserAsOpportunityAdmin(
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args('membershipData') membershipData: AssignOpportunityAdminInput
+  ): Promise<IUser> {
+    const opportunity = await this.opportunityService.getOpportunityOrFail(
```
valentinyanakiev, line 185: I know this is done consistently everywhere, but it's all copy-pasted (but one line). My suggestion is to move most of this logic in the xxxService (in this case opportunityService) and have the common logic in a method.

valentinyanakiev (commented):
I feel the amount of boilerplate copy-paste code in the API is getting out of hands and needs to be dealt with - we have too many unnecessary code paths and copy-pasting is error prone. To be discussed.

valentinyanakiev (approved):

PR #1385, opened 2021-09-16
updating of authorization policy applies the auth policy throughout the tree, not a separate path

techsmyth:
Previously there was a separate update path for authorization policy updates. This is removed: all authorization changes go through the same path now. 

Note: requires reseting the auth policy on ecoverse, or just toggling the auth policy via the admin screen on each ecoverse.


valentinyanakiev (approved):
I thoroughly enjoy this kind of change. Keep it up :+1:

PR #1407, opened 2021-09-24
Applications enhancements; flag for final state on Lifecycle; dev updates

valentinyanakiev:
API
- Extended Lifecycle entity to have a flag for whether it is finalized
- Standardized enumeration usage: Renamed AuthorizationCredential enum values to be XXX_YYY

Bug fixes
- creating applications now takes into account if a previous application is finalized
- creating applications now checks if a user is already a member of the community
- removed subscription event around avatar uploaded which was not used
- parentCommunity is now set properly for communities on newly created challenges

Developer setup
- Added Synapse Admin to quickstart-services
- Updated networking diagram


techsmyth (approved):
Nice set of updates

PR #1408, opened 2021-09-24
Fix setting of parentCommunity for new Challenges; migration to repair needed

techsmyth:
When creating challenges on an ecoverse the parent community was not being set. This meant the check for parent community for adding member was not being verified. 

Note: a seprate migration is needed that sets the parent community id field properly for all challenges. Raised as a separate item: https://github.com/alkem-io/server/issues/1409

### DIFF
--- a/src/domain/community/community/community.service.ts
+++ b/src/domain/community/community/community.service.ts
@@ -177,6 +177,28 @@ export class CommunityService {
     });
   }
 
+  async getParentCommunity(
+    community: ICommunity
+  ): Promise<ICommunity | undefined> {
+    // const communityWithParent = await this.getCommunityOrFail(community.id, {
+    //   relations: ['parentCommunity'],
+    // });
+
+    const communityWithParent = await this.communityRepository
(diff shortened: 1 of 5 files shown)


src/domain/community/community/community.service.ts
```diff
@@ -155,6 +155,20 @@ export class CommunityService {
     return true;
   }
 
+  async getParentCommunity(
+    community: ICommunity
+  ): Promise<ICommunity | undefined> {
+    const communityWithParent = await this.getCommunityOrFail(community.id, {
+      relations: ['parentCommunity'],
```
valentinyanakiev, line 162: I think it's recursive structure and has a relation to community, parentCommunityId is the id of related community, but it's not the relation name, if I understand correctly

valentinyanakiev (changes requested):
In my opinion, the service method commented won't work. It's hard to be tested as it's orphaned code - it's not invoked anywhere.

valentinyanakiev (approved):

PR #1424, opened 2021-09-29
Authorization Framework streamline

techsmyth:
Moved all Authorization administration logic to a new services / admin / authorization module (mutations, queries etc)
Removed the dependency from Authorization module on domain model
Extended AuthorizationPolicy module to have a field to return the set of privileges for the current Agent


valentinyanakiev (changes requested):
I don't believe the design works. If I run a query with a non-privileged user (ecoverse-member):
query{
  ecoverses
  {
    displayName
    authorization
    {
      myPrivileges
    }
    challenges
    {
      authorization
      {
        myPrivileges
      }
      opportunities
    	{
        displayName
        authorization
        {
          myPrivileges
        }
      }
    }
  }
}

I get an exception that I can't read the challenges due to insufficient privileges (READ). I am not sure it makes sense to be checking my privileges for e.g. challenges and get an exception that I don't have privileges to read the privileges when I run the query to understand whether I have privileges in the first place. In my opinion, this requires discussion.

valentinyanakiev (approved):

PR #1429, opened 2021-10-02
added sorting order to application questions

techsmyth:
Extends the NVP entity to store a sort order. Migration is in place.

Updated templates for applications to specify the sort order. 

Extended question creation DTO to specify the sort order.

### DIFF
--- a/src/migrations/1633155784833-application-sort.ts
+++ b/src/migrations/1633155784833-application-sort.ts
@@ -1,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class applicationSort1633155784833 implements MigrationInterface {
+  name = 'applicationSort1633155784833';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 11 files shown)


src/migrations/1633155784833-application-sort.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class applicationSort1633155784833 implements MigrationInterface {
```
hero101, line 3: The migration must also take care of the existing data.

hero101 (changes requested):

src/migrations/1633155784833-application-sort.ts
techsmyth: The migration does take care of it:
image
The field gets a default value of zero, which is fine. It means existing applications are still not sorted, but that is something we just live with. Any new applications wil be properly sorted.

hero101 (approved):

PR #1431, opened 2021-10-04
added org owner update credential

techsmyth:


Comoque1:
Just tried it on test env and still doesn't seem organizationOwner to have rights to update organization.
image
image

I don't even see in authorization rules definition for organizationOwner
image

techsmyth:
The authorization policy would need to be reset on the organizations for the updated rules to take effect. Did you do that?

Comoque1:
I missed that one.
Now looks OK!
image

Comoque1 (approved):

PR #1435, opened 2021-10-05
Kratos 0.7.6 + Communications

valentinyanakiev:
Authentication / Authorization:
- Authentication framework updated and AJAX is now used instead of redirects (better user experience). Kratos version is now 0.7.6. The change is backwards incompatible.
- Subscriptions extended to leverage existing authorization / authentication framework
- Allow users to see what privileges they have on every authorizable entity
- Refactored authorization logic to split out Administration from the core runtime logic.

Functional Improvements
- Communications (updates) robustness: caching of user data, usage of matrix state, friendly names for community rooms, .enums properly used, mutation for community room message removal, ..
- API now supports subscriptions for key events, including Communication Updates. 
- Feature flag added to enable turning on / off subscriptions on clients.
- membership data for users expanded to give additional information for Applications so that that information can be better displayed on clients
- Applications now have a sort order, so that the order of the questions can be controlled in the template. Previously the order on the client was displaying randomly.
- Community() query on Ecoverse extended to allow retrieving of any community within that Hub

Bug fixes
- Fixed issue whereby some user updates were failing due to case sensitive nameID matching
- Org owners can now also update Organizations


techsmyth (approved):
Great to have kratos + the comms in...

techsmyth (approved):
Nice update

PR #1450, opened 2021-10-09
matrix stabilisation

techsmyth:
usernames to adhere to matrix policy
only filter out explicit expected error when checking if a user exists
explicit check to defend against invalid room id

Updated rate limits in synapse configuration to avoid "too many requests" errors

### DIFF
--- a/src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
+++ b/src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
@@ -60,6 +60,7 @@ export class MatrixUserAdapterService {
   }
 
   convertMatrixUsernameToEmail(username: string) {
+    // Todo: this is problematic...
     return username.replace(/[=]/g, '@');
   }
 
--- a/src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
+++ b/src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
@@ -6,7 +6,7 @@ import { IMatrixUser } from './matrix.user.interface';
 
 @Injectable()
 export class MatrixUserAdapterService {
-  private mailRegex = /[@]/g;
+  private emailUsernameSpecialCharactersRegex = /[^a-zA-Z0-9_/-/=.]/g;
 
   constructor(
     @Inject(WINSTON_MODULE_NEST_PROVIDER)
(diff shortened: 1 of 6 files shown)


src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
```diff
@@ -26,7 +26,9 @@ export class MatrixUserAdapterService {
   }
 
   convertEmailToMatrixUsername(email: string) {
-    return email.replace(this.mailRegex, '=');
+    // TODO: "User ID can only contain characters a-z, 0-9, or '=_-./'"
+    const username = email.replace(this.mailRegex, '=');
```
valentinyanakiev: I suggest we do the validation before we push this change. It can be done via regex. If we don't do it now we risk unstable/exceptional behavior (will 100% happen).

src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
```diff
@@ -26,7 +26,9 @@ export class MatrixUserAdapterService {
...
+    return username.toLowerCase();
```
valentinyanakiev, line 45: If we expect and want to have emails / UUIDs or any values coming in as lowercase, we should do a data validation client- and server- side, similar to https://app.zenhub.com/workspaces/alkemio-5ecb98b262ebd9f4aec4194c/issues/alkem-io/server/1371. This approach will make the line redundant.

valentinyanakiev (commented):
Requires discussion - mostly looks OK but I'd like to see the email validated before it's converted to matrix user account. The other comment, imho, can be addressed separately.

valentinyanakiev:
Also, it will be nice to update the package version.

src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
techsmyth: Addressed in: cf75df7b53c8d2cb44dd4194d6fd115dd834bf5c

src/services/platform/matrix/adapter-user/matrix.user.adapter.service.ts
techsmyth: Emails do not have to be coming in as lower case. We do have data validation happening on emails. Upper case is allowed and should be supprted. 

The issue here is that the allowed character set for email usernames is broader than the allowed character set for matrix user IDs.

valentinyanakiev: Does lowercase cover all unsupported scenarios for matrix thou (supported for email, unsupported for matrix)?

techsmyth: See the updated code - symbols in emails that are not supported are removed. It does mean the backwards translation is broken. Separate card for that.

valentinyanakiev (approved):

PR #1457, opened 2021-10-11
Communications: domain mgmt of matrix ID

techsmyth:
One way mapping from email to communication ID i.e. Matrix. Bi-directional cannot be guaranteed so we cannot build on a 1-1 mapping. 

Code simplified to rely on domain model mgmt of communication ID.

Migration added to add the communicationID field to the user entity


src/services/platform/communication/communication.dto.message.received.ts
```diff
@@ -21,6 +21,7 @@ export class CommunicationMessageReceived {
   })
   message!: CommunicationMessageResult;
 
+  // TODO: is this needed? Should it not just be the user ID?
```
Enderedne: Related to the change in `src/services/platform/matrix/events/matrix.event.adapter.room.ts`.

This is going to break the subscriptions. Will address this in my commit.

src/services/platform/communication/communication.service.ts
```diff
@@ -384,13 +390,9 @@ export class CommunicationService {
...
     if (mappedDirectRoomId) {
-      const emailReceiver =
-        this.matrixUserAdapterService.convertMatrixIdToEmail(
-          mappedDirectRoomId
-        );
-      await this.convertMatrixRoomToDirectRoom(
+      return await this.convertMatrixRoomToDirectRoom(
         matrixRoom,
-        emailReceiver,
+        mappedDirectRoomId, // may need to convert from an matrix ID to matrix username
```
Enderedne: Should be as it is. Will change the name to make it clearer

Enderedne (dismissed):

src/domain/community/user/user.service.ts
```diff
@@ -92,6 +97,25 @@ export class UserService {
     }
 
     const response = await this.userRepository.save(user);
+
+    // all users need to be registered for communications at the absolute beginning
+    // there are cases where a user could be messaged before they actually log-in
+    // which will result in failure in communication (either missing user or unsent messages)
+    // register the user asynchronously - we don't want to block the creation operation
+    this.communicationService.tryRegisterNewUser(user.email).then(
```
valentinyanakiev, line 106: I believe this fails the bootstrap process. I don't believe the promise rejection is handled properly. See the screenshot:
image

valentinyanakiev: Easy to reproduce - drop alkemio db, drop homeserver db, restart synapse server, run alkemio migrations, start. Reproducible 100%

src/domain/community/user/user.resolver.subscriptions.ts
```diff
@@ -50,7 +55,7 @@ export class UserResolverSubscriptions {
     ) {
       // Note: by going through the passport authentication mechanism the "user" property on
       // the request will contain the AgentInfo that was authenticated.
-      return payload.userEmail === context.req?.user?.email;
+      return payload.userID === context.req?.user?.communicationID;
```
valentinyanakiev: That's a bit misleading, the userID in the context of matrix is matrix ID and the type defines it as the email of the user, but that is not the email of the user profile, but the matrix ID, as far as I see. Also the user context has both the user email and theh communications ID and you pick the communicationID and compare it with a property called userID - that's extremely confusing. I suggest CommunicationMessageReceived type to be changed to replace userID with matrixID, if I understand this logic correctly.

src/core/authentication/agent-info.ts
```diff
@@ -8,4 +8,5 @@ export class AgentInfo {
   lastName = '';
   credentials: ICredential[] = [];
   verifiedCredentials: VerifiedCredential[] = [];
+  communicationID = '';
```
valentinyanakiev, line 11: In my opinion, this should not be here in the final design. The AgentInfo should have properties only from the user account payload (kratos traits in our case) and credentials.

src/domain/community/user/user.service.ts
```diff
@@ -410,35 +485,40 @@ export class UserService {
...
-      const user = await this.getUserByEmail(sender);
-      if (!user) {
-        continue;
-      }
+    return communicationID;
+  }
 
-      senderMap[sender] = user.id;
+  // Not sure about the placement of this one
+  async populateRoomMessageSenders(rooms: (CommunityRoom | DirectRoom)[]) {
```
valentinyanakiev: I believe in this case (CommunityRoom | DirectRoom) can be CommunityRoom only, as DirectRoom extends CommunityRoom.

valentinyanakiev (changes requested):
Comments inline, main areas:
- naming / types can be improved in a couple of places
- async registration of users fails - needs to be fixed
- we need to pick UsarAgent communicationsID and authorization service - can be done separately.

src/domain/community/user/user.service.ts
Enderedne: Traced that down to incorrect HMAC - this means that the registration_shared_secret might not be the same, or a different sort of configuration issue.

src/domain/community/user/user.resolver.subscriptions.ts
techsmyth: Addressed

src/domain/community/user/user.interface.ts
```diff
@@ -36,6 +36,13 @@ export abstract class IUser extends INameable {
 
   agent?: IAgent;
 
+  // the internal communicationID (Matrix) for the user
+  communicationID!: string;
+
+  // Indicates if this profile is a service profile that is only used for service account style access
+  // to the platform. Temporary measure, full service account support for later.
+  serviceProfile!: boolean;
```
Enderedne, line 44: I don't understand what does this have to do with matrix?

src/services/platform/matrix/events/matrix.event.adapter.room.ts
```diff
@@ -44,31 +45,45 @@ export class AutoAcceptRoomMembershipMonitorFactory {
...
+              event.getContent().body
+            }`,
+            LogContext.COMMUNICATION
+          );
+
           onMessageReceived({
             message,
             roomId: room.roomId,
-            userEmail: message.receiver,
+            communicationID: message.receiverID,
```
Enderedne, line 86: This is matrixId when we are in the scope of matrix. It should remain matrixId. Further down the line it could be wrapped in communicationId, but right now this makes things more complicated and we're adding domain definition inside of the platform one.

src/services/platform/matrix/agent/matrix.agent.service.ts
```diff
@@ -99,11 +101,19 @@ export class MatrixAgentService {
...
+    const dmRoomMap =
+      await this.matrixRoomAdapterService.getDirectMessageRoomsMap(
+        matrixClient
+      );
+    for (const matrixUsername of Object.keys(dmRoomMap)) {
+      const room = await this.getRoom(
+        matrixAgent,
+        dmRoomMap[matrixUsername][0]
+      );
+      room.receiverCommunicationsID =
```
Enderedne, line 113: The same concern mentioned below about using communicationId vs matrixId

src/services/platform/matrix/adapter-message/matrix.message.adapter.service.ts
```diff
@@ -0,0 +1,65 @@
+import { LogContext } from '@common/enums';
+import { CommunicationMessageResult } from '@domain/common/communication/communication.dto.message.result';
+import { Inject, Injectable, LoggerService } from '@nestjs/common';
+import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import { MatrixRoomResponseMessage } from '../adapter-room/matrix.room.dto.response.message';
+
+@Injectable()
+export class MatrixMessageAdapterService {
```
Enderedne, line 8: The structure will do the fine for now but will be reworked entirely when we get to message editing.

Enderedne (commented):
Nothing major, but since we spent time making the code base more consistent,  I thought it would be good to highlight those changes.

valentinyanakiev:
Still doesn't work for me, same reproduction steps, different error.

1. Drop alkemio db.
2. Create empty alkemio db.
3. Run migrations.
4. Drop (delete) homeserver.db.
5. Restart synapse server container.
6. Enable communications (COMMUNICATIONS_ENABLED=true in .env)
7. Start the server (npm start)

image

It's 100% reproducible for me/

techsmyth:
> Still doesn't work for me, same reproduction steps, different error.
> 
> 1. Drop alkemio db.
> 2. Create empty alkemio db.
> 3. Run migrations.
> 4. Drop (delete) homeserver.db.
> 5. Restart synapse server container.
> 6. Enable communications (COMMUNICATIONS_ENABLED=true in .env)
> 7. Start the server (npm start)
> 
> image
> 
> It's 100% reproducible for me/

should now be fixed; can you please try again?

valentinyanakiev (approved):

valentinyanakiev:
Hold off merging until we increase the defaults on infra-ops

PR #1460, opened 2021-10-13
Added missing code for searching project by nameID

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/alkemio.yml
+++ b/alkemio.yml
@@ -256,7 +256,7 @@ monitoring:
 ## communications ##
 # Section defining all configuration parameters / endpoints required for communication between different actors (user-user, community-user etc) in Alkemio.
 communications:
-  enabled: ${COMMUNICATIONS_ENABLED}:true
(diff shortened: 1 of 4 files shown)


techsmyth (commented):
Actually this code is not safe. The same bug is also there in Challenge + Opportunity so we may want to fix this separately. Correct code for looking up based on NameID is there in the challengeService class - basically the nameID is only guaranteed to be safe within a particular nameable scope. So looking up directly based on nameID will fail unpredictably if there are projects in two different hubs with the same nameID. 

image

hero101:
> Actually this code is not safe. The same bug is also there in Challenge + Opportunity so we may want to fix this separately.

@techsmyth Are you referring to the fact that the nameID can be `UUID_LENGTH` long? I guess refactoring it like in `project.service` the my last commit is better than what we already have.

techsmyth (approved):
Nice catch, also re the logic

PR #1461, opened 2021-10-13
add flag to User entity to allow it to be marked as a service account

techsmyth:
Extends the User profile entity to be able to mark a profile as being for a service account. Not ideal but enough for now.

Note: this bring a migration so it needs to be merged into the communications PR.


valentinyanakiev (approved):
Code looks OK, merge at your leasure.

PR #1479, opened 2021-10-19
add check that message length should not be empty; update ignore filter to pass over messages with no body

techsmyth:


Comoque1 (approved):

Comoque1:
Using the branch 2 issues seems to be addressed:
 - validation added for creation of community update without value
image
 - after deleting community update, the rest are returned successfully

PR #1487, opened 2021-10-25
added discussion; added communication entity; modified updates; modified code structure

techsmyth:
New entities:
- Communication
- Updates
- Discussion

Migration to move data from Community entity into the right fields on the new entities. 
Authorization model is applied on Updates, and also to each Discussion. 
- Note: authorization policy is not applied to each message as that would be non-scalable. There is specific logic to allow users to delete their own messages. 

The communication adapter is also now renamed / tidied up to make it cleare that it is an adpater. There is later work needed to make this pluggable. 

**Note**: need to reset authorization policy on ecoverses after running the migration to ensure that updates will work.


src/services/platform/communication-adapter/communication.adapter.service.ts
```diff
@@ -25,7 +25,7 @@ import { CommunicationSendMessageCommunityInput } from './dto/communication.dto.
 import { CommunicationSendMessageUserInput } from './dto/communication.dto.message.send.user';
 
 @Injectable()
-export class CommunicationService {
+export class CommunicationAdapterService {
```
valentinyanakiev: CommunicationAdapter. Also, if you really want to make this strongly typed, extensible and with strong contracts, you need a domain interface implemented by this adapter. E.g. getCommunityRooms is a contract of ICommunicationProvider. Then when you inject the service in another service, you have e.g. communicationService: ICommunicationProvider and provide the adapter e.g. as a token, similar to how it is done in the notification service with the alkemio adapter. Then if we want to change the underlying provider of communications, we just implement the interface and plug it in. The contract stays the same and everything chained and dependent on it stays the same.

src/services/platform/communication-adapter/communication-adapter.module.ts
```diff
@@ -20,7 +20,7 @@ import { MatrixMessageAdapterModule } from '../matrix/adapter-message/matrix.mes
     MatrixAgentPoolModule,
     SubscriptionModule,
   ],
-  providers: [CommunicationService],
-  exports: [CommunicationService],
+  providers: [CommunicationAdapterService],
```
valentinyanakiev: CommunicationAdapter - Adapters are by definition services. The design pattern is called an adapter because it acts as a wrapper between two objects. It catches calls for one object and transforms them to format and interface recognizable by the second object. In your case you are trying to adapt the Matrix interface (external) to our Communications interface (domain).

valentinyanakiev (changes requested):
See my inline comments. Also check tests - travis is failing because a contract has changed.

src/services/platform/communication-adapter/communication.adapter.service.ts
techsmyth: This has been renamed. Also took the chance to tidy up the adapter to make clearer which are public and which are private. Moved also some functions to RoomAdapter as that is where they logically belong. 

Did not put the actual communicationProvider interface in; this is a bigger design effort as there are also other typings to be decoupled.

src/services/platform/communication-adapter/communication-adapter.module.ts
techsmyth: As above this has been renamed. Also took the chance to rename all the other matrix adapters to remove service from their name.

valentinyanakiev (changes requested):
- The ecoverseID should be 36 characters long in the migration - no need to create new tables with already messed up length so we fix it after. The field length can also be limited in the Communication entity.
- Sending updates doesn't work for me. The updatesID I get when sending a message is empty. I get:
```
{"errors":[{"message":"Variable \"$msgData\" got invalid value \"\" at \"msgData.updatesID\"; Expected type \"UUID\". UUID not valid: ","locations":[{"line":1,"column":21}],"extensions":{"code":"BAD_USER_INPUT","exception":{"stacktrace":["UserInputError: UUID not valid: ","    at UUID.validate (/home/valentin/source/repos/alkemio/server/src/domain/common/scalars/scalar.uuid.ts:37:13)","    at UUID.parseValue (/home/valentin/source/repos/alkemio/server/src/domain/common/scalars/scalar.uuid.ts:14:17)","    at coerceInputValueImpl (/home/valentin/source/repos/alkemio/server/node_modules/graphql/utilities/coerceInputValue.js:128:26)","    at coerceInputValueImpl (/home/valentin/source/repos/alkemio/server/node_modules/graphql/utilities/coerceInputValue.js:54:14)","    at coerceInputValueImpl (/home/valentin/source/repos/alkemio/server/node_modules/graphql/utilities/coerceInputValue.js:105:34)","    at coerceInputValueImpl (/home/valentin/source/repos/alkemio/server/node_modules/graphql/utilities/coerceInputValue.js:54:14)","    at coerceInputValue (/home/valentin/source/repos/alkemio/server/node_modules/graphql/utilities/coerceInputValue.js:37:10)","    at _loop (/home/valentin/source/repos/alkemio/server/node_modules/graphql/execution/values.js:109:69)","    at coerceVariableValues (/home/valentin/source/repos/alkemio/server/node_modules/graphql/execution/values.js:121:16)","    at getVariableValues (/home/valentin/source/repos/alkemio/server/node_modules/graphql/execution/values.js:50:19)","    at buildExecutionContext (/home/valentin/source/repos/alkemio/server/node_modules/graphql/execution/execute.js:203:61)","    at executeImpl (/home/valentin/source/repos/alkemio/server/node_modules/graphql/execution/execute.js:101:20)","    at Object.execute (/home/valentin/source/repos/alkemio/server/node_modules/graphql/execution/execute.js:60:35)","    at /home/valentin/source/repos/alkemio/server/node_modules/apollo-server-core/src/requestPipeline.ts:571:22","    at Generator.next (<anonymous>)","    at /home/valentin/source/repos/alkemio/server/node_modules/apollo-server-core/dist/requestPipeline.js:8:71"]}}}]}

```

techsmyth:
EcoverseID length is now set to 36.

Forgot to mention in the PR description that you need to reset the authorization policy on ecoverses after the migration to ensure the authorization policies are setup properly. If you do that and then refresh the client the sending of updates will work.

valentinyanakiev (approved):

PR #1491, opened 2021-10-26
Using fire-and-forget for notifications

valentinyanakiev:
- queues need to be split
- ChallengeID needs to be extracted (first added) from the applicationData payload


techsmyth (approved):
Working in so far as the server is running and events are being sent on application created.

One question: the module / code under services/platform feels to me as actually being part of the core of the server. So different to the other functionality hosted there. To chat through as part of the next sprint.

valentinyanakiev:
> Working in so far as the server is running and events are being sent on application created.
> 
> One question: the module / code under services/platform feels to me as actually being part of the core of the server. So different to the other functionality hosted there. To chat through as part of the next sprint.

I agree it should be moved - that's why I wanted to have a 'microservices' module instead of subscriptions one. Let's chat this through where it should logically belong.

PR #1495, opened 2021-10-27
membershipUser to only show pending applications for a user

techsmyth:


Comoque1 (approved):
Tested and works fine locally.

PR #1504, opened 2021-10-30
updated logic for creating a nameID from first / last name to strip out special characters

techsmyth:
Code is updated

Note: have split the work to fix existing data into a separate issue; this PR now just stops future bad data.


valentinyanakiev (changes requested):
Doesn't work for me, signed up with valentin_yanakiev@alkem.io with username Valentin_ name Yanakiev and I get the following accessing my profile page:

image

valentinyanakiev:
There is a new PR that replaces this one.

PR #1505, opened 2021-11-01
Split queues. Refactored microservices configurations.

valentinyanakiev:
- Split alkemio-notifactions and alkemio-subscriptions queues
- Removed redundant configuration from alkemio.yml. Variables that won't be changed are moved to the service factories.
- Moved RMQ configuration to microservices section
- Renamed Subscriptions module to Microservices module and moved it under Core


src/core/microservices/subscription.pub-sub.factory.ts
```diff
@@ -0,0 +1,71 @@
...
+    })
+    .catch(err => {
+      logger.error(
+        `Could not connect to RabbitMQ: ${err}, logging in...`,
+        LogContext.NOTIFICATIONS
+      );
+      return undefined;
+    });
+
+  return pubsub;
```
hero101: The whole function can return the promise which is returning the instance of `AMQPPubSub`.

hero101 (dismissed):

techsmyth (dismissed):
Nice clean update. 

Note I pushed a commit in which it would be good if you had a look at:
- renamed one variable
- fixed subscription event on application created as that was not working

hero101 (approved):

PR #1524, opened 2021-11-05
Migration to clean nameIDs

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/migrations/1636110445512-clean-nameids.ts
+++ b/src/migrations/1636110445512-clean-nameids.ts
@@ -1,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+const tableNames = [
+  'user',
+  'project',
+  'organization',
+  'ecoverse',
+  'challenge',
+  'opportunity',
+];
+
+const buildQuery = (tableName: string) =>
+  `update ${tableName} set nameID = LOWER(REPLACE(nameID, '_', '-')) where nameID like '%_%'`;
+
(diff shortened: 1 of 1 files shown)


src/migrations/1636110445512-clean-nameids.ts
```diff
@@ -0,0 +1,23 @@
...
+  'user',
+  'project',
+  'organization',
+  'ecoverse',
+  'challenge',
+  'opportunity',
+];
+
+const buildQuery = (tableName: string) =>
+  `update ${tableName} set nameID = LOWER(REPLACE(nameID, '_', '-')) where nameID like '%_%'`;
```
techsmyth, line 15: There are multiple other characters that may need to be replaced: (, ), ...suggest to get a list of all nameIDs from prod to see what characters are in nameIDs that should not be and then update this migration accordingly

techsmyth (changes requested):
Nice clear approach. 

Set of characters will need to be expanded that are removed...

src/migrations/1636110445512-clean-nameids.ts
valentinyanakiev: I checked production, there are messed up data with the following format:
- with _
- with ( and )
- without any firstname and lastname (captured in a separate story)
Replace _ with - and remove ( and ) and we should be good.

valentinyanakiev (approved):

techsmyth (approved):

PR #1541, opened 2021-11-17
Wallet Manager - WIP

valentinyanakiev:
Things done:
- Abstracted notificationsFactory to be generic microservices factory where different configurations can be provided
- Moved different microservice provider tokens to a separate file
- Added a new service for wallet management
- Stripped out ssi agent code - moved to the wallet manager
- Replaced ssi agent calls to call wallet manager microservice via message pattern. 

Note: split out into a separate issue:
- test out callbacks / integration with wallet manager. Replies are observables.
- currently with ssi enabled the requests don't pass the ory strategy - that needs to be investigated and fixed

To discuss:
- microservices module to not have calls into the domain data; the payload builder should be a separate service
- what happens when wallet mgr is not available and there is no message returned? Currently it is hanging.


techsmyth (approved):
Moving to develop to avoid maintenance overhead. Still needs work related to interaction to wallet mgr responses.

PR #1543, opened 2021-11-19
reworked bootstrap account generation to use user service for nameID generation; made random element optional

techsmyth:


Comoque1 (approved):
Works fine locally.

PR #1557, opened 2021-11-29
fix user deletion; renamed field on user preference

techsmyth:
Ensure that user preferences are removed when a user is deleted
Renamed the field for referring to the user preference definition.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alkemio-server",
-  "version": "0.15.3",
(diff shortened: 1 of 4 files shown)


Comoque1:
Tested OK on Test env using this branch.
Scenarios tested were:
 - Remove user
 - Add user to organization members and groups; adding user to hub and challenge members; remove the user
 - Add user to hub community and remove the hub

Comoque1 (approved):

PR #1567, opened 2021-12-01
initial pass at removing two instances of eager loading

techsmyth:
user --> profile
communication ==> discussions


valentinyanakiev (approved):
Code makes sense. Created a new discussion, listed discussions, deleted a discussion seamlessly. Also, user profile seemed to be loaded OK. Overall smoke testing it manually was fine so I suggest we merge it.

PR #1569, opened 2021-12-03
Added 3 new events

valentinyanakiev:
Didn't have time to test the new events, hence the draft.

Added the following:
- Payload builder to build different types of payloads. The focus is on reusing code/limiting code duplication and efficiency. I have gone for using the repositories directly instead of the services to limit the data load and increase efficiency (events should be optimized as by default we'd expect a lot of them). Data was lazy loaded mostly everywhere so I had to do it for at least some of the classes anyway, as there were relationships that were missing (e.g. Update / Discussion -> Communication -> Community can be looked up in the database but requires changes in the typeORM entity relationships to have the tree traversable in all directions. I want a smaller tree.).
- Added new exception type
- Added new enum - EventType. We need to decide what to do with it as we have a lot of overlap between EventType / SubscriptionType and UserPreferenceType.

Code is largely untested, albeit logical. Needs to be tested end-to-end, with all flows.


quickstart-services.yml
```diff
@@ -183,7 +183,7 @@ services:
       - 'host.docker.internal:host-gateway'
     container_name: alkemio_dev_notifications
     hostname: notifications
-    image: alkemio/notifications:v0.3.2
+    image: vyanakiev/alkemio-notifications:v0.4.6
```
techsmyth: update before finalization...

src/domain/common/notifications/notifications.module.ts
```diff
@@ -0,0 +1,21 @@
...
+    TypeOrmModule.forFeature([Ecoverse]),
+    TypeOrmModule.forFeature([Challenge]),
+    TypeOrmModule.forFeature([Opportunity]),
+    TypeOrmModule.forFeature([Community]),
+    TypeOrmModule.forFeature([Discussion]),
+  ],
+  providers: [NotificationsPayloadBuilder],
+  exports: [NotificationsPayloadBuilder],
+})
+export class NotificationsModule {}
```
techsmyth: This whole module belongs I think under services, not directly in the domain model.

techsmyth (commented):
Nice to have all event payloads in one place / service. 

It feels very odd to be going in this way to the repositories in the domain model . This kind of access is mostly via services - so I think this module needs to move.

quickstart-services.yml
valentinyanakiev: I need to run the whole flow first to see whether I need extra changes there so I can update + release notifications and use it here.

src/domain/common/notifications/notifications.module.ts
valentinyanakiev: Actually, I was thinking of removing the module altogether and adding the builder to the microservices module, as it logically belongs there (I think). Then we'll have all the logic required to send an event in one module and will import only one module instead of two (we always import the two modules (Microservices / Notifications) where we send a notification.

valentinyanakiev:
> Nice to have all event payloads in one place / service.
> 
> It feels very odd to be going in this way to the repositories in the domain model . This kind of access is mostly via services - so I think this module needs to move.



> Nice to have all event payloads in one place / service.
> 
> It feels very odd to be going in this way to the repositories in the domain model . This kind of access is mostly via services - so I think this module needs to move.

Event-driven code usually goes directly through SQL / repositories to minimize load and optimize performance. Having the code written as it is now was intentional, I even started writing queries directly with query runner to optimize it even more (but I tested the performance gain of SELECT * from Ecoverse | Challenge | Opportunity and SELECT id from Ecoverse | Challenge | Opportunity and the difference was mere 5-7% so I went for the repositories after discussing with Svetoslav). If we go through the services for the events we start relying on the typeorm definitions and typeorm - generated queries, which leads to bloated dependencies and (potentially) excessive queries. Events need to be performant and the builder goal is to share performant queries for extracting the data required for the payload of events, assuming the hypothesis that there will be many events.

src/domain/common/notifications/notifications.module.ts
techsmyth: Interesting, so basically the assertion is that Microservies **is** notifications. Are there other types of microservices that we might want to use in the future? I can definitely see the microservices as being part of the core, but building of notification payloads for me just feels different.

techsmyth:
Re comment above on way to go to the repositories, I think I could have framed it better. I actually agree that you go the way it is now, that makes sense. And indeed we may want to do that kind of approach in more places. 

It is more that the fact this is going across multiple repositories for me is a signal that it belongs in a separate service outside of the domain model.

techsmyth (dismissed):
Working nicely in that the events are being generated and received by the notification service.

techsmyth (approved):

Issue #1571, opened 2021-12-04
BUG:  Address race condition on triggered lifecycle events


techsmyth:
**Describe the bug**
Events that are triggered by XState are fire and forget. So invoking an event on a lifecycle can potentially return
before all the actions (as defined in the lifecycle definition) that were triggered by the event have completed.

Example can be seen in the CanvasCheckoutOptionsProvider: in that case the events should all be quite fast a short wait is sufficient.

**Expected behavior**
The event from Lifecycle service should only return once the triggered actions have completed? Not actually sure as there are some underlying semantics we need to design for properly.

Issue #1570, opened 2021-12-04
Initialize Lifecycles on creation


techsmyth:
## Background
As a user I want that the lifecycle is initialized when created

## Acceptance criteria
- [ ] The lifecycle gets initialized + put into the initial state.

## Additional context
The initialization currently only happens at the point when the first event is sent to the lifecycle.

This is a logic / design issue that needs to be addressed

PR #1572, opened 2021-12-04
Canvas entities on Context; Checkout lifecycle

techsmyth:
Adds Canvas entities to Context
Extend Canvas entity with flag for template
Add lifecycle to manage the checkout / checkin of Canvases
Migration added to create and set an authorization policy on existing canvases (on ecosystem models)

CanvasCheckout entity to manage the checkout lifecycle
CanvasCheckout is lazy loaded, so that existing canvas entities are ok without it. 

Two issues raised related to lifecycles:
- https://github.com/alkem-io/server/issues/1571
- https://github.com/alkem-io/server/issues/1570

Also we need to find a solution for creating lifecycles via migrations.  For now the canvasCheckout is created if not already present. Raised in the backlog under: https://github.com/alkem-io/server/issues/1573


src/migrations/1638441120817-canvas-checkout.ts
```diff
@@ -0,0 +1,120 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { RandomGenerator } from 'typeorm/util/RandomGenerator';
+
+export class canvasCheckout21638441120817 implements MigrationInterface {
+  name = 'canvasCheckout21638441120817';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE \`canvas_checkout\` (\`id\` char(36) NOT NULL, \`createdDate\` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6), \`updatedDate\` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), \`version\` int NOT NULL, \`canvasID\` varchar(36) NOT NULL, \`lockedBy\` varchar(36) NOT NULL, \`status\` varchar(255) NOT NULL DEFAULT 'available', \`authorizationId\` char(36) NULL, \`lifecycleId\` char(36) NULL, UNIQUE INDEX \`REL_353b042af56f01ce222f08abf4\` (\`authorizationId\`), UNIQUE INDEX \`REL_bd3c7c6c2dbc2a8daf4b1500a6\` (\`lifecycleId\`), PRIMARY KEY (\`id\`)) ENGINE=InnoDB`
```
valentinyanakiev: Does the status need to be varchar(255)? It's a bit excessive.

src/domain/common/lifecycle/lifecycle.service.ts
```diff
@@ -82,14 +92,20 @@ export class LifecycleService {
...
 
-    const newStateStr = JSON.stringify(machineService.state);
+    const stateToHydrate = interpretedLifecycle.state;
+    // Note: https://github.com/statelyai/xstate/discussions/1757
+    // restoring state has the hydrated actions, which unless removed will be executed again
+    stateToHydrate.actions = [];
+    const newStateStr = JSON.stringify(stateToHydrate);
+
+    // Todo: do not stop as this triggers an exit action from the last state.
+    //machineService.stop();
```
valentinyanakiev, line 106: So I guess my hunch that the machine needs to be singleton was on the right path...

src/domain/common/lifecycle/lifecycle.service.ts
```diff
@@ -82,14 +92,20 @@ export class LifecycleService {
...
-      }) event '${eventName}: from state '${startState}' to state '${machineService.state.value.toString()}', parentID: ${parentID}`,
+      }) event '${eventName}: from state '${restoredState.value.toString()}' to state '${machineService.state.value.toString()}', parentID: ${parentID}`,
       LogContext.LIFECYCLE
     );
 
-    const newStateStr = JSON.stringify(machineService.state);
+    const stateToHydrate = interpretedLifecycle.state;
+    // Note: https://github.com/statelyai/xstate/discussions/1757
+    // restoring state has the hydrated actions, which unless removed will be executed again
+    stateToHydrate.actions = [];
```
valentinyanakiev, line 102: Ok, reading the discussion above I see what was the issue, and the solution (which is the only solution for us now).
A properly defined architectural solution restoring the whole machine without changing it can be done with CQRS, but we are way off it for now. Alternatively, XState v5, if they decide to implement it.

src/domain/common/canvas-checkout/canvas.checkout.lifecycle.options.provider.ts
```diff
@@ -0,0 +1,177 @@
...
+        LogContext.LIFECYCLE
+      );
+    }
+  }
+
+  private CanvasCheckoutLifecycleMachineOptions: Partial<
+    MachineOptions<any, any>
+  > = {
+    actions: {
+      availableEntry: (_, __) => {
```
valentinyanakiev, line 80: Weren't these log messages just to debug it?

valentinyanakiev (changes requested):
I see that required some reading of XState - solution seems OK (actually possibly the only viable solution with what we have architecturally at the moment), some minor comments / clean-up required.

src/migrations/1638441120817-canvas-checkout.ts
techsmyth: Actually I think it probably should just be 'text' so that it is variable?

src/domain/common/lifecycle/lifecycle.service.ts
techsmyth: It is not that the machine needs to be a singleton, there can be many machines in parallel. It is to do with the way the state is hydrated, I think they will change it.

src/domain/common/lifecycle/lifecycle.service.ts
techsmyth: The alternative design would be to always have a machine active for every entity, and this would be a very interesting approach. 

However even in that case we would still need to be able to hydrate / rehydrate in a reliable way. 

My take is that this is simply a flaw in XState, basically implying that the kind of scenario we are using it for is not what it is designed for (long running server processes).

src/domain/common/canvas-checkout/canvas.checkout.lifecycle.options.provider.ts
techsmyth: Yes, but really super useful so I put in a flag to be able to disable them while leaving them in. Also useful to be able to explain to others what is happening. 

I think we may find this kind of lifecycle "debugging" is critical for us to really be able to manage these lifecycles in the near-medium term as they are way to much of a black box at the moment.

valentinyanakiev (approved):

PR #1574, opened 2021-12-06
Migration fix

valentinyanakiev:


Comoque1:
I've run the migration fix on test env, and for now it seems to work.

Comoque1 (approved):

PR #1597, opened 2021-12-10
Notification preferences

valentinyanakiev:
Added the following definitions + migration:
- NOTIFICATION_COMMUNICATION_DISCUSSION_CREATED_ADMIN
- NOTIFICATION_COMMUNICATION_UPDATE_SENT_ADMIN


techsmyth (commented):
Just ran the migration locally but I do not see the new preferences showing up in my profile:

image

atstoyanov:
There is a filter in the UI.

techsmyth (changes requested):
The migration is not actually creating the new preferences for users, only for those without any preferences. This migration needs to check if each user has the preferences being added.

techsmyth (approved):
Working now

PR #1610, opened 2021-12-15
subscriptions for communications

techsmyth:
split messageUpdate subscription into two: one for updates, one for discussions
reworked to return the updatesID or discussionID (instead of room ID which is no longer relevant)
moved communication related subscriptions into CommunicationModule; 
added error type for communication subscriptions
version number is bumped

Note: the communityId field is not currently returned. Please let me know how big an issue this is...


valentinyanakiev (approved):
I am merging this with the caveat that the filtering is not operational.
In my honest opinion it is inverted, const forCurrentUser =
        communication ID of the event sender--->payload.communicationID === context.req?.user?.communicationID; <---- communicationID of the subscriber
if (!forCurrentUser) return false; <---- this means, continue only if the event I created is an event I subscribed for. It should be the opposite, and maybe even this is redundant, because if we have common logic updating based on the subscription results, we want all users to be updated.

PR #1611, opened 2021-12-16
Fixed saving large canvases

valentinyanakiev:
- Switched canvas value from TEXT (max 65,535 bytes) to LONGTEXT(max 4,294,967,295 bytes).
- Increased max json payload size in express to accommodate larger values (check https://stackoverflow.com/questions/50988891/payloadtoolargeerror-with-limit-set and https://github.com/nestjs/nest/issues/529)


techsmyth (approved):
Working well for me, no errors anymore on canvas size.

PR #1625, opened 2021-12-22
Communication subscriptions from domain model

techsmyth:
Subscriptions related to communications are now sent directly from the domain model. Two new subscriptions:
* communicationUpdateMessageReceived
* communicationDiscussionMessageReceived
They are now also within the entity folder within the code structure, thus replacing the general CommunicationSubscription file. 

The triggering of domain events based on events on the Matrix room timeline has been disabled. 

Each of the above methods takes an optional parameter to restrict the entities to subscribe to:
* If the parameter is supplied then authorization is checked for each requested entity
* If the parameter is not supplied then authorization is checked as part of the filter; clearly this is heavy and will need to be optimized as we see load increasing

The following has been tested:
* a user that is not a member of a private hub does not see updates sent to that community
* a user that is not a member of a public hub **does* see updates sent to that community
* two users can both be subscribed to updates at the same time for a given public community and they both see the update message that was sent

Note: the current implementation is likely to put a heavy load on the server with a moderate number of subscription events and active subscriptions due to the check for authorization on every event. This needs to be optimized and is captured in https://github.com/alkem-io/server/issues/1632


src/domain/communication/communication/communication.service.ts
```diff
@@ -124,7 +124,15 @@ export class CommunicationService {
     communication.discussions?.push(discussion);
     await this.communicationRepository.save(communication);
 
-    // Set the Matrix membership for notifications
+    // Trigger a room membership request for the current user that is not awaited
```
valentinyanakiev, line 127: How is this not awaited?

valentinyanakiev (changes requested):
The code could work, assuming matrix messages and updates are awaited by their Alkemio wrappers (communications, adapters etc). There is a comment that a method should not be awaited but is actually awaited which is confusing. 

All API code / subscription events are based on the assumption that everything below them is awaited (if it isn't awaited we will send e.g. MESSAGE_RECEIVED when the message was sent but we wouldn't know what happened in Matrix), which could lead to problems if it isn't the case - I haven't checked all code paths, but that is a risky approach and needs to be looked at / tested carefully.

Functionally, update subscriptions were working OK for me from the API, but Discussion Updates were received every 2nd event, and that needs to be fixed.

valentinyanakiev (approved):

PR #1629, opened 2021-12-29
Authorization refresh

techsmyth:
Framework extensions
* Added another rule type: privilegeRule, which allows granting of a set of privileges based on holding another privilege. This is primarily for roles that are admin, as if they hold the generic CREATE then local create such as CREATE_CANVAS should also apply to them.
* Add notion of authorization policy rules that are not inheritable
* Extended the set of authorization privileges to include: CREATE_CANVAS, CREATE_HUB, CREATE_ORGANIZATION, READ_USERS

Platform level Authorization:
* Created platform level authorization policy
* Exposed platform level authorization policy as a top level query

Usage
* Added ability for hub admins and challenge admins to create organizations
* Allow members of a community to create canvases as well as the admins (requires client update to use)
* Made all user related top level queries depend on the READ_USERS privilege, instead of creating multiple local authorization policies

There is also a related client branch: server-1618 that is updated to use the new CREATE_CANVAS privilege to allow members to create / edit canvases.

NOTE: to use this branch you need to both run migrations and reset the authorization policy on Hubs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "alkemio-server",
-  "version": "0.16.1",
+  "version": "0.16.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "alkemio-server",
-      "version": "0.16.1",
+      "version": "0.16.2",
       "license": "EUPL-1.2",
       "dependencies": {
         "@nestjs/axios": "^0.0.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alkemio-server",
-  "version": "0.16.1",
+  "version": "0.16.2",
   "description": "Alkemio server, responsible for managing the shared Alkemio platform",
   "author": "Alkemio Foundation",
   "private": false,
--- a/quickstart-services.yml
+++ b/quickstart-services.yml
@@ -183,7 +183,7 @@ services:
       - 'host.docker.internal:host-gateway'
     container_name: alkemio_dev_notifications
     hostname: notifications
-    image: alkemio/notifications:v0.4.2
+    image: alkemio/notifications:v0.4.5
     environment:
       - RABBITMQ_HOST
       - LOGGING_CONSOLE_ENABLED
--- a/src/common/enums/authorization.privilege.ts
+++ b/src/common/enums/authorization.privilege.ts
@@ -6,6 +6,7 @@ export enum AuthorizationPrivilege {
   UPDATE = 'update',
   DELETE = 'delete',
   GRANT = 'grant', // allow the issuing / revoking of credentials of the same type within a given scope
+  CREATE_CANVAS = 'createCanvas',
 }
 
 registerEnumType(AuthorizationPrivilege, {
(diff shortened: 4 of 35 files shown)


valentinyanakiev:
There is a bug (potentially more than 1, but so far I have caught 1) in the implementation.
It doesn't seem to cover user applications.

To reproduce (on an environemnt with client + server on branch server-1618, migrations applied, authorization policies reset for all entities from cli dev branch):
1. Log in with a normal account
2. Apply to a hub
3. Logout
4. Log in with admin user
5. Try to manage the applications:

image

Expected behavior is to have the application visible

Edit: Same behavior is observed in Challenges, I suspect in Opportunities as well.

valentinyanakiev (changes requested):
The code largely works, apart from the bug I have reported (needs fixing as it blocks approving any application so it is high severity issue).

The code as lines of code looks OK, but I am confused by the design. I feel it mixes a paradigm like DDD (Domain-Driven Design) authorization, used in platforms / frameworks such as CBAS. For example, as per CBAS documentation:

```typescript
await cbas.evaluateInvocation(invocation)

const expectedReturnValue = {
  resource: {
    "type": "UBS-backup",
    "backupID": "aaabbbccc",
    "vaultID": "bbbcccfff",
    "locations": ["https://ubs.jolocom.io"]
  },
  privileges: ['update-backup', 'delete-backup', 'read-backup'],
  issuer: 'did:example:ubsServer',
  subject: 'did:example:ubsUser',
  constraints: [{
    dailyUploadLimit: "1024MB",
  }]
}
```

The privileges are defined as update-backup, delete-backup etc, and the subject (the actor in our case) is the user (user-agent in our case). This approach is DDD as the authorization is on the level of the domain (user) that can do a specific action (e.g. update-backup) and doesn't follow our largely CRUD model.

We, on the other hand, have largely used a more orthodox CRUD-based credential-based authorization, where policies are based on, largely, resource ID (or lack of it) and a set of predefined operations, largely applicable to the entities we have.

An approach somewhere in between is the AWS approach, where a policy looks like:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "",
            "Effect": "Allow",
            "Action": [
                "ec2:DescribeInternetGateways",
                "ec2:DescribeAddresses",
                "ec2:DescribeAccountAttributes"
            ],
            "Resource": "*"
        }
    ]
}
```

Here there is a generic framework that can accommodate both DDD and CRUD approach, as Users can be part of User Groups, and Users / UserGroups can have roles consisting of multiple Policies, as the one defined above.

The design flaw, in my opinion, is that we end up with a mixed approach that's hard to comprehend and/or follow and will lead to many edge cases (e.g. Policy vs Credential, in our terminology, which trumps what, where, and why). Requires a discussion on what our desired state looks like.

Other than that, fixing the bug should fix our immediate shortcomings and can be applied as a fix - that's what I am requesting changes for. The description of the bug is in the PR.

techsmyth:
I just tried the bug reported above, and it worked fine for me. No issue with seeing the application, either directly on graphql or via api. Would need to see more details about the hub / challenge being applied to - and what the authorization policy is on both the hub and the application itself. Can you share more of the request / response and also the authorization policy contents?

valentinyanakiev (approved):

PR #1651, opened 2022-01-07
Private + Public API + Playground

valentinyanakiev:


techsmyth (commented):
Really nice update :)

For the sample graphql queries:
- These should not be under src as a graphql folder. Maybe under configuration?
- The queries should be really simple, and I would probably remove the configuration one

Regarding the routing setup:

if we are tidying up the routing, then maybe we should also rename this one:
image

and should this one not also have another name:
image

And for the routing, I would suggest to group the ones that match based on path separately from the ones that match based on ports, to make it easier to read through / understand

techsmyth (approved):
Looks nice and clean

PR #1664, opened 2022-01-14
Initial pass at assigning / retrieving VC for state transition

techsmyth:
To validate this flow:
- Create a new Challenge as global admin
- Login to another browser as a user without admin privileges
- Using normal user attempt to update the state of the Challenge:
- 
image

- This will fail
- As admin, assign the normal user the VC to allow updating the state of that Challenge:
image

Repeat the attempt to update state as a normal user, it succeeds :)

image


src/domain/agent/verified-credential/dto/verified.credential.dto.result.ts
```diff
@@ -13,10 +13,10 @@ export class VerifiedCredential {
   })
   issuer: string;
 
-  @Field(() => Date, {
+  @Field(() => String, {
     description: 'The time at which the credential was issued',
   })
-  issued?: Date;
+  issued?: string;
```
valentinyanakiev, line 19: Why string and not date?

valentinyanakiev, line 19: Why string and not date?

valentinyanakiev (commented):
Fixed one error message + I have a question why a date was moved to be a string?

src/domain/agent/verified-credential/dto/verified.credential.dto.result.ts
techsmyth: Date was not resolving properly, giving that null error that we saw. It may be that the value needs to be converted in the wallet mgr before sending back, but in terms of getting it working this was the fastest route....

valentinyanakiev (approved):

PR #1670, opened 2022-01-16
Visuals as first class entity with meta data 

techsmyth:
Added new Visual entity that is a wrapper around an image with meta data
Migration to move existing images information to the new structure.
removed uploadAvatar mutation
mutation to directly edit the uri on a Visual

For later:
- look at centralizing the definition of the various image types

Client:
- switch to using the new uploadImageOnVisual mutation
- switch to using the new avatar on profile
- switch to using the visuals on context
....


valentinyanakiev:
After running it:

(node:3636375) UnhandledPromiseRejectionWarning: Error: Unable to create profiles ER_NO_DEFAULT_FOR_FIELD: Field 'minHeight' doesn't have a default value
    at BootstrapService.bootstrapEcoverse (/home/valentin/source/repos/alkemio/server/src/core/bootstrap/bootstrap.service.ts:64:13)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at bootstrap (/home/valentin/source/repos/alkemio/server/src/main.ts:25:3)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:3636375) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:3636375) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

techsmyth:
@valentinyanakiev issue you found is fixed, please try again at your leisure...

valentinyanakiev:
Seems ok, but let's hold off merging until we fix uploading avatars

hero101:
Removed some old cases and redundant code but a mutation `updateVisual` still exists and I believe it should be removed.

hero101 (approved):
Tested with the client-side changes

PR #1685, opened 2022-01-26
added logic to update the communications usage of matrix if not set

techsmyth:


Comoque1:
Test using the PR branch on test env

1. Had Hub/Challenge/Opportunity created, before the fix.
2. Created Hub/Challenge/Opportunity  after the fix was deployed.
3. Discussions with comments were created for the entities created before the fix deployment
4. Removed `communicationGroupId` and `communicationRoomId` from both `communication` and `updates` tables.
5. Restarted the server service
6. All discussions created in the first group of entities are available/editable
7. New discussions and updates were successfully created to all entities

valentinyanakiev (approved):

PR #1687, opened 2022-01-28
Subscriptions on discussions

techsmyth:
Two new subscriptions:
- discussionUpdated
- discussionMessageReceived2

The previous subscription is still there pending the client being updated

The logic for both the new subscriptions is significantly simpler, and should perform massively better at heavy load than the previous implementation.

In addition the discussionUpdated is also now triggered for the following mutations:
- discussionCreated
- discussion message deleted
So the discussions overview page will remain much more up to date. 

To do: 
- client updated for the new discussions
- old subscription removed from the server


valentinyanakiev (approved):

valentinyanakiev (dismissed):

valentinyanakiev:
Code looks OK, haven't tested it functionally - Evgeni is going over the functionality

Comoque1 (approved):

PR #1690, opened 2022-01-31
extended the users + organizations queries

techsmyth:
with optional parameters to allow limiting the number of users or orgs returned and what the selection criteria should be random or not

### DIFF
--- a/src/domain/community/organization/organization.resolver.queries.ts
+++ b/src/domain/community/organization/organization.resolver.queries.ts
@@ -1,5 +1,5 @@
 import { UUID_NAMEID } from '@domain/common/scalars';
-import { Args, Query, Resolver } from '@nestjs/graphql';
+import { Args, Float, Query, Resolver } from '@nestjs/graphql';
 import { Profiling } from '@src/common/decorators';
 import { IOrganization } from './organization.interface';
 import { OrganizationService } from './organization.service';
@@ -13,8 +13,28 @@ export class OrganizationResolverQueries {
     description: 'The Organizations on this platform',
   })
   @Profiling.api
-  async organizations(): Promise<IOrganization[]> {
-    return await this.organizationService.getOrganizations();
+  async organizations(
+    @Args({
+      name: 'limit',
+      type: () => Float,
+      description:
+        'The number of Organizations to return; if omitted return all Organizations.',
+      nullable: true,
+    })
+    limit: number,
+    @Args({
+      name: 'randomSelection',
+      type: () => Boolean,
+      description:
+        'If more organizations than the limit specified then return a random or fixed selection.',
+      nullable: true,
+    })
+    randomSelection: boolean
+  ): Promise<IOrganization[]> {
+    return await this.organizationService.getOrganizations(
+      limit,
+      randomSelection
+    );
   }
 
   @Query(() => IOrganization, {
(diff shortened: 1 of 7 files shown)


src/domain/community/organization/organization.resolver.queries.ts
```diff
@@ -13,8 +13,28 @@ export class OrganizationResolverQueries {
...
+    @Args({
+      name: 'limit',
+      type: () => Float,
+      description:
+        'The number of Organizations to return; if omitted return all Organizations.',
+      nullable: true,
+    })
+    limit: number,
+    @Args({
+      name: 'randomSelection',
```
hero101: `shuffle` would be more appropriate. I would avoid complex, long parameter names like that

src/domain/community/organization/organization.service.ts
```diff
@@ -307,9 +308,28 @@ export class OrganizationService {
...
-    const organizations = await this.organizationRepository.find();
-    return organizations || [];
+  async getOrganizations(
+    limit?: number,
+    randomSelection = false
+  ): Promise<IOrganization[]> {
+    const organizations: IOrganization[] =
+      await this.organizationRepository.find();
+    if (!organizations) return [];
+    if (!limit || organizations.length < limit) return organizations;
```
hero101: you are not taking in account the `randomSelection` param

src/domain/community/organization/organization.service.ts
```diff
@@ -307,9 +308,28 @@ export class OrganizationService {
...
+
+    // Need to restrict the set of users to return
+    if (randomSelection) {
+      const randomIndexes = generateRandomArraySelection(
+        limit,
```
hero101: in order to get the code executed through here, you need to pass the `limit` param.

src/domain/community/user/user.service.ts
```diff
@@ -475,8 +476,21 @@ export class UserService {
     return user;
   }
 
-  async getUsers(): Promise<IUser[]> {
-    return (await this.userRepository.find({ serviceProfile: false })) || [];
+  async getUsers(limit?: number, randomSelection = false): Promise<IUser[]> {
+    const result = await this.userRepository.find({ serviceProfile: false });
+    if (!result) return [];
+    if (!limit || result.length < limit) return result;
```
hero101: not taking in account the `randomSelection` param again

hero101 (changes requested):

src/domain/community/organization/organization.resolver.queries.ts
techsmyth: addressed in9a57bcd8860c2ff4031bed6d103313c4b6c9ba8f

src/domain/community/organization/organization.service.ts
techsmyth: addressed in 9a57bcd8860c2ff4031bed6d103313c4b6c9ba8f

techsmyth: Note: shuffle paramter is only applied in the case where a limit is set.

src/domain/community/organization/organization.service.ts
techsmyth: see above

src/domain/community/user/user.service.ts
techsmyth: see above

hero101 (dismissed):

hero101 (approved):

PR #1694, opened 2022-02-01
first pass at domain model update

techsmyth:
Comments entity that allows the managing of a particular set of comments

New subscription for messages sent to a particular Comments entity (not tested).

Aspect entity expanded with the following fields:
- type
- comments
- displayName
- nameID
- createdBy
- banner
- bannerNarrow
- references
- tagset

New mutations for adding aspects to context, references to aspects, adding / removing comments

Authorization 
- new CREATE_ASPECT privilege to determine if a user is allowed to create a new aspect
- authorization service for aspects to propagate the authorization to child entities. 

Migration is in place to modify the database to support above entity contents. 

Sample queries / mutations added.

Note: existing Aspect data is not migrated as there are essentially no aspects on production


techsmyth:
@Comoque1 good if you can push this new entity a bit to surface up any cases I have not considered early. In particular authorization probably needs some testing.

Comoque1 (approved):

PR #1699, opened 2022-02-03
look up contextID also on challenge / opportunity

techsmyth:


Comoque1:
Different error is thrown with this branch (probably due to: `const challenge = await this.ecoverseRepository`), when intend to create aspect on Challenge/Opportunity.
image

techsmyth:
yes that was wrong, fixed now...

Comoque1 (approved):

PR #1702, opened 2022-02-04
Removed application eager loading on community

valentinyanakiev:
Seems to be much faster with real data without it


techsmyth:
I think we also should make groups lazy loaded, they are also a heavy tree with profiles, avatars etc

valentinyanakiev:
Created a new branch off master + new PR

PR #1706, opened 2022-02-04
Remove eager loading

techsmyth:
Removes eager loading from the following relationships:
- community.groups
- profile.references
- profile.tagsets
- profile.avatar

Also fixes a bug related to removing avatar when removing a profile.

Moved to use the avatar definition on the visual service also on the profile.

### DIFF
--- a/src/domain/community/community/community.service.ts
+++ b/src/domain/community/community/community.service.ts
@@ -90,14 +90,17 @@ export class CommunityService {
   }
 
   // Loads the group into the Community entity if not already present
-  getUserGroups(community: ICommunity): IUserGroup[] {
-    if (!community.groups) {
+  async getUserGroups(community: ICommunity): Promise<IUserGroup[]> {
+    const communityWithGroups = await this.getCommunityOrFail(community.id, {
+      relations: ['groups'],
+    });
+    if (!communityWithGroups.groups) {
       throw new EntityNotInitializedException(
         `Community not initialized: ${community.displayName}`,
         LogContext.COMMUNITY
       );
     }
-    return community.groups;
+    return communityWithGroups.groups;
   }
 
   async getCommunityOrFail(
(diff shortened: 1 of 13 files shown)


src/domain/community/community/community.service.ts
```diff
@@ -90,14 +90,17 @@ export class CommunityService {
   }
 
   // Loads the group into the Community entity if not already present
-  getUserGroups(community: ICommunity): IUserGroup[] {
-    if (!community.groups) {
+  async getUserGroups(community: ICommunity): Promise<IUserGroup[]> {
+    const communityWithGroups = await this.getCommunityOrFail(community.id, {
```
valentinyanakiev, line 94: I really wanted to avoid having to do this, this is likely going to be a quite long and slow query, and we'll execute it every time we get the user groups, and we didn't need to before.

src/domain/community/profile/profile.service.authorization.ts
```diff
@@ -3,16 +3,25 @@ import { InjectRepository } from '@nestjs/typeorm';
...
   constructor(
     private authorizationPolicyService: AuthorizationPolicyService,
+    private profileService: ProfileService,
     @InjectRepository(Profile)
     private profileRepository: Repository<Profile>
   ) {}
 
-  async applyAuthorizationPolicy(profile: IProfile): Promise<IProfile> {
+  async applyAuthorizationPolicy(profileInput: IProfile): Promise<IProfile> {
+    const profile = await this.profileService.getProfileOrFail(
```
valentinyanakiev, line 18: Same logic as above, we make an extra call that is likely going to be a quite slow query.

valentinyanakiev (approved):
I hope this is faster. It will optimize some long queries, but add queries in places where we hadn't before. Let's try it out and see how it performs. Code looks OK otherwise - it seems relations are replaced properly.

valentinyanakiev (approved):

PR #1707, opened 2022-02-04
do not store the event with the authorization data with the hydrated state of the lifecycle

techsmyth:


valentinyanakiev (approved):
Needs to be functionally tested before merging. The code change makes sense.

Comoque1:
The issue is still reproducible for newly created entity, when change the lifecycle state 
image

In addition, evaluation of `machineState` on production is not possible, as the field is not exposed to the API
image

techsmyth:
Merging + closing this issue as it stops future leakage. The cleaning up of existing data I will raise in a separate issue.

PR #1709, opened 2022-02-07
explicitly remove the detailed fields

techsmyth:


Comoque1 (approved):

Comoque1:
Tested with Challenge, Opportunity, Application and Canvases events - now credentials are not preserved.

PR #1711, opened 2022-02-08
Expanded search on challenges, opportunities to include aspects

valentinyanakiev:
- Replaced title with displayName as there is no title.
- The migration is failing when trying to apply it on a db restored from demo / hub. Comments table exists already.


techsmyth (approved):
Search for me is working in terms of picking aspects on Challenges.

techsmyth:
This can be merged; the comments table migration is a separate issue...

PR #1715, opened 2022-02-09
initial pass at adding templating support to Hubs; WIP

techsmyth:
Hubs now have a template, which has a set of Aspect templates.

Ability to specify the template contents via the updateEcoverse mutation. 

Migration is in place. 

PLUS
- fixed a bug with the setting of authorization policy on user profiles + avatars


valentinyanakiev (approved):
Creating + retrieving templates works fine for me.

PR #1743, opened 2022-02-17
Ecoverse -> Hub

valentinyanakiev:


techsmyth:
Just tried to run the migration but got an error: 
image

Resetting db and running went smoothly

.scripts/rename.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
```
techsmyth, line 2: Not sure this should be in the code base?

src/common/constants/default.hub.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DEFAULT_HUB_NAMEID = 'eco1';
+export const DEFAULT_HUB_DISPLAYNAME = 'Default Hub';
+export const DEFAULT_HOST_ORG_NAMEID = 'Eco1Host';
```
techsmyth, line 3: We should raise a card to update this separately.

techsmyth (approved):
A couple of minor ones, and the migration needs to be checked. But otherwise looks good.

PR #1776, opened 2022-03-06
made preferences re-usable so not just tied to Users

techsmyth:
Migration to rename from UserPreference to Preference
Module moved to be under common
Classes / interfaces etc renamed
Update mutation to specify the Preference ID instead of the userID + type.


valentinyanakiev (approved):
Basic flows seem to be working for me.

PR #1797, opened 2022-03-11
added an enum to combine preference types user + hub - fixes issue wi…

techsmyth:
…th obtaining type field on a definition


src/common/enums/preference.type.ts
```diff
@@ -0,0 +1,23 @@
+import { registerEnumType } from '@nestjs/graphql';
```
valentinyanakiev, line 1: You already have the preferences, try replacing with the following, avoiding code duplication and forcing future compatibility when enums are changed:

```suggestion
import { registerEnumType } from '@nestjs/graphql';
import { UserPreferenceType } from './user.preference.type';
import { HubPreferenceType } from './hub.preference.type';

export const PreferenceType = {
  ...UserPreferenceType,
  ...HubPreferenceType,
};

export type PreferenceType = UserPreferenceType | HubPreferenceType;

registerEnumType(PreferenceType, {
  name: 'PreferenceType',
});

```

src/domain/common/preference/preference.definition.interface.ts
```diff
@@ -21,7 +22,7 @@ export abstract class IPreferenceDefinition extends IBaseAlkemio {
   })
   description!: string;
 
-  @Field(() => UserPreferenceType, {
+  @Field(() => PreferenceType, {
     description: 'The type of the Preference, specific to the Entity it is on.',
   })
   type!: UserPreferenceType | HubPreferenceType;
```
valentinyanakiev:
```suggestion
  type!: PreferenceType;
```

valentinyanakiev (changes requested):
Please review my suggested changes, I think we can avoid copy/pasting and code duplication without breaking functionality. I have created a server-1792-b branch so you can check out the changes.

src/common/enums/preference.type.ts
techsmyth: Feedback is incorporated in base branch

src/domain/common/preference/preference.definition.interface.ts
techsmyth: Feedback is incorporated in base branch

valentinyanakiev (approved):

PR #1802, opened 2022-03-11
renamed createApplication to be applyForCommunityMembership ++

techsmyth:
applying for membership now checks the COMMUNITY_APPLY privilege
added joinCommunity mutation that checks the COMMUNITY_JOIN privilege
updated graphql samples


src/domain/community/community/community.resolver.mutations.ts
```diff
@@ -189,6 +186,51 @@ export class CommunityResolverMutations {
...
   }
 
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => ICommunity, {
+    description: 'Join the specified Community as a member.',
+  })
+  @Profiling.api
+  async joinCommunity(
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args('joinCommunityData') joiningData: CommunityJoinInput
```
hero101, line 186: if only a user is allowed to join by themselfs, why is an userID required in the input?

src/domain/community/community/community.resolver.mutations.ts
```diff
@@ -189,6 +186,51 @@ export class CommunityResolverMutations {
...
+
+    // Todo: notification for new community membership
+    // const payload =
+    //   await this.notificationsPayloadBuilder.buildApplicationCreatedNotificationPayload(
+    //     agentInfo.userID,
+    //     joiningData.userID,
+    //     community
+    //   );
+
+    // this.notificationsClient.emit<number>(
```
hero101, line 207: This will disable the notification for that event. Isn't it better to leave in the current one and address it at a later stage?

src/domain/community/community/community.resolver.mutations.ts
```diff
@@ -189,6 +186,51 @@ export class CommunityResolverMutations {
     return savedApplication;
   }
 
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => ICommunity, {
+    description: 'Join the specified Community as a member.',
```
hero101: Maybe add something about that an application won't be created to just be explicit.

hero101 (changes requested):

src/domain/community/community/community.resolver.mutations.ts
techsmyth: Description is adjusted as per feedback

src/domain/community/community/community.resolver.mutations.ts
techsmyth: This needs to be a totally different notification, and probably be "new community member". The application notification would not be appropriate.

src/domain/community/community/community.resolver.mutations.ts
techsmyth: Good point. I have adjusted both mutations to remove the userID from the input type, and the mutation now picks that up from the agentInfo

hero101: What is the advantage of using an input object over an inline argument just for the `communityID`?

hero101 (changes requested):

techsmyth:
Re why an input DTO...we chose a long time back to basically aim for a DTO for each mutation input. Makes it easier to extend in the future, and also gives more flexibility in terms of paramters.

hero101 (approved):

PR #1805, opened 2022-03-14
Refactoring for the end-end usage of VCs, including being able to have a list of trusted issuers

techsmyth:


package.json
```diff
@@ -71,6 +71,7 @@
     "image-size": "^1.0.0",
```
valentinyanakiev, line 71: Bump the version of the server in package.json + npm install for package-lock.json

src/common/exceptions/ssi.vc.not.verifiable.ts
```diff
@@ -0,0 +1,8 @@
+import { LogContext, AlkemioErrorStatus } from '@common/enums';
+import { BaseException } from './base.exception';
+
+export class SsiVcNotVerifiable extends BaseException {
+  constructor(error: string, context: LogContext) {
+    super(error, context, AlkemioErrorStatus.SSI_ERROR);
```
valentinyanakiev: Make the error status the same granularity as the exception class. If you want grouping use LogContext.

src/services/platform/trust-registry/trust.registry.claim/claim.interface.ts
```diff
@@ -0,0 +1,3 @@
+export interface IClaim {
```
valentinyanakiev, line 1: I am not sure this is sufficient in the broad Identity space. Claims are three types:
- registered - These are a set of predefined claims which are not mandatory but recommended, to provide a set of useful, interoperable claims.
- public - These can be defined at will by those using JWTs. But to avoid collisions they should be defined in the IANA JSON Web Token Registry or be defined as a URI that contains a collision resistant namespace.
- private - These are the custom claims created to share information between parties that agree on using them and are neither registered or public claims.

valentinyanakiev (changes requested):
I think there are some minor changes required:
- refactor SsiVcNotVerifiable to provide correct feedback
- bump the version

Re claims interface - that we can take as a separate story as it has wider implications

valentinyanakiev:
I am not sure whether the issue is in the server or wallet manager, but the feedback, given when bootstrapping DIDs is different in wallet manager and alkemio server:

Wallet manager:
image

Server:
image

src/services/platform/trust-registry/trust.registry.claim/claim.interface.ts
techsmyth: agree this space needs attention, raised as a separate epic: https://github.com/alkem-io/alkemio/issues/547

src/common/exceptions/ssi.vc.not.verifiable.ts
techsmyth: addressed

package.json
techsmyth: addressed

valentinyanakiev (dismissed):

valentinyanakiev (approved):

PR #1810, opened 2022-03-15
Authorization framework with claims

techsmyth:
Expanded authorization framework to leverage VC claims.
Made all usage of authorization policy rules to work through interfaces
Made a proper interface (graphql api) class for rules for VC claims
Fixed issue with resetting VC Claim rules in auth policies
Removed old grantStateTransitionVC mutation and associated code
Added flag to enable / disable VC verification of trusted issuer; defaults to false.


src/domain/agent/verified-credential/dto/verified.credential.dto.result.ts
```diff
@@ -1,48 +1,30 @@
...
+import { IVerifiedCredential } from '../verified.credential.interface';
+import { VerifiedCredentialClaim } from './verified.credential.dto.claim.result';
+
+export class VerifiedCredential implements IVerifiedCredential {
   type: string;
 
-  @Field(() => String, {
-    description: 'The challenge issuing the VC',
-  })
   issuer: string;
```
valentinyanakiev, line 7: Issuer and Expiration are two reserved claims. I am not sure whether there is a standard in VC claims, but in JWT claims issuer is always `iss`, and expiration is always `exp`. Please have a look at standard claims notation: https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-token-claims

valentinyanakiev (approved):
Raised a separate issue to deal with claims - otherwise code works for me, I tested it locally + jolocom wallet and I managed to get the credentials in / out of the wallet.

PR #1846, opened 2022-03-31
Migration fix

valentinyanakiev:


src/migrations/1648469585965-challengeOrgPrefs.ts
```diff
@@ -36,11 +36,11 @@ export class challengeOrgPrefs1648469585965 implements MigrationInterface {
       const prefAuthId = randomUUID();
       await queryRunner.query(`
                 INSERT INTO authorization_policy (id, createdDate, updatedDate, version, credentialRules, verifiedCredentialRules, anonymousReadAccess, privilegeRules)
-                VALUES ('${prefAuthId}', NOW(), NOW(), 1, '', '', 'false', '')
+                VALUES ('${prefAuthId}', NOW(), NOW(), 1, '', '', 0, '')
```
hero101, line 39: I thought we decided on `boolean` instead?

hero101 (changes requested):

src/migrations/1648469585965-challengeOrgPrefs.ts
valentinyanakiev: Boolean in the database and boolean in the code are two different things. At the moment, this field is TINYINT(4), whereas true/false values are in columns VARCHAR(16). Our data schema is not consistent at all and that led to the problem and failing migration.

hero101 (dismissed):

src/migrations/1648469585965-challengeOrgPrefs.ts
valentinyanakiev: We have multiple problems in the data schema as well - we have primary - foreign key relationships with different data types, e.g. CHAR(16) vs VARCHAR(16).

Comoque1 (approved):

hero101 (approved):

PR #1849, opened 2022-03-31
Organizations "match domain" preference implemented

hero101:
will assign newly signed users automatically to it

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/domain/community/user/user.resolver.mutations.ts
```diff
@@ -214,4 +229,47 @@ export class UserResolverMutations {
     );
     return await this.userAuthorizationService.applyAuthorizationPolicy(user);
   }
+
+  private async assignUserToOrganizationByDomain(
+    agentInfo: AgentInfo,
+    user: IUser
+  ): Promise<void> {
+    if (!agentInfo.emailVerified) {
+      return;
```
valentinyanakiev: I think this is an exceptional case, the user should _not_ be able to reach that point - silently returning is not correct in that case. An exception should be thrown.

src/domain/community/user/user.resolver.mutations.ts
```diff
@@ -214,4 +229,47 @@ export class UserResolverMutations {
...
+    }
+
+    const userEmailDomain = user.email.split('@')[1];
```
valentinyanakiev: it might be a bit of an overkill for this instance, but in my opinion email functions should be under /src/common/utils in something like email.util.ts. There is plenty of email code scattered around the application that can be reused - regexes, splitting, validation etc.

src/domain/community/user/user.resolver.mutations.ts
```diff
@@ -214,4 +229,47 @@ export class UserResolverMutations {
...
+      { relations: ['preferenceSet', 'verification'] }
+    );
+
+    if (
+      !org ||
+      org.verification.status !==
+        OrganizationVerificationEnum.VERIFIED_MANUAL_ATTESTATION ||
+      !org.preferenceSet
+    ) {
+      return;
```
valentinyanakiev: add verbose log here

src/domain/community/user/user.resolver.mutations.ts
```diff
@@ -214,4 +229,47 @@ export class UserResolverMutations {
...
+    }
+
+    const orgMatchDomain =
+      this.preferenceSetService.getPreferenceOrFail(
+        org.preferenceSet,
+        OrganizationPreferenceType.AUTHORIZATION_ORGANIZATION_MATCH_DOMAIN
+      ).value === 'true';
+
+    if (!orgMatchDomain) {
+      return;
```
valentinyanakiev: add verbose log here

src/domain/community/user/user.resolver.mutations.ts
```diff
@@ -214,4 +229,47 @@ export class UserResolverMutations {
     );
     return await this.userAuthorizationService.applyAuthorizationPolicy(user);
   }
+
+  private async assignUserToOrganizationByDomain(
+    agentInfo: AgentInfo,
+    user: IUser
+  ): Promise<void> {
```
valentinyanakiev: it will be better to have the Promise<boolean> as return type, as it is good to know whether the user was successfully assigned or not

valentinyanakiev (changes requested):

valentinyanakiev (approved):
Works for me, tested with or without AUTHORIZATION_ORGANIZATION_MATCH_DOMAIN and with or without VERIFIED_MANUAL_ATTESTATION on the org. Seems to work properly.

One note: when refactoring the code most of the logic from the resolver should be in service.

PR #1856, opened 2022-04-04
new registration service to handle logic for registration

techsmyth:
related to new users being assigned to org based on preferences + status


valentinyanakiev (approved):
- Verified user registration locally, worked for me
- Code is OK

PR #1862, opened 2022-04-07
Aspect defaultDescription & typeDescription

valentinyanakiev:


techsmyth (changes requested):
Overall functionality looks good. However some confusion related to the renaming:
- on the Aspect entity the field should be called "description"
- on the AspectTemplate entity the field should be called "defaultDescription"

And then when the aspect is being created the client should pick up the defaultDescription field from the AspectTemplate to fill in the initial value of the description box.

graphql-samples/mutations/create/create-aspect-on-context
```diff
@@ -1,7 +1,8 @@
 mutation CreateAspect($aspectData: CreateAspectInput!) {
   createAspectOnContext(aspectData: $aspectData) {
     displayName,
-    description,
+    defaultDescription,
```
techsmyth: this should be description, defaultDescription is only in the template

graphql-samples/mutations/create/create-aspect-on-context
```diff
@@ -14,7 +15,8 @@ query variables:
     "contextID": "uuid",
     "type": "test",
     "displayName": "TechStack2",
-    "description": "How do I handle scalability?",
+    "defaultDescription": "How do I handle scalability?",
```
techsmyth: ditto

src/domain/context/aspect/dto/aspect.dto.update.ts
```diff
@@ -10,6 +10,10 @@ export class UpdateAspectInput extends UpdateNameableInput {
   @MaxLength(LONG_TEXT_LENGTH)
   description?: string;
 
+  @Field({ nullable: true })
```
techsmyth, line 13: No need to update the type description; also impacts the updateAspect method

src/domain/context/aspect/aspect.service.ts
```diff
@@ -120,6 +120,10 @@ export class AspectService {
       aspect.description = aspectData.description;
     }
 
+    if (aspectData.typeDescription) {
```
techsmyth: Not sure why we would need to update the typeDescription, it is fixed - and also questionable if it should be stored per entity - but lets not pick this up now.

src/domain/challenge/hub/dto/hub.dto.template.aspect.ts
```diff
@@ -6,10 +6,18 @@ export class AspectTemplate {
   type: string;
 
   @Field(() => String)
-  description: string;
+  defaultDescription: string;
 
-  constructor(type: string, description?: string) {
+  @Field(() => String)
+  typeDescription: string;
```
techsmyth, line 18: Good to add in the description metadata for the graphql api

src/domain/challenge/hub/dto/hub.dto.template.aspect.ts
```diff
@@ -6,10 +6,18 @@ export class AspectTemplate {
   type: string;
 
   @Field(() => String)
-  description: string;
+  defaultDescription: string;
```
techsmyth, line 13: Good to add in the description metadata for the graphql api

techsmyth (changes requested):
Mostly there, still some minor updates.

src/domain/context/aspect/dto/aspect.dto.update.ts
valentinyanakiev: So if somebody adds a type description and e.g. copy / pastes some description incorrectly, the option to correct it won't exist? Is that really what we want?

src/domain/context/aspect/aspect.service.ts
valentinyanakiev: I agree that it's not ideal to be per entity, but that will open another set of updates, similarly to preference / preferenceDefinition.

techsmyth (approved):

PR #1867, opened 2022-04-12
added validation to aspect template updates

techsmyth:


Comoque1:
Would it be possible to apply validation on `type` field not to be empty string?
image

Comoque1 (approved):

PR #1900, opened 2022-05-01
Location entity

techsmyth:
Added new Location entity with the following fields: city, country
Add location entity to Profile; 
Add location entity to Context.

As part of the initial work for Location there was also a start made on some Community area clean up. This is also in the branch. Not ideal but should not disrupt and will be needed for the pending Community membership update work:
- Cleaned up membership service DTO names + usage; 
- Added Contributor base type that is shared with User + Organization; also updated some shared DTOs.

To do:
- update test suites


valentinyanakiev:
The schema changes need to be discussed:
- Contributor entity + interface logically makes sense but the design is a bit vague, it seems like a contributor is an entity with profile and agent, and that is super generic
- Logically, contributions should be discussed as we brainstormed contributions in the team and a lot can be achieved via credentials + services. Logically, it makes sense to have it distinctively in the domain model but it is a design decision we should be certain we'd like to do
- location entity relationships are wrong in my opinion. A location is a location. There is only one Amsterdam, Netherlands, which automatically means that the relationships are not one-to-one (Profile -> Location, Context -> Location). That will require changes in the migrations as well

valentinyanakiev:
The schema changes need to be discussed:
- Contributor entity + interface logically makes sense but the design is a bit vague, it seems like a contributor is an entity with profile and agent, and that is super generic
- Logically, contributions should be discussed as we brainstormed contributions in the team and a lot can be achieved via credentials + services. Logically, it makes sense to have it distinctively in the domain model but it is a design decision we should be certain we'd like to do
- location entity relationships are wrong in my opinion. A location is a location. There is only one Amsterdam, Netherlands, which automatically means that the relationships are not one-to-one (Profile -> Location, Context -> Location). That will require changes in the migrations as well

valentinyanakiev:
I have pushed through some changes, also same branch on the client. Location seems to be working on the user profile with the location entity.

techsmyth:
@valentinyanakiev thanks for the review + comments.

Re contributor entity, indeed this would be good to chat through more. 

re location, while I agree that there is only one of each "location" logically, I would not suggest that here due to the links it would generate internally in the code base via the ORM mapping. I also do see this PR as a first step of multiple towards full location support in Alkemio, basically for now bringing in the concept and isolating it in a specific entity in the domain model. (city + country is by far not a rich enough model). It will most definitely need to be expanded - and topics like not duplicating etc should be discussed in a later step.

src/domain/context/context/context.resolver.fields.ts
```diff
@@ -73,6 +73,16 @@ export class ContextResolverFields {
     return await this.contextService.getAspects(context, ids, limit, shuffle);
   }
 
+  // @UseGuards(GraphqlGuard)
```
me-andre: I'd suggest that when we comment out a bunch of code, we also add a comment, indicating the reason for that.

valentinyanakiev: Nice catch, it's working code, I uncommented it.

valentinyanakiev (approved):

PR #1934, opened 2022-05-13
Comunity policy role limits validation

alstojanovic:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/domain/community/community/community.service.ts
```diff
@@ -416,12 +418,60 @@ export class CommunityService {
     return community;
   }
 
+  private async validateCommunityPolicyLimits(
+    community: ICommunity,
+    communityPolicyRole: CommunityPolicyRole,
+    role: CommunityRole,
+    action: CommunityContributorsUpdateType
+  ) {
+    const orgMembers = await this.getOrganizationsWithRole(community, role);
```
techsmyth: This is going to be very heavy in two dimensions: 
- you only need the count, not the full organization loaded
- this implementation always checks both users and organizations, when there will never be more than one being checked at a time.

src/domain/community/community/community.service.ts
```diff
@@ -422,54 +458,70 @@ export class CommunityService {
     community: ICommunity,
```
valentinyanakiev, line 462: This is one method that is too long and can easily be split into multiple units. As a rule of thumb, try having the body of your methods <20 LOC (and I know the codebase is not like that everywhere). Here you can split e.g.

```typescript 
if (contributorType == CommunityContributorType.USER) validateUserCommunityPolicy(...)
if (contributorType == CommunityContributorType.ORGANIZATION) validateOrganizationCommunityPolicy(...)
...

It can be further split on ADD REMOVE, but that is not necessary in my opinion. As a rule of thumb, if a method can't be caputred in a single screen, it's not a good sign, and I have a 25" high resolution screen and couldn't see the whole method.

File: src/domain/community/community/community.service.ts [hunk 3]
```diff
@@ -257,6 +258,20 @@ export class CommunityService {
     });
   }
 
+  async countUsersWithRole(
```
valentinyanakiev: You have ```CommunityContributorType``` enum, why not make a countContributorPerRole method and have the enum as input. The only difference between countUsersWithRole and countOrganizationsWithRole is one call, that can be determined by the enum - all the rest is the same (signature, return type, the membership credential call etc.).

src/common/enums/community.contributors.update.type.ts
```diff
@@ -0,0 +1,4 @@
+export enum CommunityContributorsUpdateType {
+  ADD = 'add',
```
valentinyanakiev: Here the action is `add`, in the community service method is `assign`. Better use `assign` for consistency.

valentinyanakiev (changes requested):
Please see the comments inline.
Also something unrelated to your implementation but having a huge impact on it is that the validation won't catch assigning users to role, because we have:
- assignOrganizationToRole --> assignContributorToRole --> validateCommunityPolicyLimits
- removeContributorFromRole --> validateCommunityPolicyLimits --> removeUserFromRole 
- removeOrganizationFromRole --> validateCommunityPolicyLimits --> removeOrganizationFromRole

And assignUserToRole doesn't call assignContributorToRole ...

valentinyanakiev (dismissed):

valentinyanakiev (approved):

PR #1962, opened 2022-05-31
Refactored roles service + tests

valentinyanakiev:


src/domain/community/user/user.service.ts
```diff
@@ -435,36 +435,7 @@ export class UserService {
...
-      }
-
-      await this.setUserCache(user);
-    }
-
-    return user;
-  }
-
-  validateEmail(email: string): boolean {
+  private validateEmail(email: string): boolean {
```
hero101: this could be util

src/services/domain/roles/roles.service.ts
```diff
@@ -89,96 +82,193 @@ export class RolesService {
...
+      hubsMap
+    );
+
+    for (const hubResult of hubsMap.values()) {
+      membership.hubs.push(hubResult);
+    }
+    for (const orgResult of orgsMap.values()) {
+      membership.organizations.push(orgResult);
+    }
+
```
hero101, line 95:
```ts
membership.hubs.push(...hubsMap.values());
membership.organizations.push(...orgsMap.values())
```

src/services/domain/roles/roles.service.ts
```diff
@@ -89,96 +82,193 @@ export class RolesService {
...
+    credential: ICredential,
+    hubsMap: Map<string, RolesResultHub>,
+    contributorID: string,
+    orgsMap: Map<string, RolesResultOrganization>
+  ) {
+    const group = await this.userGroupService.getUserGroupOrFail(
+      credential.resourceID
+    );
+    const parent = await this.userGroupService.getParent(group);
+    if ('hubID' in parent) {
```
hero101: use `isICommunity` to infer the correct type in the if scope.

src/services/domain/roles/roles.service.ts
```diff
@@ -89,96 +82,193 @@ export class RolesService {
...
+      const hubResult = await this.ensureHubRolesResult(
+        hubsMap,
+        (parent as ICommunity).hubID,
+        contributorID
+      );
+      const groupResult = new RolesResult(group.name, group.id, group.name);
+      hubResult.userGroups.push(groupResult);
+    } else {
```
hero101: Since we don't know what implements `IGroupable` you must have an explicit check `isIOrganization` and also infer the type.

src/services/domain/roles/roles.service.ts
```diff
@@ -303,65 +395,64 @@ export class RolesService {
...
+      community.hubID,
+      application.createdDate,
+      application.updatedDate
+    );
+
+    const isHubCommunity = await this.communityService.isHubCommunity(
+      community
+    );
+
+    if (!isHubCommunity) {
```
hero101: I would suggest using if-guard here and in the nested ifs to improve readability.

src/services/domain/roles/roles.service.ts
```diff
@@ -89,96 +82,193 @@ export class RolesService {
...
       }
     }
+  }
 
-    for (const hubResult of hubsMap.values()) {
-      membership.hubs.push(hubResult);
-    }
-    for (const orgResult of orgsMap.values()) {
-      membership.organizations.push(orgResult);
+  private async addUserGroupMemberRole(
```
hero101, line 148: arg order is diverging from the other `add[xxx]` function signatures.

src/services/domain/roles/roles.service.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+        MockApplicationService,
+        MockCommunityService,
+        MockOpportunityService,
+        MockOrganizationService,
+        MockWinstonProvider,
+        RolesService,
+      ],
+    }).compile();
+
+    rolesService = moduleRef.get<RolesService>(RolesService);
```
hero101: you can skip the type here, it is inferred from the argument.

src/services/domain/roles/roles.service.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      expect(res.hubs).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({
+            hubID: testData.hub.id,
+          }),
+        ])
+      );
+    });
+
+    it('Should throw exception', async () => {
```
hero101: be more descriptive of why it should throw

src/services/domain/roles/roles.service.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+          expect.objectContaining({
+            organizationID: testData.organization.id,
+          }),
+        ])
+      );
+
+      expect(res.hubs).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({
+            hubID: testData.hub.id,
```
hero101, line 122: I think it is also beneficial to test what are the actual roles returned.

src/services/domain/roles/roles.service.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+      jest
+        .spyOn(opportunityService, 'getOpportunityOrFail')
+        .mockResolvedValue(testData.opportunity as any);
+
+      const res = await rolesService.getOrganizationRoles({
+        organizationID: testData.organization.id,
+      });
+
+      expect(res.hubs).toEqual(
```
hero101, line 243: I am confused. The test should be testing organization roles, but you are asserting hubs.

hero101 (changes requested):

src/services/domain/roles/roles.service.spec.ts
valentinyanakiev: image

hero101 (approved):

PR #1992, opened 2022-06-09
Comment subscription remade to be aspect comment subscription

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/domain/communication/comments/comments.resolver.mutations.ts
+++ b/src/domain/communication/comments/comments.resolver.mutations.ts
@@ -1,6 +1,7 @@
 import { Inject, UseGuards } from '@nestjs/common';
 import { Resolver } from '@nestjs/graphql';
 import { Args, Mutation } from '@nestjs/graphql';
+import { ClientProxy } from '@nestjs/microservices';
 import { getConnection } from 'typeorm';
 import { CurrentUser, Profiling } from '@src/common/decorators';
 import { GraphqlGuard } from '@core/authorization';
@@ -18,11 +19,11 @@ import {
   NOTIFICATIONS_SERVICE,
   SUBSCRIPTION_ASPECT_COMMENT,
 } from '@common/constants/providers';
-import { CommentsMessageReceived } from './dto/comments.dto.event.message.received';
+import { AspectCommentsMessageReceived } from '../../context/aspect/dto/aspect.dto.event.message.received';
 import { CommentsAuthorizationService } from './comments.service.authorization';
 import { EventType } from '@common/enums/event.type';
 import { NotificationsPayloadBuilder } from '@core/microservices';
-import { ClientProxy } from '@nestjs/microservices';
+import { IComments } from '@domain/communication/comments/comments.interface';
 
 @Resolver()
 export class CommentsResolverMutations {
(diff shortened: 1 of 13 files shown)


src/domain/communication/comments/comments.resolver.mutations.ts
```diff
@@ -129,4 +108,37 @@ export class CommentsResolverMutations {
...
+    aspect: { id: string; displayName: string; createdBy: string },
+    comments: IComments,
+    commentSent: CommunicationMessageResult
+  ) {
+    // build subscription payload
+    const eventID = `comment-msg-${Math.floor(Math.random() * 100)}`;
+    const subscriptionPayload: AspectCommentsMessageReceived = {
+      eventID: eventID,
+      message: commentSent,
+      commentsID: comments.id,
```
me-andre: Isn't it a kind of overkill to send both `commentsID` and `aspectID`? If we decide that we no longer subscribe to Comments (we subscribe to an Aspect instead), then I assume references to Comments should go away.

hero101: [We are not subscribing to comments, but to aspect.](https://github.com/alkem-io/server/pull/1992/files#diff-6ef7b37e6f5879aac60b7d6bac841dabee77d55b6364dd49bb762be247d37dc0R73) This is the event payload - it is used to filter out the messages.

src/domain/communication/comments/comments.resolver.mutations.ts
```diff
@@ -129,4 +108,37 @@ export class CommentsResolverMutations {
...
     );
   }
+
+  private async processAspectCommentEvents(
+    aspect: { id: string; displayName: string; createdBy: string },
+    comments: IComments,
+    commentSent: CommunicationMessageResult
+  ) {
+    // build subscription payload
+    const eventID = `comment-msg-${Math.floor(Math.random() * 100)}`;
```
me-andre: Does it make sense to switch to `uuid` for generating those?

hero101: I don't really have a preference. It would be easier but it is already used this way across the subscriptions and it's mostly used for debugging purposes.

me-andre (changes requested):
The code looks good.

However, I still don't get why switching from subscribing to `Comments` to subscribing to a parent entity is an improvement.

hero101:
> The code looks good.
> 
> However, I still don't get why switching from subscribing to `Comments` to subscribing to a parent entity is an improvement.

@me-andre It makes more sense with our current approach - handling events with a single subscription by entity by operation, e.g message created/updated/deleted on an aspect will produce 3 subscriptions/consumers BUT on a single queue - which is contradicting the whole point of load balancing, because those 3 consumers have to receive the message, but only 1 does.
Subscribing on Comments and having all the events in the same queue is contradicting to what we currently need.

src/domain/communication/comments/comments.resolver.mutations.ts
me-andre: I see. Let's extract to an util then?

src/domain/communication/comments/comments.resolver.mutations.ts
me-andre: I see. The question is - why do we need to send the `commentsID` then?

src/domain/communication/comments/comments.resolver.mutations.ts
hero101: Comments addressed.

me-andre (approved):

PR #1997, opened 2022-06-16
Redis

valentinyanakiev:


hero101 (dismissed):

src/domain/communication/communication/communication.service.spec.ts
```diff
@@ -1,14 +1,29 @@
...
 describe('CommunicationService', () => {
   let service: CommunicationService;
 
   beforeEach(async () => {
     const module: TestingModule = await Test.createTestingModule({
-      imports: [AppModule],
-    }).compile();
+      providers: [
+        CommunicationService,
+        {
```
hero101: extract this provider and replace it inline like the others

hero101 (changes requested):
Suggestions aside, I am not sure you need to use `useMocker` if you are providing the dependencies mocked in each test.
Also, take a look at [this](https://jestjs.io/docs/configuration#automock-boolean).

src/domain/communication/communication/communication.service.spec.ts
valentinyanakiev: Fixed with a factory in https://github.com/alkem-io/server/pull/1997/commits/5433ab6f5485f5d2000b8369104266935623c945

hero101 (approved):

PR #2002, opened 2022-06-20
avoid situation whereby canvas cannot be checked in / out 

techsmyth:
Removed cached "status" of the checkout entity, which is where the problems were arising. Now derive the status from the state of the Lifecycle.

Added migration. 

NOTE: existing canvases that have this problem should also be fixed with this update.


valentinyanakiev (changes requested):
Clearly doesn't work for me. A simple 2 min smoke test failed - I created a canvas, added an object, and clicked SAVE AND STOP EDITING. Now when I click EDIT, I don't get an error, and I can't edit the canvas. SAVE / SAVE AND STOP EDITING are grayed out.

For me, this behavior is even worse than the previous one, because there is no indication of an error, and the canvas is not functional.

In my opinion, we should close this and go for a proper fix.

valentinyanakiev (changes requested):
Still doesn't work for me. Checking in and out a couple of times broke it. 
Ran the migration on a broken state and it wasn't fixed as well.

image

image

techsmyth:
What do you mean broken? By definition is cannot get out of sync any more in terms of status + lifecycle states

valentinyanakiev (approved):
Works for me...

PR #2006, opened 2022-06-22
Cached AgentInfo

valentinyanakiev:
Added a separate story for addressing the redis lock issue:
https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/alkem-io/server/2021


src/domain/agent/agent/agent.cache.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    return await this.cacheManager.set<AgentInfo>(
+      await this.getAgentInfoCacheKey(email),
+      cachedAgentInfo,
+      {
+        ttl: 1,
+      }
+    );
+  }
+
+  private async getAgentInfoCacheKey(email: string): Promise<string> {
```
hero101: the return type is incorrect

src/domain/agent/agent/agent.cache.service.ts
```diff
@@ -0,0 +1,84 @@
...
+      await this.getAgentInfoCacheKey(email)
+    );
+  }
+
+  public async setAgentInfoCache(agentInfo: AgentInfo): Promise<AgentInfo> {
+    return await this.cacheManager.set<AgentInfo>(
+      await this.getAgentInfoCacheKey(agentInfo.email),
+      agentInfo,
+      {
+        ttl: 1,
```
hero101: Extract this one and the one in `update` to a constant or config.
isn't 1 second too small? I think boosting it a little will bring more performance benefits

src/domain/agent/agent/agent.cache.service.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+    );
+  }
+
+  private async getAgentInfoCacheKey(email: string): Promise<string> {
+    return `@agentInfo:email:${email}`;
+  }
+
+  public async getAgentEmail(agentId: string): Promise<string | undefined> {
+    const users: { email: string }[] = await getConnection().query(
```
hero101, line 131: any reason you did not use the agent repo here?

src/domain/community/user/user.service.ts
```diff
@@ -771,4 +772,63 @@ export class UserService {
...
+      .select([
+        'user.id as userID',
+        'user.communicationID as communicationID',
+        'agent.id as agentID',
+        'credentials.id as credentialID',
+        'credentials.resourceId as credentialResourceID',
+        'credentials.type as credentialType',
+      ])
+      .where('user.email = :email', { email: email })
+      .getRawMany<any>()
```
hero101: put the correct type here

src/domain/community/user/user.service.ts
```diff
@@ -771,4 +772,63 @@ export class UserService {
...
+      .then(agents => {
```
hero101: `agents` seems misleading here

src/domain/community/user/user.service.ts
```diff
@@ -771,4 +772,63 @@ export class UserService {
...
+        if (!agents) {
+          throw new EntityNotFoundException(
+            `Unable to load Agent for User: ${email}`,
```
hero101: agent OR credentials?

src/domain/community/user/user.service.ts
```diff
@@ -771,4 +772,63 @@ export class UserService {
...
+        'user.communicationID as communicationID',
+        'agent.id as agentID',
+        'credentials.id as credentialID',
+        'credentials.resourceId as credentialResourceID',
+        'credentials.type as credentialType',
+      ])
+      .where('user.email = :email', { email: email })
+      .getRawMany<any>()
+      .then(agents => {
+        if (!agents) {
```
hero101: The type suggests this can't be `undefined`, but an `empty` array.

hero101 (changes requested):

src/domain/agent/agent/agent.cache.service.ts
valentinyanakiev: You can't use the agent repo, as you need the email. I do that via right join, which is unsupported by ORM. Normally, that can be done via having the query + repo + service call in users service / module, but that would require importing the users module and results in a circular dependency user <--> agent.

src/core/authentication/authentication.service.ts
```diff
@@ -66,6 +66,12 @@ export class AuthenticationService {
         LogContext.AUTH
       );
     }
+
+    const cachedAgentInfo = await this.agentCacheService.getAgentInfoFromCache(
```
me-andre, line 70: If this approach is lazily initialized cache entry, i.e. no cache on read - creating one and returning, I'd wrap this whole thing in a lock (such as redis lock). Pseudo code:
```ruby
lock_resource :resource
saved = nil
return saved if saved = try_get()
saved = store(compute())
release_resource :resource
return saved
```

This will eliminate situations when there's no cache and multiple concurrent requests are accessing the same cache entry: they will all start producing the value until the first one succeeds and updates the cache.

src/domain/agent/agent/agent.cache.service.ts
```diff
@@ -0,0 +1,93 @@
...
+      this.logger.verbose?.(
+        `No user cache found for user ${email}. Skipping cache update!`,
+        LogContext.AGENT
+      );
+
+      return undefined;
+    }
+
+    cachedAgentInfo.credentials = agent.credentials as ICredential[];
+    return await this.cacheManager.set<AgentInfo>(
```
me-andre: Can we reuse `setAgentInfoCache()` here?

src/domain/agent/agent/agent.cache.service.ts
```diff
@@ -0,0 +1,93 @@
...
+  private getAgentInfoCacheKey(email: string): string {
+    return `@agentInfo:email:${email}`;
+  }
+
+  public async getAgentEmail(agentId: string): Promise<string | undefined> {
+    const users: { email: string }[] = await getConnection().query(
+      `SELECT \`u\`.\`email\` FROM \`agent\` as \`a\`
+      RIGHT JOIN \`user\` as \`u\`
+      ON \`u\`.\`agentId\` = \`a\`.\`id\`
+      WHERE \`a\`.\`id\` = '${agentId}'`
```
me-andre: If I assume correctly that `getConnection()` provides a raw driver interface, then we should look for a safer way of passing query params.
I didn't work with mysql in node, but in `pg` it looked like this:
```js
await query('SELECT * FROM tableName WHERE id = $1', [params.id])
```

As a bonus, the need to escape backticks may go away.

hero101: the second param of the `query` method is params.

```sql
getConnection().query('select id from user where id = ?', [123]);

src/core/authentication/authentication.service.ts
valentinyanakiev: Will be addressed in https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/alkem-io/server/2021

PR #2007, opened 2022-06-22
Session extend using the kratos api

hero101:

### Governance 

- [ ] Documentation is added


src/core/authentication/authentication.service.ts
```diff
@@ -1,20 +1,44 @@
...
+    ).authentication.providers.ory.kratos_public_base_url_server;
+    this.extendAfter = Number(
+      this.configService.get(ConfigurationTypes.IDENTITY).authentication
+        .providers.ory.earliest_possible_extend
+    );
+    // todo: read from env
+    this.kratosPrivateUrl = '';
+    this.adminPasswordIdentifier = '';
+    this.adminPassword = '';
+  }
```
hero101, line 52: @valentinyanakiev The config must be read from here.

src/core/authentication/ory.api.strategy.ts
```diff
@@ -40,12 +40,16 @@ export class OryApiStrategy extends PassportStrategy(
```
valentinyanakiev, line 40: I don't believe it's necessary to refresh sessions for non-interactive (API) flows.

src/core/authentication/get.bearer.token.ts
```diff
@@ -0,0 +1,27 @@
...
+  kratosPublicUrl: string,
+  password_identifier: string,
+  password: string
+): Promise<string> | never => {
+  const kratos = new PublicApi(
+    new Configuration({
+      basePath: kratosPublicUrl,
+    })
+  );
+
```
valentinyanakiev, line 18: I'd move the checks for getting the bearer token from extendSession to here.
If a promise fails, I'd capture the error, and rethrow with a custom exception.

src/core/authentication/authentication.service.ts
```diff
@@ -80,4 +104,79 @@ export class AuthenticationService {
...
+
+    try {
+      bearerToken = await getBearerToken(
+        this.kratosPublicUrl,
+        this.adminPasswordIdentifier,
+        this.adminPassword
+      );
+    } catch (e) {
+      const message = (e as Error)?.message;
+      this.logger?.error(
```
valentinyanakiev: I don't believe we should have this.logger.error anywhere here, we should be capturing an error, wrapping it in custom exception, and then rethrowing it. It's an exceptional flow, we should not reach that point (e.g. our account should be able to retrieve a bearer token)

alkemio.yml
```diff
@@ -174,6 +174,10 @@ identity:
         # Ory Kratos URL for usage by the Alkemio server when inside a cluster.
         kratos_public_base_url_server: ${AUTH_ORY_KRATOS_PUBLIC_BASE_URL_SERVER}:http://localhost:3000/identity/ory/kratos/public
 
+        # session can't be extended until *earliest_possible_extend* **miliseconds** before it expires
+        # default is 24 hours
+        earliest_possible_extend: ${AUTH_EARLIEST_POSSIBLE_SESSION_EXTEND}:86400000
```
valentinyanakiev: I'd rather have this value in hours and then calculated, otherwise whoever manages the application needs to use a calculator here...

src/core/authentication/authentication.service.ts
```diff
@@ -1,20 +1,44 @@
...
+  ) {
+    this.kratosPublicUrl = this.configService.get(
+      ConfigurationTypes.IDENTITY
+    ).authentication.providers.ory.kratos_public_base_url_server;
+    this.extendAfter = Number(
+      this.configService.get(ConfigurationTypes.IDENTITY).authentication
+        .providers.ory.earliest_possible_extend
+    );
+    // todo: read from env
+    this.kratosPrivateUrl = '';
```
valentinyanakiev: These names are a bit confusing, as we have the following endpoints:
  - kratos admin api endpoint --> accessible only inside the cluster (in prod environments)
  - kratos public api endpoint with public URL --> accessible from outside of the cluster
  - kratos public api endpoint with cluster URL --> accessible from inside the cluster only (otherwise we'll be routing requests through traefik)
So my suggestion is:
  - `kratosPrivateUrl` --> `kratosAdminUrlServer` (meaning kratos admin URL from the server container / service)
  - `kratosPublicUrl` --> `kratosPublicUrlServer` (meaning kratos public URL from the server container / service)

Also, I see you are already using `kratos_public_base_url_server` --> at the moment all our configmaps  are misconfigured in all environments to use the public URL of the server there - if we don't fix that, the wrong URL will be taken and this fix won't work on demo / prod (where the admin endpoint is not exposed).

You will also need to add a config entry `kratos_admin_base_url_server` and take `kratosAdminUrlServer` from it.

valentinyanakiev (changes requested):

src/core/authentication/authentication.service.ts
```diff
@@ -1,20 +1,44 @@
...
+    this.adminPasswordIdentifier = '';
```
valentinyanakiev: Add the `adminPasswordIdentifier` and `adminPassword` in the config, below earliest_possible_extend:

```yml
        earliest_possible_extend: ${AUTH_EARLIEST_POSSIBLE_SESSION_EXTEND}:86400000

        admin_service_account:
          username: ${KRATOS_ADMIN_USERNAME}:kratos@alkem.io
          password: ${KRATOS_ADMIN_PASSWORD}:kr@t0$!
```

src/core/authentication/ory.strategy.ts
```diff
@@ -35,31 +36,33 @@ export class OryStrategy extends PassportStrategy(Strategy, 'oathkeeper-jwt') {
...
     let oryIdentity: OryDefaultIdentitySchema | undefined = undefined;
-    if (payload.session) {
-      oryIdentity = payload.session.identity as OryDefaultIdentitySchema;
+    const session = payload.session;
+
+    if (session) {
+      oryIdentity = session.identity as OryDefaultIdentitySchema;
+
+      if (this.authService.shouldExtendSession(session)) {
+        this.authService.extendSession(session);
```
valentinyanakiev:
```suggestion
        this.authService.extendSession(session).catch(error => {
          this.logger.error(
            `Ory Kratos session failed to be extended: ${error}`,
            LogContext.AUTH
          );
        });
```

That way you will not cancel the flow and will log the error with the correct log format.

src/core/authentication/ory.strategy.ts
```diff
@@ -35,31 +36,33 @@ export class OryStrategy extends PassportStrategy(Strategy, 'oathkeeper-jwt') {
...
-    // Todo: not sure this is correct, but am hitting a case whereby session is null
     let oryIdentity: OryDefaultIdentitySchema | undefined = undefined;
-    if (payload.session) {
-      oryIdentity = payload.session.identity as OryDefaultIdentitySchema;
+    const session = payload.session;
+
+    if (session) {
+      oryIdentity = session.identity as OryDefaultIdentitySchema;
+
+      if (this.authService.shouldExtendSession(session)) {
```
valentinyanakiev, line 55: shouldExtendSession is synchronous but its results are used only in extendSession (async), which invokes tryExtendSession.
I think it will be faster, non-blocking, if you have tryExtendSession call shouldExtendSession before anything else, and make shouldExtendSession async. That way you won't block the thread for shouldExtendSession.

valentinyanakiev (changes requested):

valentinyanakiev (approved):

PR #2010, opened 2022-06-23
reverting default for logging format

techsmyth:
Trivial but avoids having to change it for all devs.


valentinyanakiev:
> Trivial but avoids having to change it for all devs.

This pull request goes hand-in-hand with changing our env variables in all environments, otherwise, logstash transformations won't work. I was about to create a story to address our config file systematically.

valentinyanakiev (approved):

PR #2014, opened 2022-06-25
Canvas templates via templatesSet

techsmyth:
Paired with related branch on the client called canvas-templates.

This PR:
- removes notion of template from Canvas entity
- for any canvases that were templates, renames the canvas displayName to begin with "template_"
- adds in a sample mutation content for creating a canvas template
- removes canvas from ecosystemModel: this relationship was not used and was legacy
- allow creating a canvasTemplate using an existing canvas 

Note: the migration approach, just renaming, means that there will be manual clean up after this migration. The creation of canvasTemplates from the renamed Canvases can use the createCanvasTemplate mutation by specifying the canvas that starts with template_


src/migrations/1656674597777-canvasTemplates.ts
```diff
@@ -0,0 +1,131 @@
...
+      `SELECT id from canvas_template`
+    );
+    for (const canvasTemplate of canvasTemplates) {
+      await queryRunner.query(
+        `DELETE FROM canvas_template WHERE id='${canvasTemplate.id}'`
+      );
+    }
+
+    await queryRunner.query(
+      `ALTER TABLE \`canvas\` ADD \`isTemplate\` tinyint NOT NUL`
```
valentinyanakiev: There is a typo in the down migration. I am sure it has never been tested. Also, it migrates schema only, and not the data.

```suggestion
      `ALTER TABLE \`canvas\` ADD \`isTemplate\` tinyint NOT NULL`
```

valentinyanakiev (dismissed):
The migration generates rather weird data for canvas nameIDs:

`
query: UPDATE canvas SET nameID = '-7701' WHERE (id = 'ebd405ae-a031-484d-8a7f-695541182e8f')
query: UPDATE canvas SET nameID = '-6166' WHERE (id = 'ec2eeb8a-0c00-4bcf-bcde-e83cfa117eed')
query: UPDATE canvas SET nameID = '-5947' WHERE (id = 'ecac46e8-1446-405b-abda-0af3caae47ed')
query: UPDATE canvas SET nameID = '-6843' WHERE (id = 'ecce1fd1-ab75-4bd8-ae19-8bb21faaa1ce')
`

Also, when trying the migrated data in a real dataset --> create canvas in /digileefomgeving/canvases-->

image

This is not ready to be merged.

src/migrations/1656674597777-canvasTemplates.ts
techsmyth: Good spot re typo. 

Re data, we explicitly discussed this: down migrations do not migrate data, only structure.

techsmyth:
> g the migrated data in a real dataset --> create canvas in /digileefo

Need to find out what is happening with that data. 

Re migration name, we have a lot of canvases without a title - for example ecosystem-model canvases. We may want to remove that canvas usage as it is not used / adding value at the moment...

src/migrations/1656674597777-canvasTemplates.ts
techsmyth: Fixed

valentinyanakiev:
There are <20 templates that will need to be migrated on prod:
```sql
SELECT `displayName` FROM `alkemio`.`canvas` as `a` where `a`.`displayName` like '%template_%';
```

image

valentinyanakiev (approved):
`Up` and `down` migration worked for me. Selected 3 random templates and made templates from them - seemed to be working. Tested creating a `canvas` created from `canvas template` that was itself created from `canvas` from the web client - worked as well.

PR #2028, opened 2022-06-30
Convert challenge to Hub, opportunity to Challenge

techsmyth:
A new server: Conversion

It provides two mutations:
- to convert a challenge to a hub, recursively so also child opportunities get promoted
- to convert an opportunity to a challenge

For each conversion the following entities are maintained:
- contexts are fully re-used
- communications on community are re-used

In addition:
- users have old credentials removed + new ones assigned

Graphql samples are added.

This PR also improves some error messages related to community policy limits, and brings a bit of useful refactoring to the Hub service.


alstojanovic (approved):
All good. I've been wanting to update the policy role limit exception error message, thanks :)

PR #2084, opened 2022-08-16
Moved aspects / canvases counts to collaboration

techsmyth:
I saw that the activity was failing on opportunity so put in a fix. Found that since my merge another fix had gone in. 

However there was still the issue that the BaseChallenge service should never need to be importing the CanvasModule or AspectModule - these should only be known at the collaboration entity level.


alstojanovic (dismissed):

valentinyanakiev (changes requested):
I want to review it.

techsmyth:
Great if you can. We do need to take care that the module dependencies are also properly layered, so it is a red flag is the baseChallenge service has knowledge of canvasService etc.

valentinyanakiev (approved):

PR #2087, opened 2022-08-17
Lifecycle template creation 

valentinyanakiev:
- mandatory lifecycle template selection on challenge / opportunity creation
- check whether a lifecycle template of specific type, e.g. CHALLENGE / OPPORTUNITY etc is the last one of its type in the set - last template can NOT be deleted and results in an exception
   - logic is in the lifecycletemplate service. Logically, this belongs to the templatesSet service but there are no remove / delete operations there for any of lifecycleTemplate / aspectTemplate / canvasTemplate so I have kept the following logic
- defaulting to the first found template of a type in Opp --> Challenge conversion if none is provided (parameter for the template is set as `optional`)
- templateSet is mandatory for CreateChallenge / CreateOpportunity DTOs
- fixed a bug where authorization for lifecycleTemplates was messed up on auth reset


src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -113,7 +113,7 @@ export class ChallengeService {
 
     const machineConfig: ILifecycleDefinition =
       await this.lifecycleTemplateService.getLifecycleDefinitionFromTemplate(
-        challengeData.lifecycleID,
+        challengeData.lifecycleTemplateID,
```
techsmyth: Suggest to call it innovationFlowTemplateID

src/domain/challenge/challenge/dto/challenge.dto.create.ts
```diff
@@ -13,5 +13,5 @@ export class CreateChallengeInput extends CreateBaseChallengeInput {
   leadOrganizations?: string[];
 
   @Field(() => UUID, { nullable: false })
-  lifecycleID!: string;
+  lifecycleTemplateID!: string;
```
techsmyth: call it innovationFlowTemplateID - the type is Lifecycle...

src/domain/challenge/hub/hub.service.ts
```diff
@@ -507,6 +509,36 @@ export class HubService {
     );
   }
 
+  async getDefaultLifecycleTemplate(
```
techsmyth: Ditto re name...

src/domain/challenge/hub/hub.service.ts
```diff
@@ -507,6 +509,36 @@ export class HubService {
...
+      );
+
+    const allLifecycleTemplates =
+      await this.templatesSetService.getLifecycleTemplates(hub.templatesSet);
+
+    const selectableLifecycleTemplates = allLifecycleTemplates.filter(
+      x => x.type === lifecycleType
+    );
+
+    if (selectableLifecycleTemplates.length === 0)
```
techsmyth: Ditto...

techsmyth (changes requested):
The refactoring makes sense and looks clean.

I would suggest that we use the correct name in the updated DTOs and code i.e. InnovationFlow. "Lifecycle" is then the type. 

Also please then also update the updateChallengeLifecycle mutation to be called "updateChallengeInnovationFlow", renaming the associated DTO + containing field that specified the template ID. Ditto for Opportunity InnovationFlow template.

src/common/enums/authorization.privilege.ts
```diff
@@ -13,7 +13,7 @@ export enum AuthorizationPrivilege {
   CREATE_ORGANIZATION = 'create-organization',
   READ_USERS = 'read-users',
   UPDATE_CANVAS = 'update-canvas',
-  UPDATE_LIFECYCLE = 'update-lifecycle',
+  UPDATE_INNOVATION_FLOW = 'update-lifecycle',
```
alstojanovic, line 16: Should we also rename the 'update-lifecycle' literal to 'update-innovation-flow'?

alstojanovic (approved):

PR #2098, opened 2022-08-20
added defensive code to Lifecycle service to be able to handle content + context.parentID not being set

techsmyth:


Comoque1:
Tested OK on local environment with both Challenge and Opportunity.

valentinyanakiev (approved):

PR #2121, opened 2022-08-26
Callout edit field settings

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/domain/collaboration/callout/callout.service.ts
+++ b/src/domain/collaboration/callout/callout.service.ts
@@ -77,6 +77,9 @@ export class CalloutService {
     if (calloutUpdateData.visibility)
       callout.visibility = calloutUpdateData.visibility;
 
+    if (calloutUpdateData.displayName)
+      callout.displayName = calloutUpdateData.displayName;
+
     return await this.calloutRepository.save(callout);
   }
 
@@ -100,7 +103,10 @@ export class CalloutService {
     if (callout.authorization)
       await this.authorizationPolicyService.delete(callout.authorization);
 
-    return await this.calloutRepository.remove(callout as Callout);
+    const result = await this.calloutRepository.remove(callout as Callout);
+    result.id = calloutID;
+
+    return result;
   }
 
   private async setNameIdOnAspectData(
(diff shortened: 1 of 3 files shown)


src/domain/collaboration/callout/callout.service.ts
```diff
@@ -100,7 +103,10 @@ export class CalloutService {
     if (callout.authorization)
       await this.authorizationPolicyService.delete(callout.authorization);
 
-    return await this.calloutRepository.remove(callout as Callout);
+    const result = await this.calloutRepository.remove(callout as Callout);
+    result.id = calloutID;
```
alstojanovic, line 107: Doesn't remove return the id?

hero101: No

alstojanovic: I now checked, I can select it from the result and its type is `string`, not `string | undefined`, so it should be always there I would say.

alstojanovic (changes requested):
Remove changing the returned callout id to calloutID.

src/domain/collaboration/callout/callout.service.ts
techsmyth: Actually this is a pattern we have widely used, so I would keep with it. Key is that the returned data has the actual ID of the entity that was removed so that the client cache can remove it.

hero101: > I now checked, I can select it from the result and its type is string, not string | undefined, so it should be always there I would say.

**It's undefined.**

Before removal
image
After removal
image



>Current behaviour is by design. Missing id means entity doesn't exist in the db. And since we are mutating given entity, it's logically make sense. Depend of your use case, but you can simply do:
```ts
const removedEntityId = entity.id
await repository.remove(entity)
// now use removedEntityId where you need
```

[More on the topic](https://github.com/typeorm/typeorm/issues/7024)

alstojanovic: Wow, ok then it's good. Need to watch for other places we use it without reassigning.

alstojanovic (approved):

techsmyth (approved):

valentinyanakiev (approved):

PR #2124, opened 2022-08-29
UUID_NAMEID for callout field resolver & subscription

valentinyanakiev:
- Extended CalloutService to be able to get Callout by UUID_NAMEID instead of only UUID 
- Extended CalloutAspectCreated subscription to accept UUID_NAMEID
- Extended Callout FieldResolver in Collaboration receive UUID_NAMEID

* Changes should be backwards compatible

### DIFF
--- a/src/domain/collaboration/callout/callout.resolver.subscriptions.ts
+++ b/src/domain/collaboration/callout/callout.resolver.subscriptions.ts
@@ -12,7 +12,7 @@ import { AuthorizationPrivilege } from '@common/enums/authorization.privilege';
 import { SUBSCRIPTION_CALLOUT_ASPECT_CREATED } from '@common/constants/providers';
 import { CalloutService } from '@domain/collaboration/callout/callout.service';
 import { CalloutAspectCreated } from '@domain/collaboration/callout';
-import { UUID_NAMEID } from '@domain/common/scalars';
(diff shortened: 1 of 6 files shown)


src/domain/collaboration/callout/callout.resolver.subscriptions.ts
```diff
@@ -68,8 +68,8 @@ export class CalloutResolverSubscriptions {
     @CurrentUser() agentInfo: AgentInfo,
     @Args({
       name: 'calloutID',
-      type: () => UUID,
-      description: 'The ID of the Callout to subscribe to.',
+      type: () => UUID_NAMEID,
+      description: 'The ID or NameID of the Callout to subscribe to.',
```
techsmyth: The Callout NameID is only unique within the containing context, so this is not safe.

techsmyth (commented):
To discuss how the subscription is connected to the right call out

src/domain/collaboration/callout/callout.resolver.subscriptions.ts
valentinyanakiev: Addressed in b1b2a7972257c750d4bedc1796771ad439c216d2

techsmyth (approved):

PR #2132, opened 2022-09-01
#2385 Discussion Callout

ccanos:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/domain/collaboration/callout/callout.service.authorization.ts
```diff
@@ -61,6 +63,18 @@ export class CalloutAuthorizationService {
...
+
+    // Inherit for comments before extending so that the creating user does not
+    // have rights to delete comments
+    if (callout.comments) {
+      callout.comments =
+        await this.commentsAuthorizationService.applyAuthorizationPolicy(
+          callout.comments,
+          callout.authorization
+        );
+    }
```
alstojanovic: This is to prevent the creator of the Callout to be able to delete all comments on that Callout. It should stay here.

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -169,7 +175,7 @@ export class CalloutService {
 
     await this.setNameIdOnAspectData(aspectData, callout);
 
-    // Not idea: get the communicationGroupID to use for the comments
+    // Get the communicationGroupID to use for the aspect comments
```
techsmyth, line 179: Coment refers to aspect

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -364,4 +370,16 @@ export class CalloutService {
 
     return aspect;
   }
+
+  public async getComments(calloutId: string) {
```
techsmyth: Why is this done via direct repository access instead of going through ORM layer? We only tend to use repository access where performance is critical or the relationships are not navigatable. 

So I would prefer to see using getCalloutOrFail + then passing in a relation to ensure that the comments are loaded (am presuming they have eager loaded = false).

src/migrations/1661960517210-callouts-comments.ts
```diff
@@ -0,0 +1,35 @@
...
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    // await queryRunner.query(
+    //   `ALTER TABLE \`callout\` DROP CONSTRAINT \`FK_62ed316cda7b75735b20307b47e\``
+    // );
+
+    // await queryRunner.query(
+    //   `ALTER TABLE \`callout\` RENAME COLUMN commentsId TO discussionId;`
+    // );
+
+    await queryRunner.query(`UPDATE \`callout\` SET discussionId = NULL;`);
```
techsmyth: This is not going to revert the structure. Down should basically be a mirror of up.

techsmyth (changes requested):
Nice clear update. 

A couple of simple enough changes...

src/migrations/1661960517210-callouts-comments.ts
ccanos: I need to set all these values to NULL because next sentence for adding the Foreign key is going to fail if I don't do it.
They were all null before these changes anyway. 
The only problem I see here is that, if we redeploy this changes after a rollback, we may leave callouts with type='discussion' with discussionId=NULL

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -364,4 +371,17 @@ export class CalloutService {
 
     return aspect;
   }
+
+  public async getCommentssFromCallout(callout: ICallout): Promise<IComments> {
```
alstojanovic: Fix typo in method name, remove double s.

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -169,7 +176,7 @@ export class CalloutService {
 
     await this.setNameIdOnAspectData(aspectData, callout);
 
-    // Not idea: get the communicationGroupID to use for the comments
+    // Get the communicationGroupID to use for the callout comments
```
alstojanovic: This is for aspect comments, so it was good the first time. Change back to aspect. The same communicationGroupId is used for the callout comments but that is not too relevant here.

src/domain/collaboration/collaboration/collaboration.service.ts
```diff
@@ -135,6 +138,19 @@ export class CollaborationService {
...
+      const communicationGroupID =
+        await this.namingService.getCommunicationGroupIdForCallout(callout.id);
+
+      callout.comments = await this.commentsService.createComments(
+        communicationGroupID,
+        `callout-comments-${callout.displayName}`
+      );
+      await this.collaborationRepository.save(collaboration);
+    }
+
```
alstojanovic: Lines 139 and 140:
```js
    collaboration.callouts.push(callout);
    await this.collaborationRepository.save(collaboration);
```
should be after the check that the callout type is DISCUSSION. 
And remove the `await this.collaborationRepository.save(collaboration);` from line 151.

src/migrations/1661960517210-callouts-comments.ts
```diff
@@ -0,0 +1,35 @@
...
+  name = 'calloutsComments1661960517210';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE \`callout\` DROP CONSTRAINT \`FK_62ed316cda7b75735b20307b47e\``
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE \`callout\` RENAME COLUMN discussionId TO commentsId;`
+    );
```
alstojanovic: It's easier to rename but you want to DROP the `discussionId` column and add the `commentsId` column. Those will be two separate commands.

alstojanovic (dismissed):

src/domain/collaboration/collaboration/collaboration.service.ts
ccanos: I couldn't do this because I needed a callout.id already from the database

alstojanovic: You return the callout on line 138 so you have the callout.id.

alstojanovic: Fixed the underlying issue when creating a callout...

techsmyth (dismissed):
Nice to have this coming together! I made a few minor changes in last commit so please do scan that. 

Otherwise good from me to merge, please take it to develop when you can

techsmyth (approved):

PR #2134, opened 2022-09-02
Activity: activity adapter, activity entity and activity querying service

techsmyth:
This is an initial version to provide simple activity tracking per Collaboration. It is at the moment very much focused on activity around Collaborations, including membership of new users. It could be extended in other ways in the future. Ultimately the storage + retrieval of this data should be done using something like elastic - but that would be too far off. 

It adds:
- Platform level entity: Activity that is a reusable structure for being able to record + retrieve Activity. It could be seen as a Record.
- ActivityAdapter that takes care of managing the sending of the events
- ActivityLog service that takes care of querying the activity data

It is in essence a poor mans CQRS, but done in a way that would make it much easier to split this out into a separate service when we have more bandwidth.

Note: adding new platform level folder + entity is also as we expect to have additional platform level entities near term for the best practices store. 

It also adds a graphql-sample to show how to use it. 

Sample data:
image


hero101:
@techsmyth A created callout initially in draft and then publish and a created callout initially as published have different descriptions but the event is `CALLOUT_PUBLISHED`.
The first event is when the callout is created as published. The second one is when it's updated to published from draft.
image

src/domain/collaboration/callout/callout.resolver.mutations.ts
```diff
@@ -161,6 +167,13 @@ export class CalloutResolverMutations {
         EventType.CALLOUT_PUBLISHED,
         payload
       );
+
+      const activityLogInput: ActivityInputCalloutPublished = {
```
valentinyanakiev, line 171: From what I see in the code, there is one DTO type - `ActivityInputBase`. The rest are derived from it without anything different. It doesn't make any sense in the current context. I suggest `ActivityInputBase` be changed to `ActivityInput` and just get rid of all the derived types - it's literally the same type.

src/migrations/1662128944732-activity.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class activity1662128944732 implements MigrationInterface {
+  name = 'activity1662128944732';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE \`activity\` (\`id\` char(36) NOT NULL, \`createdDate\` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6), \`updatedDate\` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), \`version\` int NOT NULL, \`triggeredBy\` varchar(36) NULL, \`collaborationID\` varchar(36) NULL, \`resourceID\` varchar(36) NULL, \`description\` varchar(255) NOT NULL, \`type\` char(128) NOT NULL, PRIMARY KEY (\`id\`)) ENGINE=InnoDB`
```
valentinyanakiev: Why are you using `char(36)` and `varchar(36)` for the same field type? Searching `char` in the codebase doesn't yield any results.

src/platform/activity/dto/activity.dto.create.ts
```diff
@@ -0,0 +1,12 @@
+import { ActivityEventType } from '@common/enums/activity.event.type';
+
+export class CreateActivityInput {
+  triggeredBy!: string;
```
valentinyanakiev: This looks like `ActivityInputBase` (or alternatively `AcitvityInput`, should we decide to use it) + `collaborationID` & `type`. I'd reuse the type and nest it.

valentinyanakiev (changes requested):
Some minor changes that can simplify the codebase.

techsmyth:
(Relying to comment from Svetlo):

Yes, and I think we should probably split out the publish action from the update action, so that in the end publish comes from one specific action.

src/migrations/1662128944732-activity.ts
techsmyth: fixed

valentinyanakiev (approved):

PR #2138, opened 2022-09-05
Global roles security refinement

techsmyth:
Creation of a new global role: Global Hubs Admin
- this role can manage Hubs, assignments in Hubs, reset authorization policies
- this role **cannot** assign other global roles, nor global credentials

It comes with 2 additional mutations to add / remove the Global Hub Admins credential.

This is a high risk change, that while code wise may not be significant does impact some of the key platform flows - so it needs a thorough QA review.

New privileges:
- AUTHORIZATION_RESET: to control access to resetting of authorization
- GRANT_GLOBAL_ADMINS: to grant access to being able to assign / remove 
- PLATFORM_ADMIN: to identify users that should have access to the platform administration (allows removal of pseudo privilege determination on the client)

Updated platform level authorization policy to reflect the above changes.


src/domain/common/authorization-policy/authorization.policy.service.ts
```diff
@@ -268,6 +268,8 @@ export class AuthorizationPolicyService {
         globalRole === AuthorizationRoleGlobal.GLOBAL_COMMUNITY_ADMIN
```
valentinyanakiev: use `switch` and `case` instead of nested `if` - `else`

valentinyanakiev (changes requested):
Small code-style improvement comment inline. Other than that, light tested, reset auth policies twice and didn't find any very obvious issues.

valentinyanakiev (approved):

PR #2145, opened 2022-09-09
made collaboration interest notification work also for hub / challenges

techsmyth:
refactored generation of hubPayload + improved typing

### DIFF
--- a/src/core/microservices/notifications.payload.builder.ts
+++ b/src/core/microservices/notifications.payload.builder.ts
@@ -332,14 +332,15 @@ export class NotificationsPayloadBuilder {
   }
 
   private async buildHubPayload(community: ICommunity): Promise<HubPayload> {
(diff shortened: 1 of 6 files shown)


src/core/microservices/notifications.payload.builder.ts
```diff
@@ -361,16 +353,26 @@ export class NotificationsPayloadBuilder {
...
-          id: community.parentID,
+          id: opportunityID,
+          nameID: opportunityNameID,
         },
-      };
-    }
+      },
+    };
+    return result;
   }
```
hero101, line 368: I think variables here should not default to an empty string.
`HubPayload ` suggests  that if you have challenge or opportunity fields then a value is required.

hero101 (changes requested):

hero101 (approved):

PR #2170, opened 2022-09-16
isolated platform authorization 

techsmyth:
query and definition are now split out of the authorizationPolicy module, and are a separate module under platform

Note: this is a pure refactor, no functionality is added or removed: api is the same, platform authorization policy definition is the same, no new database entries.

The issue was that we had all the management of the key platform authorization policy definition within the AuthorizationPolicy module, which was simply wrong.


hero101 (dismissed):

hero101 (approved):

Comoque1:
The auth test suite for global roles "myPrivileges" passed.

PR #2172, opened 2022-09-17
new notificationAdapter module to encapsulate interactions with the notification service

techsmyth:
A pure refactoring, no functionality added or removed. Simply about maintainability of the code, and removing functionality from core/microservices module that did not belong there.

Context for this PR is that it has been clear for a while that there was a mixture of code within the microservices module that did not belong there. When starting on the notifications-143 issue it felt too risky / complicated to address as is to took the step to address the technical debt related to our interaction with the notifications service first. 

## Changes

Removed any interaction with ORM layer from the core/microservices module so that that module now only looks after subscriptions to the messaging layer.

Moved notificationsPayloadBuilder to be a service within the NotificationAdapter

Moved all knowledge of event types for Notifications within the new module

Added DTOs to make clear what is needed from the client service to send a notification event


src/services/platform/notification-adapter/dto/notification.dto.input.aspect.comment.ts
```diff
@@ -0,0 +1,10 @@
+import { IAspect } from '@domain/collaboration';
+import { IComments } from '@domain/communication/comments/comments.interface';
+import { CommunicationMessageResult } from '@domain/communication/message/communication.dto.message.result';
+import { NotificationInputBase } from './notification.dto.input.base';
+
+export class NotificationInputAspectComment extends NotificationInputBase {
```
hero101: why are all DTOs classes?

src/services/platform/notification-adapter/notification.adapter.ts
```diff
@@ -0,0 +1,245 @@
...
+  constructor(
+    @Inject(WINSTON_MODULE_NEST_PROVIDER)
+    private readonly logger: LoggerService,
+    private notificationPayloadBuilder: NotificationPayloadBuilder,
+    @Inject(NOTIFICATIONS_SERVICE) private notificationsClient: ClientProxy
+  ) {}
+
+  async calloutPublished(
+    eventData: NotificationInputCalloutPublished
+  ): Promise<boolean> {
```
hero101: when does it return `false`?
Maybe `void` is more suitable when the same result is always returned.

src/services/platform/notification-adapter/notification.adapter.ts
```diff
@@ -0,0 +1,245 @@
...
+@Injectable()
+export class NotificationAdapter {
+  constructor(
+    @Inject(WINSTON_MODULE_NEST_PROVIDER)
+    private readonly logger: LoggerService,
+    private notificationPayloadBuilder: NotificationPayloadBuilder,
+    @Inject(NOTIFICATIONS_SERVICE) private notificationsClient: ClientProxy
+  ) {}
+
+  async calloutPublished(
```
hero101: At an SSE meeting, we agreed to include all the access modifiers so the code is less error-prone and increases visibility.

hero101 (changes requested):
- There are failing tests.

src/services/platform/notification-adapter/notification.adapter.ts
techsmyth: fixed

src/services/platform/notification-adapter/notification.adapter.ts
techsmyth: updated to be void

src/services/platform/notification-adapter/dto/notification.dto.input.aspect.comment.ts
techsmyth: converted to interfaces

hero101 (approved):

PR #2184, opened 2022-09-21
Archiving of Hubs

techsmyth:
Extend Hub with a field to control its visibility, which can be ACTIVE, ARCHIVED or DEMO. 
- demo visibility is for later

api:
* New mutation to allow updating the hub visibility: updateHubVisibility
* hubs() query: takes optional argument to allow filtering of hubs by visiblity; defaults to active hubs
* metadata query metrics are updated to show only active hubs, challenges, opportunities counts

modules:
* new HubFilter module to manage the filtering of Hubs based on visibility in a consistent way

samples:
* new sample query to get hubs by visibility
* new sample query to get roles by visibility of Hubs for which credentials are for

Key changes:
* Updated HubAuthorizationService so that if Hub is archived then the HubAdmin does not get assigned privileges, nor do any privileges cascade down into the contained entities. 

Note: due to VS Code changes had to mark hubID on Challenge / Opportunity to be optional as otherwise type validations were failing.

Search:
* search is filtering out results based on users access, so to test it out in action run the search as a **non-global-admin** user

There are likely to be additional enhancements identified re filtering as we move demo onto prod + do the FFE e.g.
* metadata metrics to be specified to the url being used?
* ...

To look at it the authorization in action, run the following query:
query {
  hubs {
    nameID
    visibility
    authorization {
      anonymousReadAccess
      credentialRules {
        grantedPrivileges
        resourceID
        type
      }
    }
  }
}

and then run the mutation to update the visibility to be archived. When in active visibility then credential rules looks like: 
image

when archived:

image

### DIFF
--- a/graphql-samples/mutations/update/update-hub-visibility
+++ b/graphql-samples/mutations/update/update-hub-visibility
@@ -8,6 +8,6 @@ Variables:
 {
   "visibilityData": {
     "hubID": "uuid_nameid",
-    "visibility": "active"
+    "visibility": "ARCHIVED"
   }
 }
--- a/graphql-samples/mutations/update/update-hub-visibility
+++ b/graphql-samples/mutations/update/update-hub-visibility
@@ -1,0 +1,13 @@
+mutation updateHubVisibility($visibilityData: UpdateHubVisibilityInput!) {
+  updateHubVisibility(visibilityData: $visibilityData) {
+    id
+  }
+}
+
+Variables:
+{
+  "visibilityData": {
+    "hubID": "uuid_nameid",
+    "visibility": "active"
+  }
+}
--- a/src/domain/challenge/challenge/challenge.entity.ts
+++ b/src/domain/challenge/challenge/challenge.entity.ts
@@ -42,7 +42,7 @@ export class Challenge extends BaseChallenge implements IChallenge {
   parentHub?: Hub;
 
   @Column()
-  hubID?: string;
+  hubID?: string; //toDo make mandatory https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/alkem-io/server/2196
 
   @OneToOne(() => PreferenceSet, {
     eager: false,
--- a/src/domain/challenge/challenge/challenge.entity.ts
+++ b/src/domain/challenge/challenge/challenge.entity.ts
@@ -42,7 +42,7 @@ export class Challenge extends BaseChallenge implements IChallenge {
   parentHub?: Hub;
 
   @Column()
-  hubID!: string;
+  hubID?: string;
 
   @OneToOne(() => PreferenceSet, {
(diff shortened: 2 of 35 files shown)


graphql-samples/mutations/update/update-hub-visibility
```diff
@@ -0,0 +1,13 @@
...
+  updateHubVisibility(visibilityData: $visibilityData) {
+    id
+  }
+}
+
+Variables:
+{
+  "visibilityData": {
+    "hubID": "uuid_nameid",
+    "visibility": "active"
```
valentinyanakiev:
```suggestion
    "visibility": "ACTIVE"
```

valentinyanakiev: `active`, lowercase, is invalid value

src/domain/challenge/challenge/challenge.entity.ts
```diff
@@ -42,7 +42,7 @@ export class Challenge extends BaseChallenge implements IChallenge {
   parentHub?: Hub;
 
   @Column()
-  hubID!: string;
+  hubID?: string;
```
valentinyanakiev: Currently, `hubID` is mandatory. We don't really support nested `Challenge`. Why was this changed?

src/domain/challenge/hub/hub.service.ts
```diff
@@ -256,7 +278,14 @@ export class HubService {
       .leftJoinAndSelect('challenge.opportunities', 'opportunities')
       .getMany();
 
-    const sortedHubs = hubsDataForSorting.sort((a, b) => {
+    const visibleHubs = hubsDataForSorting.filter(hub => {
```
valentinyanakiev, line 281: You could do the filter and sort in the query itself. It will be much faster and the code will be less.

valentinyanakiev (changes requested):

graphql-samples/mutations/update/update-hub-visibility
techsmyth: Fixed

src/domain/challenge/challenge/challenge.entity.ts
techsmyth: There has been a change in VS Code which means that it no longer considers the repositories between Hub / Challenge / Opportunity the same as BaseChallenge if there are mandatory fields in the entity definition. So the whole project is flashing with errors without this change.

src/domain/challenge/hub/hub.service.ts
techsmyth: Yes of course that is faster. But I also wanted to go through the same code path (in the hubFilter service) and also not spend time working a more complex db query. If in the future the number of hubs gets into the 10,000 range then worth considering optimising this, but could not justify that now.

valentinyanakiev (approved):

PR #2199, opened 2022-09-28
Root level hubs query filtered by IDs

valentinyanakiev:
Draft PR for filtering by IDs for root level `hubs` query and `challenges` field resolver on hubs.
 - Added common IDs, LimitAndShuffle, IDs + LimitAndShuffle query args. Could be done better with composition over inheritance for the combined entity.
 - Added filtering of IDs for root level `hubs` query
 - Added parameter for `Challenges` field resolver. There is  //toDo Carlos filter challenges with IDs as I didn't have time to properly filter the IDs today and the filtering needs to be applied.

Tested with `hubs`, not with `Challenges`. Needs to be wrapped up. Opportunities can also be added if needed.


hero101:
### Having the below data
image
### and the below query
image
### generates this error
image

ccanos (approved):

PR #2202, opened 2022-10-01
added preference for allow hub members to create challenges; 

techsmyth:
migration up / down; 
refactored hub authorization service as this had become way too complex

Note: in the process found an issue with assignment of privileges for global roles that would have prevented resetting authorization on archived hubs by global admins.

Primary focus for QA testing needs to be around authorization on Hubs + below.


alstojanovic (approved):
Migration up and down works. Code is good.

PR #2235, opened 2022-10-19
updated entry for new member

techsmyth:
To discuss in combination with a potential client update


techsmyth:
Included in https://github.com/alkem-io/server/pull/2244

PR #2246, opened 2022-10-25
stronger typings / long switch statement refactored

me-andre:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/services/api/activity-log/activity.log.builder.service.ts
+++ b/src/services/api/activity-log/activity.log.builder.service.ts
@@ -8,7 +8,6 @@ import { IActivityLogEntryChallengeCreated } from '@services/api/activity-log/dt
 import { IActivityLogEntryOpportunityCreated } from '@services/api/activity-log/dto/activity.log.dto.entry.opportunity.created';
 import { IActivityLogEntryCalloutCardComment } from '@services/api/activity-log/dto/activity.log.dto.entry.callout.card.comment';
 import { IActivityLogEntryCalloutDiscussionComment } from '@services/api/activity-log/dto/activity.log.dto.entry.callout.discussion.comment';
-import { IUser } from '@src/domain';
 import { ActivityService } from '@platform/activity/activity.service';
 import { UserService } from '@domain/community/user/user.service';
 import { CalloutService } from '@domain/collaboration/callout/callout.service';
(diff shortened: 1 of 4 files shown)


src/services/api/activity-log/activity.log.builder.service.ts
```diff
@@ -0,0 +1,165 @@
...
+import { CanvasService } from '@domain/common/canvas/canvas.service';
+import { ChallengeService } from '@domain/challenge/challenge/challenge.service';
+import { OpportunityService } from '@domain/collaboration/opportunity/opportunity.service';
+import { CommunityService } from '@domain/community/community/community.service';
+
+interface ActivityLogBuilderFunction<TypedActivityLogEntry> {
+  (rawActivity: IActivity): Promise<TypedActivityLogEntry>;
+}
+
+interface IActivityLogBuilder {
```
hero101: I would prefer to enforce exhausting all members of `ActivityEventType` to reduce potential mistakes.

hero101 (changes requested):

src/services/api/activity-log/activity.log.builder.service.ts
techsmyth: One type per file. That is what is done everywhere else on the server and lets keep to that pattern unless we explicitly decide to move away from this. So please create an activity.log.builder.interface file with this type in it. Also the type name should start with "I" to designate an interface.

src/services/api/activity-log/activity.log.builder.service.ts
```diff
@@ -0,0 +1,165 @@
...
+  [ActivityEventType.CALLOUT_PUBLISHED]: ActivityLogBuilderFunction<IActivityLogEntryCalloutPublished>;
+  [ActivityEventType.CARD_CREATED]: ActivityLogBuilderFunction<IActivityLogEntryCalloutCardCreated>;
+  [ActivityEventType.CANVAS_CREATED]: ActivityLogBuilderFunction<IActivityLogEntryCalloutCanvasCreated>;
+  [ActivityEventType.CHALLENGE_CREATED]: ActivityLogBuilderFunction<IActivityLogEntryChallengeCreated>;
+  [ActivityEventType.OPPORTUNITY_CREATED]: ActivityLogBuilderFunction<IActivityLogEntryOpportunityCreated>;
+  [ActivityEventType.CARD_COMMENT]: ActivityLogBuilderFunction<IActivityLogEntryCalloutCardComment>;
+  [ActivityEventType.DISCUSSION_COMMENT]: ActivityLogBuilderFunction<IActivityLogEntryCalloutDiscussionComment>;
+}
+
+export interface ActivityLogEntryBase {
```
techsmyth: One type per file. That is what is done everywhere else on the server and lets keep to that pattern unless we explicitly decide to move away from this. So please create an activity.log.entry.base.interface file with this type in it. Also type name should start with "I" to match the naming convention on the server.

techsmyth (changes requested):
Nice to have the stronger typing and split of functionality

src/services/api/activity-log/activity.log.builder.service.ts
me-andre: Actually exhausting is enforced at usage, where the method is called, you just have to instantiate the class `as Interface`. Otherwise the class itself adds the missing keys/implementations and commenting certain props out has no effect.

techsmyth (approved):
Thanks for splitting - nice improvement overall.

PR #2258, opened 2022-10-29
Search2

techsmyth:
Applying the approach from ActivityLog also to search api.

Key advantages:
- No longer have the Searchable interface which was causing coupling in the code base across core domain entities
- ability to provide more contextual information in search results meaning much less work to do client side,

Initial implementation, some stabilisation still pending. 

graphql samples are updated


valentinyanakiev:
I am not sure how I feel about these changes as a whole, the whole reasoning of the Search service was to be temporary and FAST until we introduce elastic. These changes add a lot of boilerplate code and a lot of types (I am not sure it's an improvement in this case) but I am most concerned about `SearchResultBuilderService`. At first glance looks cool and stylish, but looking at this deeper I see that for each result there are multiple roundtrips made to the database which, for a larger search result set, will result in a horrendous performance. I am not sure this implementation is fit-for-purpose.

techsmyth:
> I am not sure how I feel about these changes as a whole, the whole reasoning of the Search service was to be temporary and FAST until we introduce elastic. These changes add a lot of boilerplate code and a lot of types (I am not sure it's an improvement in this case) but I am most concerned about `SearchResultBuilderService`. At first glance looks cool and stylish, but looking at this deeper I see that for each result there are multiple roundtrips made to the database which, for a larger search result set, will result in a horrendous performance. I am not sure this implementation is fit-for-purpose.

The hits are happening anyway at the moment via follow up calls from the client:
image

Maybe there is a way to make this more effective, but far better to have the calls on the server maybe with caching than looping back to the client + back again.

valentinyanakiev (approved):

PR #2261, opened 2022-10-30
Auth refinements

techsmyth:
Authorization fixes made:
* Enable deletion for user-self
* Registered user querying other's user authorization, returns now return an empty array [ ] by having the field without authorization check
* Global-Hubs-Admins, don't have CREATE_COMMENT privilege on Aspect: new privilege rule to grant the CREATE_COMMENT privilege when the CREATE is already granted.
* Querying other self/other user privileges returns or doesn't grant privilege - the grant privilege, should not be returned when querying user's data: GRANT privilege is removed from platform authorization policy and moved to be directly in Hub + Organization

Not reproducible:  Registered user not member of any community, cannot see the verification_status of the organizations - must have read privilege: the status was readable...

In addition, as part of trying to fix the other authorization issues identified the code to resolve community via other entities (e.g. community from callout etc) has been split into a separate service so it can be re-used.


valentinyanakiev:
I am not sure the user should be able to delete itself. User deletion is, as is, partial / fragmented and there are some pieces that should be done manually if a user is deleted resolving in deletion of the user in all platforms / services (Kratos + 3rd party platforms). The user deleting itself, as is, is misleading as it will be a partial delete.

valentinyanakiev:
`Global-Hubs-Admins, don't have CREATE_COMMENT privilege on Aspect: new privilege rule to grant the CREATE_COMMENT privilege when the CREATE is already granted.` - why was that changed, what was the reasoning?

techsmyth:
> `Global-Hubs-Admins, don't have CREATE_COMMENT privilege on Aspect: new privilege rule to grant the CREATE_COMMENT privilege when the CREATE is already granted.` - why was that changed, what was the reasoning?

The Global Hub Admins should be able to create comments on cards, but the privilege was not assigned to them - but the CREATE one was. So I added a privilege rule to assign the CREATE_COMMENT privilege if a user has the CREATE one.

techsmyth:
> I am not sure the user should be able to delete itself. User deletion is, as is, partial / fragmented and there are some pieces that should be done manually if a user is deleted resolving in deletion of the user in all platforms / services (Kratos + 3rd party platforms). The user deleting itself, as is, is misleading as it will be a partial delete.

I am ok that they have this privilege, even if there are still gaps in the flow. Note that it is not exposed to them on the client at the moment.

PR #2272, opened 2022-11-04
updated activityLogOnCollaboration query to both directly sort and add optional limit param

techsmyth:
Directly doing the sorting in the db query for performance. 

Note: to test also requires server to have the single card server pr merged in as there is a disconnect at the moment client - server develop branches.


alstojanovic:
Re single card server being merged: we are still battling the probable circular dependency issue.

alstojanovic (approved):

PR #2275, opened 2022-11-04
My changes

valentinyanakiev:


techsmyth:
@valentinyanakiev why are you removing the @domain for the imports? That does not have any impact on the circular dependency issues...

alstojanovic:
There is an issue when using barrel imports. One of the issues is that madge is not able to detect circular dependencies correctly. Look at this example:

```js
// Original code
// 1 circular dependency
import { IPreference } from './preference.interface';
import { PreferenceDefinition } from './preference.definition.entity';
import { PreferenceSet } from '../preference-set';

// Fix PreferenceSet import to import only what we need
// 2 circular dependency 
import { IPreference } from './preference.interface';
import { PreferenceDefinition } from './preference.definition.entity';
import { PreferenceSet } from '../preference-set/preference.set.entity';

// Use barrel imports and madge can't detect the circular dependencies anymore.
// 0 circular dependency
import { IPreference } from '@domain/common/preference/preference.interface';
import { PreferenceDefinition } from '@domain/common/preference/preference.definition.entity';
import { PreferenceSet } from '@domain/common/preference-set/preference.set.entity';
```

Example is also the issue with failing unit tests. I resolved all circular dependency issues under common and domain (madge result was 0 circular dependencies detected) but the problematic unit tests were still failing.

techsmyth:
Interesting analysis. Great that the ones in the src/common src/domain are now zero.

Is it possible to do an analysis also using the test tree combined with the src tree?

Btw, I would suggest to merge all these changes down to develop + then take it further from there - there are quite some files impacted I'd imagine.

alstojanovic:
Well, the ones under common and domain are solved according to madge output but it seems some of them are still left because the unit test is failing. I was solving them using @ imports and that might just mask the issues but not solve them. We need to try and solve them with relative path imports.

Take a look at the branch `server-2239-b-cdp`. If you run `madge --circular --extensions ts src/common` or `madge --circular --extensions ts src/domain` you can see it reports 0 circular dependencies but if you try and run unit tests with `npm run test:ci:no:coverage /src/domain/template` you will see the error that is due to a circular dependency.

On this branch, there are 19 circular dependency issues left and they are all in the activity-log and search services.

PR #2288, opened 2022-11-08
Authorization fixes

techsmyth:
Added cascade of authorization to cardTemplate on Aspect

Fixed initiatialization of authorizationPolicy on Profile.

Removed code that was hiding a bug deep in authorization policy service

### DIFF
--- a/src/domain/community/profile/profile.service.ts
+++ b/src/domain/community/profile/profile.service.ts
@@ -41,8 +41,6 @@ export class ProfileService {
   async createProfile(profileData?: CreateProfileInput): Promise<IProfile> {
     const profile: IProfile = Profile.create({
       description: profileData?.description,
-      tagsets: profileData?.tagsetsData,
-      references: profileData?.referencesData,
     });
     profile.authorization = new AuthorizationPolicy();
     profile.avatar = await this.visualService.createVisualAvatar();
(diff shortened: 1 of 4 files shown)


src/domain/community/profile/profile.service.ts
```diff
@@ -41,21 +41,29 @@ export class ProfileService {
...
     profile.authorization = new AuthorizationPolicy();
     profile.avatar = await this.visualService.createVisualAvatar();
     profile.location = await this.locationService.createLocation(
       profileData?.location
     );
 
-    if (!profile.references) {
-      profile.references = [];
+    profile.references = [];
+    if (profileData?.referencesData) {
```
valentinyanakiev, line 55: With this code, if you put a `reference` with value, e.g. `https://alkem.io` and you already have a `reference` with value `https://alkem.io` you will end up having semantically the same reference twice. It is the same for `tags` in a `tagset`. I don't believe this is desired behavior.

valentinyanakiev (changes requested):

src/domain/community/profile/profile.service.ts
valentinyanakiev: My bad, there is a check in the reference service

valentinyanakiev (approved):

PR #2303, opened 2022-11-10
communityPolicy 

techsmyth:
The community policy requires a lot more flexibility to support the use cases needed for the wider community management setup asked for by customers.

As such it makes sense to have this as a separate module, with its own entity table etc so that we can evolve the logic around it.

No api changes beyond that there is now an id on the policy entity below community. 

Key updates:
- new entity: CommunityPolicy
- CommunityEntity has two fields: member, lead
- Each field has a rolePolicy as a string, which includes the parent credentials for that role
e.g.

```json
{
    "credential": {
        "type": "opportunity-member",
        "resourceID": "12a93a68-cc66-4c97-8f23-32541252bcc0"
    },
    "minOrg": 0,
    "maxOrg": -1,
    "minUser": 0,
    "maxUser": -1,
    "parentCredentials": [
        {
            "type": "challenge-member",
            "resourceID": "1ff979f0-6e81-4f7a-a6db-df63c93f12dc"
        },
        {
            "type": "hub-member",
            "resourceID": "ef9c808b-425f-4d61-84cc-eb8b269f4a8f"
        }
    ]
}
```

Primary testing is ensuring that after migration that community policies are still working and enforced when adding members / leads.


src/domain/community/community-policy/community.policy.interface.ts
```diff
@@ -1,11 +1,8 @@
...
 @ObjectType('CommunityPolicy')
-export abstract class ICommunityPolicy {
-  @Field(() => ICommunityPolicyRole)
-  member!: ICommunityPolicyRole;
-
-  @Field(() => ICommunityPolicyRole)
-  lead!: ICommunityPolicyRole;
+export abstract class ICommunityPolicy extends IBaseAlkemio {
+  member!: string;
+  lead!: string;
```
hero101, line 7: Why were the decorators removed?

src/domain/community/community/community.interface.ts
```diff
@@ -22,7 +23,7 @@ export abstract class ICommunity extends IAuthorizable {
 
   parentCommunity?: ICommunity;
 
-  policy!: string;
+  policy?: ICommunityPolicy;
```
hero101: why is the policy optional now?

src/migrations/1668109973567-communityPolicy.ts
```diff
@@ -0,0 +1,183 @@
...
+        leadPolicy.parentCredentials = parentLeadCredentials;
+        const leadStr = JSON.stringify(leadPolicy);
+
+        await queryRunner.query(
+          `UPDATE community_policy SET member = '${memberStr}', lead = '${leadStr}' WHERE (id = '${communityPolicy.id}')`
+        );
+      }
+    }
+
+    // Todo:
```
hero101: What about this todo?

src/domain/community/community-policy/community.policy.interface.ts
techsmyth: because there are field resolvers for those fields in community.policy.resolver.fields

src/migrations/1668109973567-communityPolicy.ts
techsmyth: Ah good catch - I was struggling with gettng the FK to work on the relationship enforcment community-community-policy. Can you please have a quick look to see what I was missing?

src/domain/community/community/community.interface.ts
techsmyth: Logically it is required, but TypeORM cannot deal with required entity-entity relationship.

techsmyth (commented):
Thanks for reviewing - 2 of the 3 are easy enough to explain + one I need a bit of help on.

src/domain/community/community/community.interface.ts
hero101: I don't see any issues with querying the data when it's required.

techsmyth: I just made the change...and it is working. I am seriously puzzled....it has been a TypeORM limitation since we started using it that for entity relationships that the fields were optional. Maybe it was only for 1-many and that 1-1 are ok? If so we can probably simplify out quite some 1-1 relationships + undefined checking in the server...

Please also try it + if required works then go ahead + merge; we can raise another card for the systematic review.

hero101 (approved):
Tests are run locally.
The only outstanding issue is the `CARD_MOVE` privilege which is covered or will be covered with #2312.

PR #2312, opened 2022-11-13
communityPolicy plus

techsmyth:
This PR updates the authorization framework to:
- make the rules easier to read
- make it easier to pick up the policy at multiple levels of the hierarchy
- sets the foundation for key enhancements such as being able to have hub members contribute at all levels

It likely will also deliver a slight performance gain as there should be significantly less rules in the database after this. Further optimization is possible. 

CommunityPolicy is passed throughout the authorization hierarchy, instead of the membershipCredential. 

CommunicationPolicy holds a set of flags that are calculated separately (and only in memory) at the start of the authorization cascade (easier to follow). 

AuthorizationPolicy service extended with helper methods to make it easier to create a credential rule with multiple criteria.

Platform Authorization and Hub authorization updated to leverage new credential rules creation setup - which drastically improves the readability of the authorization logic (in my opinion)

Removed MOVE_CARD from platform level authorization and down to the AspectAuthoriationService

...


Comoque1:
@techsmyth, I get error on `createCallenge` mutation.
```
console.log
    {
      errors: [
        {
          message: "Cannot read properties of undefined (reading 'type')",
          locations: [Array],
          path: [Array],
          extensions: [Object]
        }
      ],
      data: null
    }

```
This impacts almost all api tests.

techsmyth:
@Comoque1 thanks for running the tests. I have just pushed what I think will fix it - it was using the hub instead of the challenge community policy...

hero101:
Closing in favour of #2319 .

PR #2313, opened 2022-11-14
added auth invalid policy exception that throws if finds a cred rule …

techsmyth:
…with no critieria; migration throws an error if it finds an auth policy without cred rules


techsmyth:
@valentinyanakiev this is the code created during the debugging Monday morning - worth having in as some extra checks.

valentinyanakiev (approved):
Code makes sense and I don't get real errors on real (`Demo`) dataset.

PR #2319, opened 2022-11-17
communityPolicy plus

hero101:
This PR updates the authorization framework to:
- make the rules easier to read
- make it easier to pick up the policy at multiple levels of the hierarchy
- sets the foundation for key enhancements such as being able to have hub members contribute at all levels

It likely will also deliver a slight performance gain as there should be significantly less rules in the database after this. Further optimization is possible. 

CommunityPolicy is passed throughout the authorization hierarchy, instead of the membershipCredential. 

CommunicationPolicy holds a set of flags that are calculated separately (and only in memory) at the start of the authorization cascade (easier to follow). 

AuthorizationPolicy service extended with helper methods to make it easier to create a credential rule with multiple criteria.

Platform Authorization and Hub authorization updated to leverage new credential rules creation setup - which drastically improves the readability of the authorization logic (in my opinion)

Removed MOVE_CARD from platform level authorization and down to the AspectAuthoriationService

...


Comoque1:
Global Admin Hubs role, doesn't have `MOVE_CARD` privilege

Comoque1:
Users don't have rights to apply for a community
image

techsmyth:
@Comoque1 @hero101 good to get clear where the gaps are. Both ones above look like straight forward ones - I can have a look later - but maybe Svetlo can already address them?

techsmyth:
@Comoque1 just pushed commits for both of the above issues, great if you can check if they are now resolved (they are working locally)

Comoque1:
The `application` issue seems to be working now.
Another one appeared, when deleting Challenge / Opportunities (on clean DB)

Comoque1:
And one more:
Registered user not member of a community previously was receiving error: `Authorization: unable to grant 'create-aspect' privilege: create aspect on callout"`, when try to create card on a closed callout. Now the error that is received is: `"New collaborations to a closed Callout with id: 'fac60068-58e1-4a6f-b538-56ae0b63b0a5' are not allowed!\`, as the once received from users members of community, who attempt to create card on closed callout.

techsmyth:
@Comoque1  re deletion one above, that will require an update to the migration 
@hero101 see error above re deleting - this will also require a change to the migration which you will be doing anyway for 

UPDATE: just pushed a separate branch to fix the deletion issue: https://github.com/alkem-io/server/pull/2322 - that should be merged down to dev, and then for anywhere the the migration has been run already do a revert + reapply the migration e.g. on de (npm run migration:revert, npm run migration:run). Alternative is to create another migration to fix. Thoughts?

techsmyth:
@Comoque1  re error message shown, looking at the code am pretty sure the user is a member. Can you please show the authorization credential rules on that entity and also the myPrivileges results?

Comoque1 (approved):

PR #2321, opened 2022-11-17
IPFS Upload + Delete

valentinyanakiev:
To test upload:

```bash
actionUrl=$(\
    curl -s -X GET -H "Accept: application/json" \
    "http://localhost:3000/identity/ory/kratos/public/self-service/login/api" \
    | jq -r '.ui.action'\
    )
sessionToken=$(\
curl -s -X POST -H  "Accept: application/json" -H "Content-Type: application/json" \
    -d '{"password_identifier": "admin@alkem.io", "password": "your_password", "method": "password"}' \
    "$actionUrl" | jq -r '.session_token' \
    )
```

You can test (assuming default endpoint configuration) creating a file and then uploading it with the following CURL request:

```bash
touch hello-alkemio.pdf

curl http://localhost:3000/api/private/non-interactive/graphql \
 -H "Authorization: Bearer $sessionToken" \
 -F operations='{"query":"mutation UploadFile($file:Upload!) {uploadFile(file:$file)}", "variables": { "file": null }}' \
 -F map='{ "0": ["variables.file"] }' \
 -F 0=@"hello-alkemio.pdf"
```

Deletion is actual [unpinning](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/PIN.md#ipfspinrmipfspath-options) in IPFS + forced garbage collection. Removal of a file is in the graphql-samples, e.g.:

```gql
mutation deleteFile($deleteData: DeleteFileInput!) {
  deleteFile(deleteData: $deleteData)
}
```

```json
{
  "deleteData":
  {
    "CID": "Qmde6CnXDGGe7Dynz1pnxgNARtdVBme9YBwNbo4HJiRy2W"
  }
}

```

After unpinning a garbage collection is enforced and triggered (note: it may take some time to execute GC + do a hard reload on your browser to validate when a file is removed as files, including IPFS ones, are often in cache...).

Supported upload fileS:
- bmp
- jpeg
- x-png
- png
- gif
- pdf

NOTE: to support files e.g. Excel or Word will be an extensive list. Usually different versions of these applications have different mime types.

NOTE2: IPFS has different APIs for its mutable file system (MFS) and the IPFS file system. There are IPFS paths (e.g. `/ipfs/Qmde6CnXDGGe7Dynz1pnxgNARtdVBme9YBwNbo4HJiRy2W` and MFS paths (e.g. `/tmp/hello-alkemio.pdf`) and these are not compatible.


src/domain/common/file-manager/file.manager.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    fileName: string,
+    mimetype: string
+  ): Promise<string> {
+    if (!readStream)
+      throw new ValidationException(
+        'Readstream should be defined!',
+        LogContext.FILE_MANAGER
+      );
+
+    if (!mimetype)
```
hero101: `readStream` and `mimetype` are not nullable to begin with.

src/domain/common/file-manager/file.manager.service.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      return true;
+    } catch (error: any) {
+      throw new IpfsDeleteFailedException(
+        `Ipfs removing file at path ${CID} failed! Error: ${error.message}`
+      );
+    }
+  }
+
+  private async validateMimeTypes(mimeType: string): Promise<boolean> {
```
hero101: this function is synchronous

src/domain/common/file-manager/file.manager.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    const mimeTypes: string = this.configService.get(ConfigurationTypes.STORAGE)
+      ?.ipfs.mime_types;
+    const allowedMimeTypes: string[] = mimeTypes.split(',');
+    if (!allowedMimeTypes.includes(mimeType)) return false;
```
hero101:
```javascript
return allowedMimeTypes.includes(mimeType)
```

src/domain/common/file-manager/file.manager.service.ts
```diff
@@ -0,0 +1,73 @@
...
+        `Tried to import invalid mimetype ${mimetype}!`,
+        LogContext.FILE_MANAGER
+      );
+    }
+
+    const buffer = await streamToBuffer(readStream);
+
+    try {
+      return await this.ipfsService.uploadFileFromBuffer(buffer);
+    } catch (error: any) {
```
hero101: The correct type is `unknown` here. 
Try something like the below.  [More on the topic](https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript)
 ```javascript
(error as Error).message ?? String(error)
```

File: src/services/adapters/ipfs/ipfs.service.ts [hunk 5]
```diff
@@ -1,29 +1,30 @@
...
+import { CID, create, IPFSHTTPClient } from 'ipfs-http-client';
+import { IpfsDeleteFailedException } from '@common/exceptions/ipfs.delete.exception';
+import { IpfsGCFailedException } from '@common/exceptions/ipfs.gc.exception';
 
 @Injectable()
 export class IpfsService {
-  private ipfsEndpoint = '';
-  private ipfsClientEndpoint = '';
+  private ipfsClient: IPFSHTTPClient;
+  private ipfsClientEndpoint;
```
hero101: Fields assigned only in the constructor can be made `readonly`.

src/services/adapters/ipfs/ipfs.service.ts
```diff
@@ -41,13 +42,47 @@ export class IpfsService {
...
+        );
+
+        if (gcFile.err) {
+          this.logger.error(
+            `Error in collection ${gcFile.err}`,
+            LogContext.IPFS
+          );
+        }
+      }
+    } catch (error: any) {
```
hero101, line 80: same as the other related comment

src/platform/authorization/platform.authorization.service.ts
```diff
@@ -129,11 +132,18 @@ export class PlatformAuthorizationService {
...
       AuthorizationPrivilege.CREATE
     );
     privilegeRules.push(createPrivilege);
 
+    const deletePrivilege = new AuthorizationPolicyRulePrivilege(
+      [AuthorizationPrivilege.FILE_DELETE],
+      AuthorizationPrivilege.DELETE
+    );
+    privilegeRules.push(deletePrivilege);
+
```
hero101, line 146: Could you explain how is this tide up to a specific role?
I don't believe the privileges are assigned correctly as per the AC.

src/domain/common/file-manager/file.manager.service.ts
```diff
@@ -0,0 +1,73 @@
...
+      );
+    }
+  }
+
+  async removeFile(CID: string): Promise<boolean> {
+    try {
+      await this.ipfsService.unpinFile(CID);
+      await this.ipfsService.garbageCollect();
+
+      return true;
```
hero101: return the success outside of the try block

src/domain/common/file-manager/file.manager.resolver.mutations.ts
```diff
@@ -0,0 +1,74 @@
...
+      `file upload: ${filename}`
+    );
+    const readStream = createReadStream();
+    const uri = await this.fileManagerService.uploadFile(
+      readStream,
+      filename,
+      mimetype
+    );
+
+    return uri;
```
hero101: return the method directly

src/domain/common/file-manager/file.manager.resolver.mutations.ts
```diff
@@ -0,0 +1,74 @@
...
+      this.platformAuthorizationService.getPlatformAuthorizationPolicy();
+
+    await this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      authorizationPolicy,
+      AuthorizationPrivilege.FILE_DELETE,
+      `file delete: ${CID}`
+    );
+
+    return await this.fileManagerService.removeFile(CID);
```
hero101: `await` is not needed

src/domain/common/file-manager/file.manager.resolver.mutations.ts
```diff
@@ -0,0 +1,74 @@
...
+    );
+
+    return uri;
+  }
+
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => Boolean, {
+    description: 'Removes a file.',
+  })
+  async deleteFile(
```
hero101: missing method modifiers as per the SEE agreement

src/domain/common/file-manager/file.manager.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    try {
+      return await this.ipfsService.uploadFileFromBuffer(buffer);
+    } catch (error: any) {
+      throw new IpfsUploadFailedException(
+        `Ipfs upload of ${fileName} failed! Error: ${error.message}`
+      );
+    }
+  }
+
+  async removeFile(CID: string): Promise<boolean> {
```
hero101: missing method modifiers as per the SEE agreement

hero101 (changes requested):
- Move the ipfs exceptions to an ipfs folder, to improve visibility
- I believe we should create a scalar type for the CID and validate it via a package like [s-ipfs](https://github.com/ipfs-shipyard/is-ipfs) for example
- the `max_file_size` var in `alkemio.yml` is 1 MB - and if reused, must be 5MB per the AC

src/domain/common/file-manager/file.manager.service.ts
valentinyanakiev: I agree about the CID, will have a look. Also thought about the ipfs exceptions and it makes sense, but it should be applied structurally as a pattern for all types, and we haven't done it in the past, hence I didn't do it. I will have a look at only it and what would make sense for the rest (won't have refactoring more items in this PR thou, just addressing the comment). The file size I haven't changed but I think your comment is redundant as this is development configuration and the prod configuration is provided in a variable that we can change / manage separately.

src/platform/authorization/platform.authorization.service.ts
valentinyanakiev: There were no acceptance criteria for deletion at all, I have discussed this with Neil and taken this approach. Permissions in this space will need to tidy up regardless as the AC was not granularly defined.

PR #2322, opened 2022-11-17
fixed migration with FK properly set

techsmyth:
FK was previously not dealing with null


techsmyth:
@Comoque1  if this works then you should be just able to apply it to dev + then the deletion should work

hero101:
Pushed in the existing migration at #2319.

PR #2324, opened 2022-11-19
Authorization framework cleanup + new preferences

techsmyth:
New preferences for Challenges:
- allowing members to create opportunity
- allowing to be private (there but not yet used in business logic)
- allowing hub members to contribute

Note: preference for turning challenges private is not yet in use.

New privileges:
- CONTRIBUTE, used to cascade privileges through collaboration
- CREATE_OPPORTUNITY, used in client PR: https://github.com/alkem-io/client-web/pull/2972

Updated privileges:
- UPDATE_CANVAS

Business logic:
- Converted most of the rules creation to use authorization policy service rather than direct instantiation. 
- simplified the rules in some places, also combining them where possible to reduce the overall set of rules which is nice for both performance and logic readability
- Opportunity authorization services also now uses the community policy

Client:
- separate issue to allow the updating of the preferences via the admin UI: https://github.com/alkem-io/client-web/issues/2973

Test suites
========
Changes should primarily affect myPrivileges results. Any other changes should be flagged as an error.
- Collaboration entities + below: additional CONTRIBUTE privilege, and UPDATE_CANVAS privilege should now only be on the Canvas entity
- Challenge entity: new CREATE_OPPORTUNITY privilege for GHA / GA + admins of the Hub / Challenge. 

There will be additional test cases needed for the additional preferences and flags, I would suggest these are covered separately and take this PR to be able to run against existing test cases.


Comoque1:
GlobalAdminHubs, don't have 'Contribute' privilege to Hub / Challenge collaboration
image

There is no 'Contribute' privilege for Opportunity for both GlobalAdmin and GlobalAdminHubs

techsmyth:
> GlobalAdminHubs, don't have 'Contribute' privilege to Hub / Challenge collaboration image
> 
> There is no 'Contribute' privilege for Opportunity for both GlobalAdmin and GlobalAdminHubs

Good catch, this is now fixed with the last commit.

Comoque1:
The above fixes, work!
Is it expected opportunity.collaboration.relations the GHA and GA to have contribute privilege?
image

techsmyth:
yes CONTRIBUTE cascades into all entities in a Collaboration

Comoque1:
Hub Member, doesn't have rights to send comment to a card created from GA
image

Comoque1:
The test: `one opportunity admin to remove another admin from opportunity` is failing with error:
image

Comoque1:
When preference: `ALLOW_HUB_MEMBERS_TO_CONTRIBUTE`:**false**, and `ALLOW_CONTRIBUTORS_TO_CREATE_OPPORTUNITIES`: **true** user doesn't have rights to create opportunity
image

techsmyth:
Please raise this as another issue
Sent from my iPhoneOn 24 Nov 2022, at 12:07, Evgeni Dimitrov ***@***.***> wrote:﻿
When preference: ALLOW_HUB_MEMBERS_TO_CONTRIBUTE:false, and ALLOW_CONTRIBUTORS_TO_CREATE_OPPORTUNITIES: true user doesn't have rights to create opportunity

techsmyth:
> When preference: `ALLOW_HUB_MEMBERS_TO_CONTRIBUTE`:**false**, and `ALLOW_CONTRIBUTORS_TO_CREATE_OPPORTUNITIES`: **true** user doesn't have rights to create opportunity image

actually the logic is correct. If a user is not a member of the challenge, then the user is not a contributor if the hub members flag is false - so the user should not have the privilege to create an opportunity

Comoque1:
> > When preference: `ALLOW_HUB_MEMBERS_TO_CONTRIBUTE`:**false**, and `ALLOW_CONTRIBUTORS_TO_CREATE_OPPORTUNITIES`: **true** user doesn't have rights to create opportunity image
> 
> actually the logic is correct. If a user is not a member of the challenge, then the user is not a contributor if the hub members flag is false - so the user should not have the privilege to create an opportunity

In that case, there must be logic to not allow `ALLOW_CONTRIBUTORS_TO_CREATE_OPPORTUNITIES`: **true** if `ALLOW_HUB_MEMBERS_TO_CONTRIBUTE`:**false**, because the first is dependent on the second.

Still, I was thinking that `ALLOW_HUB_MEMBERS_TO_CONTRIBUTE` relates to user privileges to operate on collaboration level (`CREATE_CALLOUT`, `CREATE_ASPECT`, etc..) and `ALLOW_CONTRIBUTORS_TO_CREATE_OPPORTUNITIES` purely to allow user creating an Opportunity (even as placeholder)

valentinyanakiev (approved):

PR #2340, opened 2022-11-25
Extended geo verbose logging

valentinyanakiev:


hero101 (dismissed):
Build is fixed in #2345

PR #2354, opened 2022-12-02
updated to use new event names for notifications 

techsmyth:
event payloads are updated to send over additional information.

trigger notification when a user account is deleted

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "alkemio-server",
-  "version": "0.37.2",
+  "version": "0.37.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "alkemio-server",
-      "version": "0.37.2",
+      "version": "0.37.3",
       "license": "EUPL-1.2",
       "dependencies": {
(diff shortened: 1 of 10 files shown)


valentinyanakiev:
In the `notifications.payload.builder.ts`, you have:

  ```typescript
async buildApplicationCreatedNotificationPayload(
    applicationCreatorID: string,
    applicantID: string,
    community: ICommunity
  ) {
    const hubPayload = await this.buildJourneyPayload(community);
    const payload: any = {
      applicationCreatorID,
      applicantID,
      community: {
        name: community.displayName,
        type: community.type,
      },
      hub: hubPayload,
    };

    return payload;
  }
```

This is not updated, not strongly-typed and not corresponding to what is expected in the notifications service.

techsmyth:
@valentinyanakiev yes it seems some of the event types were not strongly typed. That should now be addressed - all payloads that are build using the notification adapter now use types. Makes the case for the notifications-lib very strongly - it was simply a lot of 'any's that were being used in the old code + I had missed some.

valentinyanakiev (approved):

PR #2386, opened 2022-12-15
added recommendations without content being populated

techsmyth:
adds a new field to Context: recommendations

always exactly 3 recommendations

check to avoid mutation for deleting a reference being used to remove a Recommendation

migration up / down working.

Note: data is not migrated, this can be done either via a script or manually post release as a lot of the hubs will anyway require manual updates.


valentinyanakiev (commented):
Logically, this will work but deviates from our approach. This whole approach cries out loud for ReferenceSet entity, similar to Tagset. Having multiple pointers to a named set of the same type shows this smell:

```typescript
  @ManyToOne(() => Context, context => context.references, {
    eager: false,
    cascade: false,
    onDelete: 'CASCADE',
  })
  context?: Context;

  @ManyToOne(() => Context, context => context.recommendations, {
    eager: false,
    cascade: false,
    onDelete: 'CASCADE',
  })
  contextRecommendation?: Context;
```

techsmyth:
Fully agree, can you add an issue for this onto the backlog?

valentinyanakiev:
I just found a bug running this query, an old one, I will fix it...:

```
query hubs {
  hubs {
    context
    {
      recommendations
      {
        description
      }
    }
  }
}
```

Reference.description was nullable in the DB but not the API. This will easily be hit on the client as reference descriptions with the migration are `null`.

valentinyanakiev:
> Fully agree, can you add an issue for this onto the backlog?

https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/server/2397

valentinyanakiev (approved):

PR #2387, opened 2022-12-15
added sorting by activity for callouts field on Collaboration

techsmyth:
Args for callouts field on Collaboration also split out into a separate type.


valentinyanakiev (changes requested):
`UUID_NAMEID` removal regresses the client, it shouldn't be done that way.

techsmyth:
@valentinyanakiev there is also a PR for the NAMEID removal on the client

valentinyanakiev (approved):

PR #2393, opened 2022-12-19
added publisher meta info to callout; …

techsmyth:
added mutation to allow setting published information for global admins via a new privilege

Open: to set the publishedDate to be the same as the createdDate or not? I was not able to get this to work easily - needs a look. Also ties into the wider discussion of timestamps vs dates etc.


src/domain/collaboration/callout/dto/callout.dto.update.publisher.ts
```diff
@@ -0,0 +1,24 @@
...
+    description: 'The identifier of the publisher of the Callout.',
+  })
+  publisherID!: string;
+
+  @Field(() => Number, {
+    nullable: true,
+    description: 'The timestamp to set for the publishing of the Callout.',
+  })
+  publishDate!: number;
+}
```
alstojanovic: This dto is named `UpdateCalloutPublisherInput` but it is changing both the publisher and publishDate. Consider changing the name.

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -122,6 +124,22 @@ export class CalloutService {
     return await this.calloutRepository.save(callout);
   }
 
+  public async updateCalloutPublishInfo(
+    callout: ICallout,
+    publisherID: string,
+    publishedTimestamp?: number
+  ): Promise<ICallout> {
```
alstojanovic: 1. Use a proper dto for input type.
2. The implementation suggests that PublisherID is always changed and publishDate cannot be changed without changing the publisherID. Do we want it like this or make both optional so that we can choose what we want to change?

src/domain/collaboration/callout/callout.resolver.mutations.ts
```diff
@@ -225,6 +233,31 @@ export class CalloutResolverMutations {
     return result;
   }
 
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => ICallout, {
+    description: 'Update the visibility of the specified Callout.',
```
alstojanovic: Description is wrong.

src/domain/collaboration/callout/callout.resolver.mutations.ts
```diff
@@ -225,6 +233,31 @@ export class CalloutResolverMutations {
...
+  })
+  @Profiling.api
+  async updateCalloutPublisher(
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args('calloutData') calloutData: UpdateCalloutPublisherInput
```
alstojanovic: If this mutation needs to change only the publisher then we need an appropriate dto for only sending the publisherID.

alstojanovic (changes requested):

src/domain/collaboration/callout/dto/callout.dto.update.publisher.ts
techsmyth: Ok, will rename to UpdateCalloutPublishInformationInput

src/domain/collaboration/callout/callout.service.ts
techsmyth: 1. No, this is a method inside the servie, there is a DTO used on the mutation which is where it is required. We do not pass around DTO
2. Ok, both can be optional

src/domain/collaboration/callout/callout.resolver.mutations.ts
techsmyth: Ok, will fix

src/domain/collaboration/callout/callout.resolver.mutations.ts
techsmyth: No, one mutation to set either, will rename the mutation

alstojanovic (approved):

PR #2395, opened 2022-12-20
added check that a callout is published 

techsmyth:
before triggering notifications on 
* aspect created, 
* canvas created or 
* comment on discussion callout

This issue was hit live today at a session with UWV.


alstojanovic (approved):

Comoque1:
The issue is not related to the attached bug.

PR #2410, opened 2022-12-22
filter callouts that can be returned to a user based on the privileges

techsmyth:
The field "callouts" on a Collaboration now returns results as follows:
- if visibility is published, then check the READ privilege
- if visibility is draft, then check the UPDATE privilege

Not ideal, as would prefer to do it all via authorization policy, but the alternative was a very deep update to the authorization setup which needs a deeper design.


Comoque1:
Tested OK locally with this server branch and develop client branch.

techsmyth:
Good to hear @Comoque1

valentinyanakiev (approved):

PR #2411, opened 2022-12-22
only trigger activity entries for published callouts

techsmyth:


Comoque1:
When card callout is unpublished and community member checks its activity log :
 - The card creation is not displayed anymore.
image

 - Cards comment creation is displayed on activity log. The user is able to open the card from the comment activity and to send a comment as well
image

techsmyth:
@Comoque1 the card / canvas / comment top level activity entries are no longer generated. Yes the comments on card one still is, but that is actually quite tricky to not have triggered as the card that the mutation is on does not have access to the Callout. In practice I expect the primary use case will be that an admin creates a callout (card / comments / canvas) and does some initial population - which should not trigger activity entries. If they add comments on cards then indeed the activity entry still gets generated. I can block access to the callout based on the visibility...

techsmyth:
@Comoque1 actually in combination with the other fix re callouts being filtered before returned based on whether the user should have access or not, then I think that the callout will not actually be visible...can you please test in combination with the other PR for issue #3154? You may want to approve this one, and then bring it up into the other PR...the activity log entry will still show up, but the user will not be able to load the referenced callout.

valentinyanakiev (approved):

PR #2424, opened 2023-01-03
Added top level queries for context, community + collaboration

techsmyth:
Authorization is applied, so users need to have read privilege to be able to execute the query.


alstojanovic (approved):
Code look good. Tested the queries locally as global admin.

PR #2426, opened 2023-01-03
Platform entity with communication

techsmyth:
New Platform entity that is a singleton

Platform entity has a field for a Communication entity that is created on bootstrap if not existing

Library entity also now accessed via Platform entity

Configuration information for clients such as URLs for support etc now have a different entity name.

PlatformAuthorizationService renamed to PlatformAuthorizationPolicyService to keep it isolated from the logic of propagating authorzation through the Platform entity. 

Client updates added:
- access to library

cli updates added:
- reset auth on platform level


src/domain/communication/communication/communication.service.authorization.ts
```diff
@@ -5,33 +5,33 @@ import { IAuthorizationPolicy } from '@domain/common/authorization-policy/author
...
 
+    communication.authorization = this.appendPrivilegeRules(
+      communication.authorization
+    );
+
     communication.authorization = this.extendAuthorizationPolicy(
-      communication.authorization,
-      policy
+      communication.authorization
     );
```
alstojanovic, line 35: Do we need this call if we are not extending the policy with any rules?

src/platform/platfrom/platform.service.ts
```diff
@@ -0,0 +1,84 @@
...
+import { InjectRepository } from '@nestjs/typeorm';
+import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import { FindOneOptions, Repository } from 'typeorm';
+import { Platform } from './platform.entity';
+import { IPlatform } from './platform.interface';
+
+@Injectable()
+export class PlatformService {
+  constructor(
+    private libraryService: LibraryService,
```
alstojanovic: This service is unused, please remove.

alstojanovic (changes requested):

src/domain/communication/communication/communication.service.authorization.ts
techsmyth: it could probably go, but good to keep it there as it is highly likely some rules will be added there near term.

src/platform/platfrom/platform.service.ts
techsmyth: ok.

alstojanovic (approved):

PR #2432, opened 2023-01-07
fixed callout published notification

techsmyth:
Trivial fix.
Also made a generic log event function for logging event data in notification adapter.


alstojanovic (approved):
Tested locally, the notification is received.

PR #2434, opened 2023-01-07
tidied up validation on callout + comments; refined messageID scalar type

techsmyth:
Quite a lot of validation was missing on the Callout entity related DTOs


alstojanovic (approved):
There was a broken import and the validations were not working because they were not added to the base handler. Code is generally ok and now works good.
Fixed validations on creating Callout description and posting comment on discussion Callout.

PR #2458, opened 2023-01-12
Added self management for callouts

valentinyanakiev:


techsmyth (approved):
Working nicely. It solves the case that a user creating a callout can see it.

PR #2462, opened 2023-01-13
Authorization safety

techsmyth:
Split the static authorization policies into two:
- root: that is used to inherit by other entities such as user, org, hub, platform etc
- platform: which has the specific rules that govern platform level permissions

This avoids any situation whereby modification to platform can impact the rules applied to other entities. It is simply much safer.

Also fixed an issue with the library as it should inherit from platform authorization via a parameter rather than going directly.

Note: kept the authorization policy as a separate module to avoid having linked dependencies between lots of the domain modules and the platform module.

### DIFF
--- a/src/domain/challenge/challenge/challenge.service.authorization.ts
+++ b/src/domain/challenge/challenge/challenge.service.authorization.ts
@@ -234,7 +234,7 @@ export class ChallengeAuthorizationService {
 
     authorization = this.authorizationPolicyService.reset(authorization);
     authorization =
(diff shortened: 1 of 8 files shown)


valentinyanakiev (approved):

valentinyanakiev (approved):
logically makes sense but before merging the auth test suite needs to be executed

Comoque1:
Create hub mutation fails: `"message": "Please provide a card template",`
image

techsmyth:
@Comoque1 nothing to do with the authoriztion changes, this is a bug introduced with the canvas template addition to callout

PR #2471, opened 2023-01-15
Added allowed discussion categories to communication

techsmyth:
Migration to add discussionCategories field to Communication and populate existing ones with the appropriate defaults
Logic to set the allowed categories at communication creation
Logic to check that a valid category was supplied

Fixed issue with authorization policy preventing global admins from creating discussions.

Note: this includes the fix for canvas template that was blocking testing of this issue.


alstojanovic (approved):
Code looks good. Tested the migration up/down. Tested the sample mutation.

PR #2481, opened 2023-01-17
allow global roles to be able to send messages to a discussion

techsmyth:


Comoque1:
Global admin can send comment with this branch.
Private hub member cannot send comment. @techsmyth, could you please have a look?
image

techsmyth:
Actually the end solution for this is that communication needs to move out of community and sit alongside collaboration etc. We do not have discussions active on the journeys at the moment so please raise an issue and put it in the ice-box - cannot pick it up now...

Comoque1 (approved):

PR #2507, opened 2023-01-27
gave credentialRules names and logged which granted the requested privilege

techsmyth:
Added a name to all credentialRules:
image

Added logging so that the rule used to grant a privilege is logged:
image


valentinyanakiev (changes requested):
- Exported all names to constants
 -  Grouped them to:
    -  credential rules
    -  typed credential rules
    -  policy rules
    - global policy rules
 - I saw that you had implicitly used capital letters for policy rules - I have made them lowercase, I think that is too implicit and we need to define a standard format for the source of the rule, e.g. credential rule - ..., privilege rule - ... etc

techsmyth:
@valentinyanakiev is there action needed from me on this? I am fine with the changes you have made, just keen to see it in develop...

valentinyanakiev (approved):

PR #2510, opened 2023-01-27
made all time entries dates

techsmyth:


techsmyth:
@valentinyanakiev  is there any blocker on this one going in?

PR #2526, opened 2023-02-03
added optional parameter to publish mutation on callout to control if notification is sent

techsmyth:
Parameter is optional, and defaults to true so that existing mutation usage is not affected. 

Separate issue to be raised to allow this flag to be set from the client, for now we can use this via the api.

### DIFF
--- a/graphql-samples/mutations/update/update-callout-visibility
+++ b/graphql-samples/mutations/update/update-callout-visibility
@@ -8,6 +8,7 @@ Variables:
 {
   "calloutData": {
     "calloutID": "uuid",
-    "visibility": "published"
+    "visibility": "published",
+    "notification": true
   }
 }
--- a/graphql-samples/mutations/update/update-callout-visibility
+++ b/graphql-samples/mutations/update/update-callout-visibility
@@ -8,7 +8,7 @@ Variables:
 {
   "calloutData": {
     "calloutID": "uuid",
-    "visibility": "published",
-    "notification": true
+    "visibility": "PUBLISHED",
+    "sendNotification": true
   }
 }
--- a/src/domain/collaboration/callout/callout.resolver.mutations.ts
+++ b/src/domain/collaboration/callout/callout.resolver.mutations.ts
@@ -217,9 +217,7 @@ export class CalloutResolverMutations {
           Date.now()
         );
 
-        // Optionally trigger a notification
-        const sendNotification = calloutData.notification;
-        if (sendNotification) {
+        if (calloutData.sendNotification) {
           const notificationInput: NotificationInputCalloutPublished = {
             triggeredBy: agentInfo.userID,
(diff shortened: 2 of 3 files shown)


graphql-samples/mutations/update/update-callout-visibility
```diff
@@ -8,6 +8,7 @@ Variables:
 {
   "calloutData": {
     "calloutID": "uuid",
-    "visibility": "published"
+    "visibility": "published",
```
alstojanovic: Needs to be uppercase `"visibility": "PUBLISHED"`

src/domain/collaboration/callout/dto/callout.dto.update.visibility.ts
```diff
@@ -18,4 +18,11 @@ export class UpdateCalloutVisibilityInput {
     description: 'Visibility of the Callout.',
   })
   visibility!: CalloutVisibility;
+
+  @Field(() => Boolean, {
+    nullable: true,
+    defaultValue: true,
+    description: 'Send a notification on publishing.',
+  })
+  notification!: boolean;
```
alstojanovic: I would name the field `sendNotification`.

src/domain/collaboration/callout/callout.resolver.mutations.ts
```diff
@@ -217,11 +217,15 @@ export class CalloutResolverMutations {
           Date.now()
         );
 
-        const notificationInput: NotificationInputCalloutPublished = {
-          triggeredBy: agentInfo.userID,
-          callout: savedCallout,
-        };
-        await this.notificationAdapter.calloutPublished(notificationInput);
+        // Optionally trigger a notification
+        const sendNotification = calloutData.notification;
```
alstojanovic: If the parameter is renamed to `sendNotification` then it would be obvious what it's meant for even without the assignment to a variable.

alstojanovic (changes requested):
Tested, works fine. 
Just update the sample query and I would change the parameter name to `sendNotification` and remove `const sendNotification = calloutData.notification`.

graphql-samples/mutations/update/update-callout-visibility
techsmyth: Yep also saw that but forgot to fix it.

src/domain/collaboration/callout/dto/callout.dto.update.visibility.ts
techsmyth: agreed this is better

src/domain/collaboration/callout/callout.resolver.mutations.ts
techsmyth: agreed

alstojanovic (approved):

PR #2545, opened 2023-02-08
Filtering by hub for challenges, opportunities, users, cards

techsmyth:
Added parameter to search query to filter to a particular Hub
Removed parameter not used for filtering to Challenges

Added filterIDs for all filterable searched: challenges, opportunities, users, cards

Updated all search queries to optionally filter by ID.

Separately: return the hierarchy of the card so that the client can create links to it.


techsmyth:
@valentinyanakiev I started looking at this issue, but it has some heavy implications to do it in a way that is performant + that does imply all kinds of loops on the client. Suggest we try to find some time tomorrow / friday to talk this one through as non-trivial I think...

NS> updated: found a clean setup to filter by Hub, please have a look.

valentinyanakiev (approved):
Seems to be working. Could be easier for `journeys`, as they have `hubID` and a simple optional left join on the queries would work the same as the rest. `User` requires credentials for membership, so is necessary, and `Card` requires `RIGHT JOIN`, which is not possible on typeORM repositories, so also necessary. Anyway, I have tested it, and seems to be OK. Also I see no real usage of search in challenges in the client so I am going to merge this.

PR #2550, opened 2023-02-10
SearchResultCard expanded to include nameIDs of all parents

valentinyanakiev:
That's how a result looks at the moment:

image

With all parent nameIDs needed to construct an URL. Should be sufficient for the client.


src/services/api/search/dto/search.result.dto.entry.card.ts
```diff
@@ -15,4 +15,30 @@ export abstract class ISearchResultCard
     description: 'The Card that was found.',
   })
   card!: IAspect;
+
+  @Field(() => String, {
+    nullable: false,
+    description: 'The Hub nameID of the Card.',
+  })
+  hub!: string;
```
techsmyth: suggest to name the field hubNameID - and do the same for all the fields, otherwise it is confusing...also please look at the description.

techsmyth (changes requested):
Looks like it will provide the information to create the links, which is enough for now. I can imagine later we may want to expose the full entities directly - but that can be later.

So only some minor naming updates and then this can go in.

valentinyanakiev:
> Looks like it will provide the information to create the links, which is enough for now. I can imagine later we may want to expose the full entities directly - but that can be later.
> 
> So only some minor naming updates and then this can go in.

I will update the names. For exposing the full entities - I am not sure. It will require additional DB calls and was not in scope in the AC, so it was going to make matters worse in the scope of this task. If the requirement arises in the future - it is a different story.

techsmyth (approved):

PR #2555, opened 2023-02-10
allow supression of event transition errors

techsmyth:
Stop errors being triggered if event transition fails

Requires client to also be updated

### DIFF
--- a/src/domain/common/canvas-checkout/canvas.checkout.lifecycle.options.provider.ts
+++ b/src/domain/common/canvas-checkout/canvas.checkout.lifecycle.options.provider.ts
@@ -38,15 +38,26 @@ export class CanvasCheckoutLifecycleOptionsProvider {
 
     this.logCheckoutStatus(eventName, canvasCheckout, 'event triggered');
 
-    await this.lifecycleService.event(
-      {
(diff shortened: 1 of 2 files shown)


src/domain/common/canvas-checkout/canvas.checkout.lifecycle.options.provider.ts
```diff
@@ -38,15 +38,26 @@ export class CanvasCheckoutLifecycleOptionsProvider {
...
+      await this.lifecycleService.event(
+        {
+          ID: canvasCheckout.lifecycle.id,
+          eventName: eventName,
+        },
+        this.CanvasCheckoutLifecycleMachineOptions,
+        agentInfo,
+        canvasCheckout.authorization
+      );
+    } catch (error) {
```
me-andre, line 51: If possible, I would add a check for some specific error attribute or message match to make sure it's a transition error.

techsmyth: This whole area needs a more detailed review, for now I think this is enough - simply higher priority items we need to get after...

valentinyanakiev (approved):
Works for me

PR #2559, opened 2023-02-11
ApplicationForm per Community

techsmyth:
Allow the questions asked for applications to a Community to be configurable, something that is coming in from all active NL pilots.

New entity Form, that contains a list of questions and a description. It is generic, and designed to be re-used - for example in a SurveyCallout.

Added applicationForm field to each Community, with default entries per community type

Mutation to update the applicationForm on a Community, including sample usage.

Migration up / down.

Removed old global template content from configuration service that is replaced by this functionality - as well as some other obsolete template information that is not used.


valentinyanakiev:
@techsmyth this feature is kind of blocked because applications are broken at the moment in develop, I can not verify whether it's working. Will be working on resolving other blockers until this is testable.

valentinyanakiev:
Managed to circumvent the issues and tested it, will be raising separate bugs with our findings...

valentinyanakiev (approved):

PR #2574, opened 2023-02-15
updated banner image dimensions to allow for more resolution

techsmyth:
Max width now 1536
Min width now 384
==> both a factor of 2 up

Min height now corrected to be 64 (32 was too low by a factor of 2).

Migration up / down added, and also corrected defaults in the code.

### DIFF
--- a/src/domain/common/visual/visual.service.ts
+++ b/src/domain/common/visual/visual.service.ts
@@ -145,9 +145,9 @@ export class VisualService {
       {
         name: 'banner',
         minWidth: 384,
-        maxWidth: 768,
-        minHeight: 32,
-        maxHeight: 128,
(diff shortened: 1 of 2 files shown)


src/domain/common/visual/visual.service.ts
```diff
@@ -145,9 +145,9 @@ export class VisualService {
       {
         name: 'banner',
         minWidth: 384,
```
ccanos, line 147: is there a reason to not increase the minWitdth also?

ccanos: ok, 32x6 was 192, this was already 384. all good

ccanos (approved):

PR #2600, opened 2023-02-25
Generic Profile entity, base for wider usage

techsmyth:
Updates the Profile entity to be ready for wider usage
- added tagline
- added displayName
- multiple visuals

Note: left nameID on User  / Organization as that feels like a core identifier so that should not be in a sub-entity. Moved displayName to Profile. 

New fields on Profile
- to load a particular visual: visual(visualType)
- to load the default tagset: tagset

Migration up / down is added for profile update.

Migrated all cardProfile usage to be Profile

Updated usage of Profile entity so that the containing entity specifies the set of tagsets to be used; this was previously not properly controlled. 

Tagsets can no longer be specified as part of the profile creation, only updated afterwards. 

Graphql samples updated

Note: needed to remove the data loader for avatar on user / org, so worth paying attention to what impact that has on performance. 

Configuration provided by the server releated to tagsets is removed. 

Client updates needed:
- no longer specify the creation of tagsets on user / org creation; this is now handled by the server cleanly
- move displayName on calendarEvent create/update, and ditto for aspect


Comoque1:
`deleteUser` request is failing
image

Comoque1:
`updateUser` request is failing
image

Comoque1:
Querying organization `tagline` throws an error: `"message": "Cannot return null for non-nullable field Profile.tagline.",`
image

Comoque1:
Querying organization `tagline` throws an error: `"message": "Cannot return null for non-nullable field Profile.tagline.",`
image

Comoque1:
Organizations list on admin page is failing: `"message":"No entity column "displayName" was found."`
image

Comoque1:
Organizations list on admin page is failing: `"message":"No entity column "displayName" was found."`
image

Comoque1:
All notifications tests fail:
image

src/domain/collaboration/callout/callout.entity.ts
```diff
@@ -19,7 +19,7 @@ import { AspectTemplate } from '@domain/template/aspect-template/aspect.template
 import { CanvasTemplate } from '@domain/template/canvas-template/canvas.template.entity';
 
 @Entity()
-export class Callout extends NameableEntity implements ICallout {
+export class Callout extends NameableEntityOld implements ICallout {
```
valentinyanakiev, line 22: That name is bizarre, why do we have `NameableEntityOld`?

src/platform/configuration/config/template/template.interface.ts
```diff
@@ -1,11 +1,9 @@
 import { IOpportunityTemplate } from './opportunity.template.interface';
 import { IUserTemplate } from './user.template.interface';
-import { IOrganizationTemplate } from './organization.template.interface';
 
 export interface ITemplate {
   name: string;
   description?: string;
   users?: IUserTemplate[];
   opportunities?: IOpportunityTemplate[];
-  organizations?: IOrganizationTemplate[];
```
valentinyanakiev, line 10: How is this change related to the PR?

src/platform/configuration/config/template/user.template.interface.ts
```diff
@@ -10,6 +10,5 @@ export interface IReferenceTemplate {
 }
 export interface IUserTemplate {
   name: string;
-  tagsets?: ITagsetTemplate[];
```
valentinyanakiev, line 13: How is this change related to the PR?

Comoque1:
Messages cannot be sent to Organization
image

Comoque1:
Mention message body must return user mentioned display name instead of markdown
image

src/domain/collaboration/callout/callout.entity.ts
alstojanovic: This is only for the transition period until we finish with other entities (journeys, callouts and canvases). They use the nameable entity with both nameID and displayName. We could name it differently but for the transition period I'm ok with this.

alstojanovic:
Sending direct message to organization and markdown in text are fixes are pushed. (markdown fix is in notifications service)

src/platform/configuration/config/template/template.interface.ts
techsmyth: The tagsets are created now by the entity holding the profile; they are no longer created by the client. This was used by the client to create tagsets - so it has to be removed.

src/platform/configuration/config/template/user.template.interface.ts
techsmyth: Ditto re comment on org template: the tagsets are now created by the entity, not by the client

PR #2610, opened 2023-03-02
Typeorm Upgrade

valentinyanakiev:
A few major notes:
   - Transitive eager loaded dependencies below 3rd level (A --> B --> C --> D), B will eagerly load C, but C won't eager load D. Added explicit loading of those entities on all places we found in the test-suites repo.
   - .create invokes the default constructor of the `Entity`. I have altered some definitions to match the database - mandatory (not null) fields in the DB can (and should) be mandatory in the entities as well, and then they are not required in the constructor.
   - FindByIds is replaced with `FindBy`, `where` and `In`
   - Find + where works differently, the whole `where` is inside the `options` now
   - Typeorm introduces the concept of a `DataSource` now. It is used by the migrations, migrations altogether use different parameters
   - `OrmConfig` concept in TypeORM is dropped
   - To create a new migration, run `npm run migration:run src/migrations/[name_of_migration]` - this is slightly different to the behavior before
   - `DataSources` use named connections. The default name is `default`. It needs to be provided when the new @InjectEntityManager is used. It simplifies usage of repositories in scenarios where we want to get any repository generically. 
   - There are a lot of deprecated methods...
   - I have added `| never` when we query data and throw an exception
   - `undefined` is largely replaced as `null` as return type
   - `findOne` default behavior is changed now - if we pass no parameters it doesn't return the first entry of the table, it returns `undefined`


Comoque1:
Previously, when creating an `actorGroup`, without actors, the attribute `actors` where `null` now it is `[ ]`
image

Also when querying created `actors`, the `actorGroup` returns null
image

Comoque1:
Aspect template tagsets are returning `null` instead of the actual value
image

Comoque1:
Querying group parent by organization type, returns, community type:
image

valentinyanakiev:
> Aspect template tagsets are returning `null` instead of the actual value image

that is fixed now

valentinyanakiev:
> Previously, when creating an `actorGroup`, without actors, the attribute `actors` where `null` now it is `[ ]` image
> 
> Also when querying created `actors`, the `actorGroup` returns null image

fixed

Comoque1:
Community Admin doesn't receive notification, when the preference is enabled.

src/domain/agent/credential/credential.entity.ts
```diff
@@ -18,10 +18,7 @@ export class Credential extends BaseAlkemioEntity implements ICredential {
   })
   agent?: Agent;
 
-  constructor(type: string, resourceID: string) {
+  constructor() {
```
hero101, line 21: why was this required?

src/domain/agent/credential/credential.service.ts
```diff
@@ -19,13 +19,13 @@ export class CredentialService {
...
   ): Promise<ICredential> {
-    const credential = Credential.create(credentialInput);
+    const credential = Credential.create({ ...credentialInput });
     await this.credentialRepository.save(credential);
     return credential;
   }
 
   async getCredentialOrFail(credentialID: string): Promise<ICredential> {
-    const credential = await this.credentialRepository.findOne({
+    const credential = await this.credentialRepository.findOneBy({
```
hero101, line 28: in agent.service.ts findOne was used instead - which change is the correct one or it's was because of the provided options?

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -601,24 +609,24 @@ export class ChallengeService {
...
-    }[] = await getConnection().query(sqlQuery);
+    }[] = await this.entityManager.connection.query(sqlQuery);
 
     return queryResult.challengesCount;
   }
 
   async getChildChallengesCount(challengeID: string): Promise<number> {
-    return await this.challengeRepository.count({
-      where: { parentChallenge: challengeID },
+    return await this.challengeRepository.countBy({
```
hero101, line 626: why

src/domain/collaboration/agreement/agreement.entity.ts
```diff
@@ -26,9 +23,8 @@ export class Agreement extends BaseAlkemioEntity implements IAgreement {
   @JoinColumn()
   tagset: Tagset;
 
-  constructor(name: string) {
+  constructor() {
     super();
-    this.name = name;
-    this.tagset = new Tagset(RestrictedTagsetNames.DEFAULT);
+    this.tagset = new Tagset();
```
hero101, line 28: this is possibly changing behaviour

src/domain/collaboration/aspect/aspect.service.ts
```diff
@@ -57,7 +57,7 @@ export class AspectService {
   public async deleteAspect(deleteData: DeleteAspectInput): Promise<IAspect> {
     const aspectID = deleteData.ID;
     const aspect = await this.getAspectOrFail(aspectID, {
-      relations: ['profile'],
+      relations: ['profile', 'profile.tagset'],
```
hero101: i don't see tagset used in this method

src/domain/collaboration/aspect/aspect.service.ts
```diff
@@ -132,7 +132,7 @@ export class AspectService {
 
   public async getCardProfile(aspect: IAspect): Promise<ICardProfile> {
     const aspectLoaded = await this.getAspectOrFail(aspect.id, {
-      relations: ['profile'],
+      relations: ['profile', 'profile.tagset'],
```
hero101: the proper way to do this imo is to provide either options or the relation param - the method is loading the profile from a card, the tagset is used just in one use-case where the method is used, the other use-cases probably won't need the tagset loaded here.

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -450,7 +450,7 @@ export class CalloutService {
     shuffle?: boolean
   ): Promise<IAspect[]> {
     const loadedCallout = await this.getCalloutOrFail(callout.id, {
-      relations: ['aspects'],
+      relations: ['aspects', 'aspects.comments'],
```
hero101: comments are not needed for the impl of this method, a use-case is requiring the comments to be loaded

src/domain/common/preference-set/preference.set.service.ts
```diff
@@ -66,11 +66,11 @@ export class PreferenceSetService {
   async getPreferenceSetOrFail(
     preferenceSetID: string,
     options?: FindOneOptions<PreferenceSet>
-  ): Promise<IPreferenceSet> {
-    const preferenceSet = await PreferenceSet.findOne(
-      { id: preferenceSetID },
-      options
-    );
+  ): Promise<IPreferenceSet | never> {
+    const preferenceSet = await PreferenceSet.findOne({
```
hero101, line 70: in nvp.service.ts findOne is changed with findOneBy, here it isn't- why?

src/domain/community/community/community.service.ts
```diff
@@ -420,7 +420,7 @@ export class CommunityService {
 
   async getCommunication(communityID: string): Promise<ICommunication> {
     const community = await this.getCommunityOrFail(communityID, {
-      relations: ['communication'],
+      relations: ['communication', 'communication.updates'],
```
hero101: the imp is not dependant on updates

src/domain/community/organization/organization.service.ts
```diff
@@ -216,6 +216,7 @@ export class OrganizationService {
         'agent',
         'verification',
         'preferenceSet',
+        'preferenceSet.preferences',
```
hero101: where is this used

hero101: where is this used

src/domain/community/organization/organization.service.ts
```diff
@@ -447,7 +448,7 @@ export class OrganizationService {
     );
     // Try to find the organization
     const organization = await this.getOrganizationOrFail(orgID, {
-      relations: ['groups'],
+      relations: ['groups', 'groups.profile'],
```
hero101, line 450: not used in the impl

src/domain/community/profile/profile.service.ts
```diff
@@ -208,8 +208,11 @@ export class ProfileService {
   async getProfileOrFail(
     profileID: string,
     options?: FindOneOptions<Profile>
-  ): Promise<IProfile> {
-    const profile = await Profile.findOne({ id: profileID }, options);
+  ): Promise<IProfile | never> {
+    const profile = await Profile.findOne({
+      where: { id: profileID },
+      ...options,
```
hero101, line 213: the where in options will override the where

src/domain/community/user/user.service.ts
```diff
@@ -284,7 +284,12 @@ export class UserService {
   async deleteUser(deleteData: DeleteUserInput): Promise<IUser> {
     const userID = deleteData.ID;
     const user = await this.getUserOrFail(userID, {
-      relations: ['profile', 'agent', 'preferenceSet'],
+      relations: [
+        'profile',
+        'agent',
+        'preferenceSet',
+        'preferenceSet.preferences',
```
hero101: where is this used

src/library/library/library.service.ts
```diff
@@ -24,7 +24,7 @@ export class LibraryService {
   ) {}
 
   async getLibraryOrFail(): Promise<ILibrary> {
-    const library = await this.libraryRepository.findOne();
+    const library = await this.libraryRepository.findOne({});
```
hero101, line 27: so... will this work?

src/platform/platfrom/platform.service.ts
```diff
@@ -22,9 +22,11 @@ export class PlatformService {
   ) {}
 
   async getPlatformOrFail(
-    options?: FindOneOptions<Platform>
-  ): Promise<IPlatform> {
-    const platform = await this.platformRepository.findOne(options);
+    options: FindOneOptions<Platform>
```
hero101: why aren't the options optional?

src/platform/platfrom/platform.service.ts
```diff
@@ -22,9 +22,11 @@ export class PlatformService {
...
+  ): Promise<IPlatform | never> {
+    let platform: IPlatform | null = null;
+    platform = (await this.platformRepository.find({ take: 1, ...options }))[0];
```
hero101:
```typescript
const platform = (await this.platformRepository.find({ take: 1, ...options }))?.[0];
```
will find return empty array of falsy?
the above code works both ways

src/platform/platfrom/platform.resolver.queries.ts
```diff
@@ -16,7 +16,7 @@ export class PlatformResolverQueries {
     description: 'Alkemio Platform',
   })
   async platform(): Promise<IPlatform> {
-    return await this.platformService.getPlatformOrFail();
+    return await this.platformService.getPlatformOrFail({});
```
hero101: why were the options changed to mandatory?

hero101 (changes requested):
Let's go through the comments together because some of them are from lack of knowledge about what changed, others can be marked as won't fix.

src/domain/collaboration/agreement/agreement.entity.ts
valentinyanakiev: The `name` column of `Tagset` is annotated with a default:

```typescript
  @Column({ default: RestrictedTagsetNames.DEFAULT })
  name: string;
```

It makes this call and parameter redundant.

src/domain/community/organization/organization.service.ts
valentinyanakiev: this is duplicate with the one above

hero101 (approved):

PR #2613, opened 2023-03-04
Journey migration to Profile

techsmyth:
Update Journeys to use Profile entity.

Data migration up / down prototyped.
Domain model logic updated to use Profile for Hub / Challenge/Opportunity

Simplified how Profile entities are created: now the creating entity does all the creation of Tagsets/Visuals.

Migration of data:
- Hub/Challenge/Opportunity tagset ==> Profile.Tagset
- Hub/Challenge/Opportunity displayName ==> Profile.displayName
- Context.tagline ==> Profile.tagline
- Context.background ==> Profile.description
- Context.location ==> Profile.location
- Cotext.references ==> Profile.references
- Context.visuals ==> Profile.visuals
- Added profile field to Hub / Challenge / Opportunity

Note: the above changes remove a lot of linkages within the code base, and simplify the code significantly.

Removed mutation addReferenceToContext

For later: creation of ReferenceSet to be able to also migration Recommendations on Context.

Sample graphql queries updated

To do:
- test with new setup
- test suites updated
- client-lib / populator updated


Comoque1:
Input parameters for all journeys contain `referencesData` under profile, but for opportunityUpdate and challengeUpdate  mutations, it is required `references`. We should keep it consistent.
image
image

techsmyth:
@Comoque1 agreed this needs to be consistent, and it goes further: why is it profileData? It maybe should be just profile. That said, I would suggest we pick this up as a separate issue rather than delay this branch going through. Can you please raise it separately?

Comoque1:
Error is thrown when user query its own data:
image.

The interesting part is that the same error is thrown, when that same user creates an entity (challenge/organization), but the actual id is user related and not entity related.
image

valentinyanakiev:
Closing in favour of https://github.com/alkem-io/server/pull/2653

PR #2614, opened 2023-03-06
NestJS native data loader instantiation

hero101:
**The file structure isn't final, but it was easier to have all the files related to one feature in the same place.**


hero101:
@valentinyanakiev The file structure isn't final, but it was easier to have all the files related to one feature in the same place.

valentinyanakiev (approved):

PR #2620, opened 2023-03-09
Generic findByBatch util for data loaders

hero101:
Depends on #2618.


src/core/dataloader/utils/findByBatchIds.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const resultsById = new Map<string, TResult>(
+    results.map<[string, TResult]>(x => [x.id, x[relation]])
+  );
+  // ensure the result length matches the input length
+  return ids.map(
+    id =>
+      resultsById.get(id) ??
+      new EntityNotFoundException(
+        `Could not load user ${id}`,
```
valentinyanakiev: It is not always a user, and `LogContext` need to be injected

src/core/dataloader/utils/findByBatchIds.ts
```diff
@@ -0,0 +1,57 @@
...
+const parentAlias = 'parent';
+const resultAlias = 'relation';
+
+export const findByBatchIds = async <
+  TParent extends { id: string } & { [key: string]: any }, // todo better type
+  TResult
+>(
+  repo: Repository<TParent>,
+  ids: string[],
+  relation: EntityRelations<TParent>,
```
valentinyanakiev, line 21:
```suggestion
  relation: EntityRelations<TParent>[],
```

valentinyanakiev (changes requested):

src/core/dataloader/utils/findByBatchIds.ts
hero101: This is addressed with a better and typed approach in the latest changes.

valentinyanakiev (approved):

PR #2625, opened 2023-03-11
Profiles: applied to other entities

techsmyth:
The following entities now have profiles:
- Callout
- Canvas
- InnovationPack
- Prooject
- AspectTemplate
- LifecycleTemplate
- CanvasTemplate

The TemplateInfo is removed and replaced by Profile.
The NameableOld base entity is removed.

Migration:
- Added up / down. 

The Nameable base entity type now has a profile. 

Graphql samples updated.

To do:
- test migration on real data
- dataloaders added for new profiles


Comoque1:
Error is thrown when create a hub

techsmyth:
@Comoque1 this may be an issue with the test...can you describe the scenario? Important is to know where the test is trying to access the "tagset" field

Comoque1:
> @Comoque1 this may be an issue with the test...can you describe the scenario? Important is to know where the test is trying to access the "tagset" field

The issue was happening as I was querying `tagset` as part of Organization data, but it is unavailable for User and Organization Profiles. Updated in the tests.

Comoque1:
Update callout is failing:
image
image

Comoque1:
Cannot create aspect on default callout
image

Comoque1:
Sending message on discussion callout, throws an error
image

Comoque1:
all server issues above are resolved

PR #2627, opened 2023-03-12
Mutation for providing the callouts sort order

techsmyth:
New mutation updateCalloutsSortOrder that sets the sort order for the provided callouts to be increasing as per the order the callouts are provided in the arguments.

Sample graphql query is provided. 

Note: this PR does also have the fix for the index on the migration removed for server-2540 - needed to get the server to be able to startup. But that can be ignored for the purpose of this PR. 

Note: at the moment the PR is merging into server-2540, which is also where the PR forks from, as that is where migrations are running, after the profiles work goes to develop we may want to do a new PR.


src/domain/collaboration/collaboration/collaboration.service.ts
```diff
@@ -363,4 +364,51 @@ export class CollaborationService {
...
+      if (callout.sortOrder < minCalloutSortOrder) {
+        minCalloutSortOrder = callout.sortOrder;
+      }
+      calloutsInOrder.push(callout);
+    }
+
+    // Now update all the callouts sort order
+    for (const callout of calloutsInOrder) {
+      callout.sortOrder = minCalloutSortOrder;
+      minCalloutSortOrder += 2;
```
alstojanovic, line 411: Should be 1?

techsmyth: I thought 2 gave a bit of space; does not really matter as long as they go up...

alstojanovic (approved):
Tested locally, works fine.

PR #2649, opened 2023-03-15
Downgraded to typeorm@0.3.11, which resolves the paths correctly on win

hero101:
https://github.com/typeorm/typeorm/issues/9840#issuecomment-1464948483
> I found out the reason is the version of the dependency "glob" and the use of PlatformTools.pathNormalize in the file [DirectoryExportedClassesLoader.ts](https://github.com/typeorm/typeorm/blob/master/src/util/DirectoryExportedClassesLoader.ts) at line 37.
When using glob@8.1.0, patterns using backslashes as directory separator doesn't seem to work.
If you're on Windows, the forward slashes in the migrations and entities paths are replaced by backslashes when typeorm calls pathNormalize, so glob can't find any files.
Typeorm should make sure all path separators are forward slashes instead of using OS's default.
I fixed my code by downgrading typeorm from version 0.3.12 to 0.3.11.
I think it's simple to resolve. I'm rushing another project right now, but if nobody opens a PR in the next days, I will do it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "redlock": "^5.0.0-beta.2",
         "replace-special-characters": "^1.2.5",
         "rxjs": "^7.3.0",
-        "typeorm": "0.3.12",
+        "typeorm": "0.3.11",
         "winston": "^3.3.3",
         "xstate": "^4.26.1",
         "yaml": "^2.0.0-7"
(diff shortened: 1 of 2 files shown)


ccanos (approved):

hero101:
Merging it - seems to work on Windows 10, as tested by @ccanos.

valentinyanakiev:
This is not a small change, although it might be perceived as one, literally requires FULL regression testing... (in theory it does not, if typeorm guys are diligent about semantic versioning BUT...).

Comoque1:
There is no screenshot of integrations suites, but there are only from the failing ones. after rerun, they were fine.

I am fine to proceed with publishing of this changes to ACC and the a smoke testing from the client.

PR #2668, opened 2023-03-20
added migration to update the set of types for visuals to allow usage of webp images

techsmyth:


alstojanovic (commented):
I think there is a constructor for Visual that may need an update as well.

techsmyth:
@alstojanovic yes good catch - that constructor also needs to be updated.

techsmyth:
@alstojanovic fixed the constructor

alstojanovic (approved):
Tested migration and both up and down work.

PR #2689, opened 2023-03-25
renamed inheritable to cascade

techsmyth:
much better name that makes the code clearer


hero101 (approved):
Also renamed the field in some test data

PR #2730, opened 2023-04-07
Authorizable private rest endpoint

hero101:
- private endpoint accessible through `/api/private/rest/*
- `/api/private/rest/storage` endpoint available and authorizable
- comments added to reduce the ambiguity and questions raised through the process


techsmyth (approved):
Nicely done, and good to have the clarity re guard handling in the comments

PR #2754, opened 2023-04-26
Recommendations migrated to link callouts

hero101:
- Requires an authorization reset to view the new callouts through the API

PR #2763, opened 2023-04-27
Secure file storage

valentinyanakiev:
- there is a sample downloading a file locally from IPFS. 
- to test, upload a file, get its CID (externalID) and execute HTTP GET (you can use browser) on http://localhost:3000/api/private/rest/storage/document/{CID}
- your file will be downloaded in the root directory
- file is returned `as is` the type returned from IPFS - AsyncIterable<Uint8Array>. This will require further work to be handled either on the client or on the server (remains an open question WHAT exactly format we need returned?)


valentinyanakiev:
@techsmyth please have a look, the return type requires some thinking. I am currently returning document value `as is` returned by the IPFS client. There are libraries that can handle it on the client. It can also be converted on the server - the big question is, to what... (and if it is indeed needed to be converted)

techsmyth:
@valentinyanakiev code looks clean. 

Re api, important is that it is the Document UUID that is specified - not the externalID. The code I think is doing it properly but this needs to be very explicit. 

Re return type, we have the mimetype so we can presumably set that? It should be that we are able to set the right response headers like any other response request. I would suggest looking at the headers that are returned when we e.g. retrieve a document or image from IPFS, and ensure the same are set on our responses from this api call.

techsmyth:
@valentinyanakiev and for removing the documentBucket from Profile that still needs to be added?

valentinyanakiev:
> @valentinyanakiev and for removing the documentBucket from Profile that still needs to be added?

I added it in https://github.com/alkem-io/server/pull/2754 - as without it, the migration was failing...

valentinyanakiev:
> @valentinyanakiev code looks clean.
> 
> Re api, important is that it is the Document UUID that is specified - not the externalID. The code I think is doing it properly but this needs to be very explicit.
> 
> Re return type, we have the mimetype so we can presumably set that? It should be that we are able to set the right response headers like any other response request. I would suggest looking at the headers that are returned when we e.g. retrieve a document or image from IPFS, and ensure the same are set on our responses from this api call.

I can set the `Content-Type` header with the `mime type` and return the same type, if that is sufficient. It is as straightforward as it gets...

techsmyth:
`key criteria is that the client can retrieve and work with the content the same whether it goes via this route or directly via ipfs.` - that's pretty tough key criteria. That's what I wanted to highlight yesterday but I believe it didn't come across to you.  We don't really have a web server serving files with this approach (apart from IPFS nodes, that we want to close going forward) and the client won't be retrieving content the same way. We have an HTTP API with a GET request, that should return the data in some format, which was not specified originally. I can return the same type as what a specific IPFS method (`CAT`, as per specification) is returning, but that doesn't mean it is the same format returned by the **IPFS Gateway**.

valentinyanakiev:
@techsmyth @hero101 managed to resolve the main issue where I was downloading the file - now everything happens in memory an is efficient. Now the main AC where endpoint can access e.g. an image `as is` is met. Performance will be evaluated separately on Acc.

PR #2768, opened 2023-05-01
removed concept of group on communication

techsmyth:
This concept has been deprecated in more recent versions of matrix. 

To do:
- remove communicationGroupID field from db


valentinyanakiev (approved):
Best code update --> Deleting code :) Worked seamlessly for me... :+1:

PR #2772, opened 2023-05-02
Link callout now in group HOME_0

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/migrations/1682490418949-recommendationsToLinkCollection.ts
+++ b/src/migrations/1682490418949-recommendationsToLinkCollection.ts
@@ -8,6 +8,10 @@ export class recommendationsToLinkCollection1682490418949
   name = 'recommendationsToLinkCollection1682490418949';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE \`profile\` DROP COLUMN \`storageBucketId\``
+    );
+
     const references: { contextRecommendationId: string }[] =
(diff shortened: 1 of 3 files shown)


src/migrations/1682490418949-recommendationsToLinkCollection.ts
```diff
@@ -8,10 +8,6 @@ export class recommendationsToLinkCollection1682490418949
   name = 'recommendationsToLinkCollection1682490418949';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(
```
valentinyanakiev: If you remove this your migration will fail.

valentinyanakiev (changes requested):

src/migrations/1682490418949-recommendationsToLinkCollection.ts
hero101: It exists in `storage1681736452222` in both up and down.

valentinyanakiev: Yes, it does, but it is the opposite logic. This is cleaning it up, as we changed the approach in the previous PR(for the storage buckets), which was worked simultaneously with the Link Callouts. The migration for the storage buckets added a new field to profile that breaks the Link Callouts migration. Therefore the need to remove it before anything else.

valentinyanakiev (approved):

PR #2779, opened 2023-05-03
Fixed issue with unauthenticated users not seeing main page queries

valentinyanakiev:


hero101 (approved):

hero101:
@techsmyth Can you shed some light on this change, because it's causing issues to the graphql API

techsmyth:
@hero101 I presume you are talking about the second commit? I was not involved - so would need to understand why @valentinyanakiev made that change.

PR #2787, opened 2023-05-04
Fixed uploading files on callout / aspect on opportunity

valentinyanakiev:


Comoque1 (approved):

Comoque1:
The fix is tested locally and addresses both: BUG: When new update canvas on opportunity, console error is thrown #3941 and BUG: Uploading files on Opportunity callout and aspectfails #3948

PR #2792, opened 2023-05-05
Migrating Visuals and References 

valentinyanakiev:
to be filled in


src/domain/storage/document/document.module.ts
```diff
@@ -1,5 +1,5 @@
 import { AuthorizationPolicyModule } from '@domain/common/authorization-policy/authorization.policy.module';
-import { Module } from '@nestjs/common';
+import { Module, forwardRef } from '@nestjs/common';
```
techsmyth: I would like to avoid forwardRefs - they are typically a smell that something needs to be decoupled.

src/domain/storage/document/document.module.ts
```diff
@@ -16,7 +16,7 @@ import { IpfsModule } from '@services/adapters/ipfs/ipfs.module';
     AuthorizationPolicyModule,
     AuthorizationModule,
     TagsetModule,
-    UserModule,
+    forwardRef(() => UserModule),
```
techsmyth: this is old I think - it should not be needed

src/domain/storage/storage-bucket/storage.bucket.module.ts
```diff
@@ -21,7 +21,7 @@ import { ReferenceModule } from '@domain/common/reference/reference.module';
     AuthorizationPolicyModule,
     IpfsModule,
     VisualModule,
-    EntityResolverModule,
+    forwardRef(() => EntityResolverModule),
```
techsmyth: This direction is ok, it is the other direction that needs to go

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
```diff
@@ -169,4 +169,14 @@ export class StorageBucketResolverMutations {
     };
     return await this.referenceService.updateReference(updateData);
   }
+
+  @UseGuards(GraphqlGuard)
```
techsmyth: this mutation does not belong on this module. It is a pure admin mutation, so please put it in the platform/admin module. This will also resolve the circular dependency. 

Also the name should be "adminStorageMigrateIpfsUrls" or something along those lines.

src/services/infrastructure/entity-resolver/entity.resolver.module.ts
```diff
@@ -1,5 +1,5 @@
 import { User } from '@domain/community/user/user.entity';
-import { Module } from '@nestjs/common';
+import { Module, forwardRef } from '@nestjs/common';
```
techsmyth: remove forwardRef

src/services/infrastructure/entity-resolver/storage.bucket.resolver.service.ts
```diff
@@ -1,32 +1,45 @@
...
   ) {}
 
   public async getStorageBucketIdForProfile(
     profileID: string
-  ): Promise<string> {
+  ): Promise<string | never> {
     // First iterate over all the entity types that have storage spaces directly
     for (const entityName of Object.values(DirectStorageBucketEntityType)) {
-      const match = await this.getDirectStorageBucketForProfile(
+      const match = await getDirectStorageBucketForProfile(
```
techsmyth: what is the rational for moving the functions out of the class?

techsmyth (changes requested):
Nice to have this in place. 

Key comment is that the mutation does not belong in the module where it is now - this must move to the platform/admin module.

src/services/infrastructure/entity-resolver/storage.bucket.resolver.service.ts
valentinyanakiev: The rational was that there were two ways to go about this one-off (my understanding) migration:

1. Quick and dirty code, develop as quickly as possible, test, deploy, execute, and revert the unnecessary code. No additional authorization etc. needed, as this will be executed once, and executing it twice won't actually change anything. 
2. Spend a little more time and polish it up - easier to review, better code, looks better. Takes more time.

My understanding provided the feedback around InnovationSpaces, is we prefer `1` over `2`. That was my reasoning. Regardless, I have already adjusted the B branch and merged it in this one, which reverts to `2`, and pretty much addresses all comments.

techsmyth: I know that was the steer, and we do need to get this out - but it was fairly quick for me to move the code so that it was aat least isolated.

techsmyth (approved):
Looking forward to seeing this in!

PR #2799, opened 2023-05-09
Innovation Hub domain model + other

hero101:
- domain model finished
- service 'create' method + authorization
- bootstrap for both 'demo' and 'uwv' hubs
- graphql read api


src/domain/innovation-hub/innovation.hub.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+    length: SUBDOMAIN_LENGTH,
+    unique: true,
+  })
+  subdomain!: string;
+
+  @Column()
+  type!: InnovationHubType;
+
+  @Column('varchar', {
+    length: 255,
```
valentinyanakiev, line 25: `255` length is excessive for `active|archived|demo`. It should be `16` or `32` max.

src/domain/innovation-hub/innovation.hub.service.ts
```diff
@@ -0,0 +1,154 @@
...
+        `Space list filter for Innovation Hub with id: '${hubId}' not found!`,
+        LogContext.INNOVATION_HUB
+      );
+    }
+
+    return hub.hubListFilter;
+  }
+}
+
+const tryValidateCreateInput = ({
```
valentinyanakiev: You are validating here, not trying to validate. Validation fails in certain cases, then you throw an exception, which is part of the validation process. `validateCreateInput` would be a better name for the method.

valentinyanakiev (changes requested):
Some minor comments

PR #2808, opened 2023-05-11
Two Innovation Hubs e2e

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


alkemio.yml
```diff
@@ -74,6 +74,9 @@ security:
     # Cors allowed headers.
     allowed_headers: ${CORS_ALLOWED_HEADERS}:Origin,X-Requested-With,Content-Type,Accept,Authorization
 
+innovation_hub:
+  header: ${INNOVATION_HUB_HEADER}:myhost
```
valentinyanakiev: change the default to the real header used

src/domain/challenge/hub/hub.resolver.queries.ts
```diff
@@ -15,8 +19,15 @@ export class HubResolverQueries {
...
-    return await this.hubService.getHubs(args);
+  async hubs(
+    @InnovationHubDecorator() innovationHub: InnovationHub | undefined,
+    @Args({ nullable: true }) args: HubsQueryArgs
+  ): Promise<IHub[]> {
+    if (!innovationHub) {
+      return await this.hubService.getHubs(args);
+    }
+
+    return this.hubService.getHubsForInnovationHub(innovationHub);
```
valentinyanakiev: maybe a more appropriate name would be `getSpacesForInnovationHub`. It is getting increasingly confusing.

src/domain/challenge/hub/hub.service.ts
```diff
@@ -286,6 +288,61 @@ export class HubService {
     return hub.visibility;
   }
 
+  public getHubsForInnovationHub({
```
valentinyanakiev:
```suggestion
  public getSpacesForInnovationHub({
```

valentinyanakiev: Don't change the rest of the methods yet but this one at least would make sense

valentinyanakiev (changes requested):
Comments inline

src/domain/challenge/hub/hub.resolver.queries.ts
hero101: I was hesitant to start using the new convention, but if you also believe so, I will start.

PR #2823, opened 2023-05-17
Fix that works (supposedly) with prod data

valentinyanakiev:


src/migrations/1683791393534-whiteboardsValueMigration.ts
```diff
@@ -57,6 +55,8 @@ export class whiteboardsValueMigration1683791393534
       } catch (error) {
         continue;
       }
+      if(!templateValue.elements) continue;
```
alstojanovic, line 59: We should probably have this check for the canvases as well. (Even tho it was already tested and it didn't break, I think we should add it)

alstojanovic (approved):

PR #2838, opened 2023-05-20
Invitation entity + logic

techsmyth:
adds Invitation entity to the domain model

Very similar to Application conceptually, with a different set of privileges and lifecycle

In place:
- entity / lifecycle / mutations / queries
- sample graphql samples
- rolesUser updated to allow querying the open invitations for the user

Also some renaming / tidy up on application entity usage to align. 

Note: pattern related to User link to Invitation is different than with Application. To review later whether to also make that change on Application to match what is in use here as it avoids a ManyToMany relationship in the ORM (but which may be useful for data integrity).

Client does not need changing except when it starts to use the new api (separate issue)


src/domain/community/application/application.resolver.mutations.ts
```diff
@@ -0,0 +1,38 @@
...
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args('deleteData') deleteData: DeleteApplicationInput
+  ): Promise<IApplication> {
+    const application = await this.applicationService.getApplicationOrFail(
+      deleteData.ID
+    );
+    await this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      application.authorization,
+      AuthorizationPrivilege.UPDATE,
```
valentinyanakiev:
```suggestion
      AuthorizationPrivilege.DELETE,
```

src/domain/community/community/community.entity.ts
```diff
@@ -60,6 +61,12 @@ export class Community
   })
   applications?: IApplication[];
```
valentinyanakiev: This surely is a mistake, right? We shouldn't have reference to GraphQL in our entities. Import should be updated as well...
```suggestion
  applications?: Application[];
```

src/domain/community/invitation/invitation.entity.ts
```diff
@@ -0,0 +1,28 @@
...
+    cascade: false,
+    onDelete: 'CASCADE',
+  })
+  community?: Community;
+
+  @Column('char', { length: 36, nullable: true })
+  invitedUser!: string;
+
+  @Column('char', { length: 36, nullable: true })
+  invitedBy!: string;
```
valentinyanakiev: This is confusing, I assume it is `createdBy`, the user creating this entity. It would be better to be consistent in that aspect.

src/domain/community/invitation/invitation.resolver.mutations.ts
```diff
@@ -0,0 +1,39 @@
...
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args('deleteData') deleteData: DeleteInvitationInput
+  ): Promise<IApplication> {
+    const application = await this.invitationService.getInvitationOrFail(
+      deleteData.ID
+    );
+    await this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      application.authorization,
+      AuthorizationPrivilege.UPDATE,
```
valentinyanakiev:
```suggestion
      AuthorizationPrivilege.DELETE,
```

src/domain/community/invitation/invitation.service.ts
```diff
@@ -0,0 +1,162 @@
...
+      );
+    return user;
+  }
+
+  async findExistingInvitations(
+    userID: string,
+    communityID: string
+  ): Promise<IInvitation[]> {
+    const existingInvitations = await this.invitationRepository
+      .createQueryBuilder('invitation')
```
valentinyanakiev: this is pretty straightforward query, can be done with `find` + options

src/domain/community/invitation/invitation.service.ts
```diff
@@ -0,0 +1,162 @@
...
+        communityID: communityID,
+      })
+      .getMany();
+    if (existingInvitations.length > 0) return existingInvitations;
+    return [];
+  }
+
+  async findInvitationsForUser(userID: string): Promise<IInvitation[]> {
+    const existingInvitations = await this.invitationRepository
+      .createQueryBuilder('invitation')
```
valentinyanakiev: this is pretty straightforward query, can be done with `find` + options

valentinyanakiev (changes requested):

valentinyanakiev:
Also, the corresponding NotificationEventTypes need to be published to NPM so this works.

src/domain/community/application/application.resolver.mutations.ts
techsmyth: This was actually a bug then in the old code as this mutation is just moved to a new class. Fixed now.

src/domain/community/community/community.entity.ts
techsmyth: Agreed, also legacy code that had a bug.

src/domain/community/invitation/invitation.entity.ts
techsmyth: Ok, will switch it to be createdBy

src/domain/community/invitation/invitation.resolver.mutations.ts
techsmyth: Ok

src/domain/community/invitation/invitation.service.ts
techsmyth: Indeed simpler

src/domain/community/invitation/invitation.service.ts
techsmyth: I did it for the other one, but did not see how to do a find on a child entry, if you know how please go ahead and update the code...

techsmyth:
@valentinyanakiev nice feedback. All resolved except for the optimization for the second find. Not critical that this goes in, if you know how to do it quickly then go for it - otherwise just leave it as the current code does work.

valentinyanakiev:
I have updated one of the examples and tested all mutations. All works fine, with the caveat that we expect invitations on `Space`. We need additional business logic for handling other `Journeys`, as they expect the user to be a member of the parent community. @techsmyth I will merge this once the build passes.

valentinyanakiev (approved):

PR #2926, opened 2023-06-17
external invitations

techsmyth:
New entity for InvitationExternal, with fields for email, welcome message, firstName, lastName

Mutation to add an external invitation to a community

Extended registration service to look up external invitations when the user profile is being created, and if any are present then:
- create a new full invitation
- mark the external invitation as having been used

An example flow:
- create an external invitation for the email address "neil@thesmyths.eu":

image

- that user logs in, and the existing external invitation is marked as accepted and a new invitation now exists! 
image


src/domain/community/invitation.external/invitation.external.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+  community!: Community;
+
+  @Column('varchar', { length: 255, nullable: false })
+  email = '';
+
+  @Column('varchar', { length: 255, nullable: true })
+  firstName = '';
+
+  @Column('varchar', { length: 255, nullable: true })
+  lastName = '';
```
alstojanovic, line 24: Is there a reason to have these this long?

techsmyth: Basically copying over what is there on the user entity, so that if we later wanted to re-use this for actually creating the user profile we could...

alstojanovic (approved):

PR #2929, opened 2023-06-20
Server renaming hub to space

techsmyth:
renames hub to space across code base, including samples

breaks dependency on main code base for one old migration


techsmyth:
@alstojanovic it is getting now an error on this query:

query {
  me {
      id
    nameID
	preferences
    {
      definition
      {
        group
        displayName
        description
        valueType
        type
      }
      value
    }
  
  }
}

and it seems to be the "type" field above that is causing the issue - remove it + the query works fine. 

This is somehow introduced with this migration...

alstojanovic (dismissed):

alstojanovic (dismissed):

alstojanovic (approved):

PR #2943, opened 2023-06-21
Removes duplicate storing of the parent profile displayName from Community

techsmyth:
there was an issue whereby the displayName for a community was still not being updated, but logically this is the same data stored now in two locations. 

This PR removes the displayName from community (where it does not make sense), essentially addressing some technical debt. 

Note: external api has not changed, so client does not need updating for this PR. A separate issue #4287 has been raised for looking at usage of this field on the clent + potentially removing its usage there, which is also not logical.


alstojanovic (approved):

alstojanovic:
Please don't merge until we fix some migration issue on DEV env.

PR #2948, opened 2023-06-25
CommunityPolicy to also cover admins, hosts

techsmyth:
Huge simplification of logic around Community roles, removes 10 mutations, and associated DTOs + code

Extends roles in community to include Admin + Host.
Host is now separate from Lead on Space community, + not used on Challenge / Opportunity.

Includes:
- migration up + down
- updated graphql samples
- hugely simplified set of logic paths related to roles in a community

Community field resolvers for users / organizations made generic, removes 3 fields and adds 2 - allowing for retrieving of all roles. 

Finally, all assignment of roles to users / orgs return the user / org - not the Community. 

Client has also been updated.


Comoque1:
When I try to start the server, I get this error:

techsmyth:
@Comoque1 some code that got merged in still was using the old api, now fixed...

Comoque1:
Now the space fails with relation to profile:
image

Comoque1:
Creating a space throws an  error:
image

After a space is created, the error from the above comment is hit.


The newly created spaces have missing data:
image

techsmyth:
@Comoque1 issue is clear, the underlying objects were getting updated before converting to strings for putting into the community policy fields. Have pushed an update, will test it later but if you already want to try the update out it should I hope be better now...

Comoque1:
> @Comoque1 issue is clear, the underlying objects were getting updated before converting to strings for putting into the community policy fields. Have pushed an update, will test it later but if you already want to try the update out it should I hope be better now...

@techsmyth, the issue is still reproducible:
image

Comoque1:
In addition, when I comment the peace, where creator of journey is assigned as member / admin, the space can be created successfully.
image

If I assign a user via api, it is successful
image


From the client - the assignment fails with this error:
image

Comoque1:
The creation of Space is successful, with latest change.
The context and collaboration is not set for the newly, created spaces.
image


In addition, a Space cannot be deleted:
image
image

Comoque1:
With this changes space leads don't receive messages, when contacting hosts

image
image

techsmyth:
@Comoque1 good catch - please raise the updating of notifications to go to leads vs hosts as a separate issue. Near term they will be teh same so we have time to address this.

valentinyanakiev (approved):

PR #2951, opened 2023-06-26
Some Ids migrated to char(36) and UUID strategy improvement

hero101:
This PR includes
- innovation_hub id, authorizationId, profileId migrated to char(36)
- the TypeORM data source driver to handle UUID generation strategy [correctly](https://github.com/typeorm/typeorm/issues/7228) with a workaround

I seem to have found the core issue with the migration generation not working correctly - generating a couple of hundred of "alter" statements.
At some point in time, TypeORM has changed the default behavior of some primary column generation strategies and the migration module is trying to migrate that to the newest strategy.
I have made an adjustment to handle at least our case for the UUID strategy - now the primary IDs are going to be correctly generated as char(36) instead of varchar(36).
**Unfortunately, that won't fix the generation, but at least more of the statement will be correct now.**


valentinyanakiev (approved):
Code looks fine, tested locally - up and down work fine.

PR #2954, opened 2023-06-27
Upload files to storage bucket directly, not through references

ccanos:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
+++ b/src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
@@ -192,7 +192,7 @@ export class StorageBucketResolverMutations {
     this.authorizationService.grantAccessOrFail(
       agentInfo,
       storageBucket.authorization,
-      AuthorizationPrivilege.READ, //FILE_UPLOAD,
+      AuthorizationPrivilege.FILE_UPLOAD,
       `create document on storage: ${storageBucket.id}`
     );
 
--- a/src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
+++ b/src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
@@ -24,6 +24,7 @@ import { Visual } from '@domain/common/visual';
 import { EntityNotInitializedException } from '@common/exceptions';
 import { LogContext } from '@common/enums';
 import { StorageBucketUploadFileInput } from './dto/storage.bucket.dto.upload.file';
+import { StorageBucketUploadFileOnReferenceInput } from './dto/storage.bucket.dto.upload.file.on.reference';
 
 @Resolver()
 export class StorageBucketResolverMutations {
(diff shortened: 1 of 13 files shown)


src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
```diff
@@ -170,4 +171,49 @@ export class StorageBucketResolverMutations {
     };
     return await this.referenceService.updateReference(updateData);
   }
+
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => String, {
+    description:
+      'Create a new Document on the Storage and return the value as part of the returned Reference.',
```
hero101: fix the description

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
```diff
@@ -170,4 +171,49 @@ export class StorageBucketResolverMutations {
...
+
+    const documentAuthorized =
+      await this.documentAuthorizationService.applyAuthorizationPolicy(
+        document,
+        storageBucket.authorization
+      );
+
+    const uri =
+      this.documentService.getPubliclyAccessibleURL(documentAuthorized);
+    return uri;
```
hero101: return it inline

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
```diff
@@ -170,4 +171,49 @@ export class StorageBucketResolverMutations {
...
+  ): Promise<string> {
+    const storageBucket =
+      await this.storageBucketService.getStorageBucketOrFail(
+        uploadData.storageBucketId
+      );
+
+    this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      storageBucket.authorization,
+      AuthorizationPrivilege.READ, //FILE_UPLOAD,
```
hero101: This has to be addressed on policy level I think with the proper privilege.

hero101 (changes requested):
Great overall. Tested locally.
Minor adjustments are needed.

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
hero101: @techsmyth We need to discuss how is this going to be achieved. We need an extra credential rule to apply FILE_UPLOAD on the bucket if the user has a membership to the parent journey.

techsmyth: Agreed, but needs a bit of discussion about how to assign that privilege as do now want that the privilege shows up everywhere in the tree as it is very specific to the storage bucket entity. Suggest we chat it through...

One option would be to assign it in the following location(so line 422):
image

+ then raise another issue for fixing that the FILE_UPLOAD is applied too widely. So as to not hold this issue up.

hero101 (approved):
Tested locally, with the latest rule changes.
Auth reset is required.

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
hero101: Resolved with https://github.com/alkem-io/server/pull/2956

PR #3015, opened 2023-07-11
Migrated old callout tags in a tagset to be the same format as new tags in a new callout

valentinyanakiev:
Migrated old callout tags in a tagset e.g. KNOWLEDGE to be in the format KNOWLEDGE in the DB and not ["\"KNOWLEDGE\""]


src/migrations/1689087486575-tagset-names.ts
```diff
@@ -0,0 +1,73 @@
...
+      await queryRunner.query(`UPDATE tagset SET tags = ? WHERE id = ?`, [
+        tagset.defaultSelectedValue,
+        tagset.id,
+      ]);
+    }
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    console.warn(
+      "Migration 'tagsetNames1689087486575' is not revertible. Please make sure you have a backup of your data before running this migration."
```
hero101, line 61: You will receive this message a little bit late, after you have applied the UP.

src/migrations/1689087486575-tagset-names.ts
```diff
@@ -0,0 +1,73 @@
...
+      INNER JOIN tagset_template tt
+      ON t.tagsetTemplateId = tt.id
+      WHERE t.name = 'flow-state'
+    `);
+
+    // Iterate through each record
+    for (const tagset of flowStateTagsets) {
+      // Update the 'tags' field with new value
+      await queryRunner.query(`UPDATE tagset SET tags = ? WHERE id = ?`, [
+        tagset.defaultSelectedValue,
```
hero101, line 53: I hope this as a single value isn't still written as an array.

hero101 (approved):

PR #3024, opened 2023-07-13
updated mutation for innovationFlow to specify optionally the set of visibleStates

techsmyth:
The mutation to updateInnovationFlow is extended to allow specifying the set of visibleStates.

Services then updates the associated tagsetTemplate after validating all states are included in those set from the Lifecycle.


valentinyanakiev (changes requested):
I probably don't understand something, but I don't believe this is sufficient, nor that it will work. I keep seeing the `innovationFlow.profile.Id` being provided to the `UpdateProfileSelectTagsetInput`, but this is not the profileID of the tagsets filtering the `Callouts`. Each `Callout` has a classification tagset with name `flow-state` and this tagset has profileID the profileID of the `Callout`. So to update what is actually driving the filtering, we need to provide the `Callout` profileID. Essentially, we need to loop through all `Callouts` for a specified `Collaboration` linked to a `Challenge` or `Opportunity`. We change the `InnovationFlow` of that `Challenge` or `Opportunity`, we need to loop through all `flow-state` tagsets in `Callouts`. I don't see that here. Or I am missing something big time.

techsmyth:
I think it is the relationship between the tagset and the tagset-template that is not being considered. 

The allowedValues on a tagset is a calculated field, with the result being the allowedValues field on the TagsetTemplate. 

Both the flow-state tagset on the InnovationFlow and the flow-state tagset on each Callout all point back to **the same** tagset-template entity.

So updating the single tagset-template also updates the allowed values on all the tagsets on the callouts.

So I think the above should work...

valentinyanakiev:
> I think it is the relationship between the tagset and the tagset-template that is not being considered.
> 
> The allowedValues on a tagset is a calculated field, with the result being the allowedValues field on the TagsetTemplate.
> 
> Both the flow-state tagset on the InnovationFlow and the flow-state tagset on each Callout all point back to **the same** tagset-template entity.
> 
> So updating the single tagset-template also updates the allowed values on all the tagsets on the callouts.
> 
> So I think the above should work...

While what you are saying is true, that will work for creating `new` entities. It won't updated `existing` ones. Updating the `tagsetTemplate` value goes through `updateSelectTagset` and while it indeed updates the `tagsetTemplate`, there is nothing for updating the `tagsets` `templated` by it..

valentinyanakiev (approved):

PR #3029, opened 2023-07-18
Fix

valentinyanakiev:


Comoque1:
Updating challenge innovation flow state, updates the state of its opportunity (looks like the instead of opportunityId is queried challengeId)

image

Comoque1 (approved):

PR #3054, opened 2023-07-27
ChatGuidance PoC

valentinyanakiev:
To use this:

- update .env.docker OPENAI_API_KEY=... your API key (contact me for details). This key should not be pushed...
- run `npm run start:services`

Enjoy the API. Works with `genai-poc` branch on the client.

Note: The implementation doesn't have authorization, preferences etc wired up yet - this will be done separately. Also, there is room for improvement in the typings + return types - in some places boolean may be more than sufficient... Up for discussion.


src/services/adapters/chat-guidance-adapter/chat.guidance.adapter.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { HttpModule } from '@nestjs/axios';
+import { ChatGuidanceAdapter } from './chat.guidance.adapter';
+
+@Module({
+  imports: [HttpModule],
```
techsmyth: This is not used I think

src/services/adapters/chat-guidance-adapter/chat.guidance.event.type.ts
```diff
@@ -0,0 +1,5 @@
+export enum ChatGuidanceEventType {
```
techsmyth: All enums under enum folder?

src/services/api/chat-guidance/dto/chat.guidance.query.result.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+import { IChatGuidanceResult } from './chat.guidance.result.dto';
+
+@ObjectType('ChatGuidanceQueryResult')
+export abstract class IChatGuidanceQueryResult extends IChatGuidanceResult {
```
techsmyth, line 5: I do not understand the design with the two DTO's. Not sure why one inherits from the other, and the base one is used in various places where it does not make sense as far as I can see.

src/services/api/chat-guidance/chat.guidance.resolver.mutations.ts
```diff
@@ -0,0 +1,37 @@
...
+  ): Promise<IChatGuidanceResult | undefined> {
+    return this.chatGuidanceService.resetUserHistory(agentInfo);
+  }
+
+  @UseGuards(GraphqlGuard)
+  @Mutation(() => IChatGuidanceResult, {
+    description: 'Resets the interaction with the chat engine.',
+  })
+  @Profiling.api
+  async ingest(): Promise<IChatGuidanceResult | undefined> {
```
techsmyth, line 34: api needs to be thought through, for more services etc + so that mutations are clearly associated with this service.

techsmyth: Plus this is really admin functionality, so need to think how to isolate it.

techsmyth (approved):
Really super to have this working!

Lets get this down to the poc branch.

Key points to consider:
- name of the repo
- api / dtos exposed
- graphql samples added
- speed

Also need to think through how to present the answers, as I was seeing text coming back that did not make a lot of sense, even after ingesting. 

Plus correlation of req / res.

So plenty to do - but a huge step forward...

src/services/api/chat-guidance/dto/chat.guidance.query.result.dto.ts
valentinyanakiev: I left it like that because the genai api returns the result as such - with payload { result: ... } and the query returns a superset of that. It is fit-for-purpose, at least between the gen ai API - rabbitMQ - Alkemio Server. I doubt it is needed on Alkemio API level, hence the stories raised - but merits a discussion.

src/services/adapters/chat-guidance-adapter/chat.guidance.event.type.ts
valentinyanakiev: It doesn't make sense to be in shared, it is used and will be used in only one place, so moved it there and made it private.

src/services/adapters/chat-guidance-adapter/chat.guidance.adapter.module.ts
valentinyanakiev: Agreed and removed.

src/services/api/chat-guidance/chat.guidance.resolver.mutations.ts
valentinyanakiev: I agree about that, and already mentioned it - we can discuss how to split it.

PR #3071, opened 2023-07-29
allow external users to be invited to multiple space communities, but not duplicated for the same community

techsmyth:
So a user cannot be invited twice to the same community but can be invited to multiple communities
Note: also removed some profiling decorators as they are not needed. - can ignore that change.


src/domain/community/community/community.service.ts
```diff
@@ -949,15 +955,17 @@ export class CommunityService {
...
         email
       );
-    if (existingExternalInvitation.length > 0) {
-      throw new InvalidStateTransitionException(
-        `An invitation with the provided email address already exists: ${email}`,
-        LogContext.COMMUNITY
-      );
+    for (const existingExternalInvitation of existingExternalInvitations) {
+      if (existingExternalInvitation.community.id === communityID) {
+        throw new InvalidStateTransitionException(
```
valentinyanakiev: I know this is was there already, but this exception type is extremely misleading and 100% incorrect.

valentinyanakiev (changes requested):

Comoque1:
External invitation tests passed.
image

src/domain/community/community/community.service.ts
techsmyth: Created new CommunityMembershipException and used that in all cases to replace the InvalidStateTransition in this file

valentinyanakiev (approved):

PR #3072, opened 2023-07-29
added check on size of file being uploaded

techsmyth:
Note: issue can only be tested via the api, as the client does prevent uploads larger than the allowed size - so this is primarily a security / stability issue as it would not impact users.


src/domain/storage/storage-bucket/storage.bucket.service.ts
```diff
@@ -300,10 +296,24 @@ export class StorageBucketService {
   private validateMimeTypes(
     storageBucket: IStorageBucket,
     mimeType: string
-  ): boolean {
+  ): void {
```
valentinyanakiev: I would not do that. `Validate` in software development is used as a method to validate the correctness of a condition and return a `boolean` value. 

This altered implementation is a hybrid between `validate` and `tryGet...`. It is unclean, confusing, and not fit-for-purpose in my opinion. Also, if it were to stay, the actual return type is `void | never`, as an exception is thrown inside the method under certain conditions.

valentinyanakiev (changes requested):

src/domain/storage/storage-bucket/storage.bucket.service.ts
valentinyanakiev: It is consistent with the rest at least, so I will just merge it...

valentinyanakiev (approved):

PR #3076, opened 2023-07-29
Fix number of space leads to be two

techsmyth:
The community policies on space had previously a limit of 1 for migrated spaces, but this should be two.


valentinyanakiev (approved):
Tested locally, worked fine for me.

PR #3085, opened 2023-07-31
added support for beta testers; added security around using guidance …

techsmyth:
Added support for granting users the credential as a beta tester
Added platform flag for disabling guidance engine
Added new privilege related to accessing guidance engine


valentinyanakiev (approved):
Works for me, also added the client logic to make this work.

PR #3088, opened 2023-08-01
Updated me query + applications / invitations via field resolvers on rolesUser

techsmyth:
me query updated to return the same results for invitations / applications as in rolesUser. The full entity is not useful as do not know the community for it.

added field resolver for the invitations / applications fields on rolesUser to only allow it for platform admins. Only a significant optimization

### DIFF
--- a/src/services/api/me/me.resolver.fields.ts
+++ b/src/services/api/me/me.resolver.fields.ts
@@ -51,7 +51,7 @@ export class MeResolverFields {
     })
     states: string[]
   ): Promise<InvitationForRoleResult[]> {
-    return await this.meService.getUserInvitations(agentInfo.userID, states);
+    return this.meService.getUserInvitations(agentInfo.userID, states);
   }
 
   @UseGuards(GraphqlGuard)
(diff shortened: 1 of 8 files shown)


src/services/api/me/me.resolver.fields.ts
```diff
@@ -50,16 +50,16 @@ export class MeResolverFields {
       description: 'The state names you want to filter on',
     })
     states: string[]
-  ): Promise<IInvitation[]> {
-    return this.meService.getUserInvitations(agentInfo.userID, states);
+  ): Promise<InvitationForRoleResult[]> {
+    return await this.meService.getUserInvitations(agentInfo.userID, states);
```
me-andre: It's a good convention to always `await`, but just FYI - from the point of JS runtime it's a no-op:
`return await` means unwrap a Promise, get it's resolved value, wrap into a Promise and then return.
However, it starts to matter once `return await` gets into `try catch`: without `await` it can only catch sync exceptions.
Feel free to dismiss this comment.

me-andre (dismissed):

src/services/api/me/me.resolver.fields.ts
me-andre: Also see https://eslint.org/docs/latest/rules/no-return-await why `return await` may actually be better.

me-andre (approved):

PR #3102, opened 2023-08-03
allow access to lifecycle state without read access

techsmyth:


Comoque1 (approved):

Comoque1:
Tested locally and works fine.

PR #3107, opened 2023-08-03
extended users query to take an optional paramter IDs, removed usersByIDs query

techsmyth:


src/domain/community/user/user.service.ts
```diff
@@ -574,6 +566,10 @@ export class UserService {
           credentialsFilter: credentialsFilter,
         })
         .getMany();
+    } else if (args.ids) {
+      users = await this.userRepository.find({
```
valentinyanakiev: This logic doesn't make sense. If you have credentials filter and IDs this won't correct return data set, but a super set of it.

valentinyanakiev (changes requested):

valentinyanakiev (approved):

PR #3110, opened 2023-08-04
Added logic to check for templates already there with the same name

techsmyth:
Also added check for same whiteboard template with same display name


valentinyanakiev (changes requested):
This is very odd and something I have deliberately avoided - making decisions based on displayName. Literally one space in the name and this won't work. I don't believe this should go in.

valentinyanakiev (approved):

PR #3112, opened 2023-08-04
move of timeline to be under collaboration

techsmyth:
To do: 
- revert migration is failing when creating the FK:
20:23:24	ALTER TABLE `space` ADD CONSTRAINT `FK_3005ed9ce3f57c250c59d6d5065` FOREIGN KEY (`timelineId`) REFERENCES `timeline`(`id`) ON DELETE SET NULL ON UPDATE NO ACTION	Error Code: 1452. Cannot add or update a child row: a foreign key constraint fails (`alkemio`.`#sql-alter-1-a6`, CONSTRAINT `FK_3005ed9ce3f57c250c59d6d5065` FOREIGN KEY (`timelineId`) REFERENCES `timeline` (`id`) ON DELETE SET NULL ON UPDATE NO ACTION)	0.078 sec


src/services/infrastructure/entity-resolver/timeline.resolver.service.ts
```diff
@@ -70,9 +68,11 @@ export class TimelineResolverService {
   ): Promise<string | undefined> {
     const result = await this.entityManager.findOne(Space, {
```
valentinyanakiev: This is dangerous. We can't imply that a calendarID is for a Space anymore. There should be, at the very least, defensive logic against it.

valentinyanakiev (changes requested):
Haven't tested + fixed down. I believe there are potential problems with the activity logging. Also, worth mentioning, it may be worth it taking new activity entries for `Challenge`, `Opportunity` as part of this PR. The lookup queries are missing altogether for those Journeys.

valentinyanakiev:
> To do:
> 
> * revert migration is failing when creating the FK:
>   20:23:24	ALTER TABLE `space` ADD CONSTRAINT `FK_3005ed9ce3f57c250c59d6d5065` FOREIGN KEY (`timelineId`) REFERENCES `timeline`(`id`) ON DELETE SET NULL ON UPDATE NO ACTION	Error Code: 1452. Cannot add or update a child row: a foreign key constraint fails (`alkemio`.`#sql-alter-1-a6`, CONSTRAINT `FK_3005ed9ce3f57c250c59d6d5065` FOREIGN KEY (`timelineId`) REFERENCES `timeline` (`id`) ON DELETE SET NULL ON UPDATE NO ACTION)	0.078 sec

that seems like corrupt data. Tested with production dataset, multiple ups and downs worked fine. Also, the migration seems perfectly fine to me.

valentinyanakiev (approved):

PR #3120, opened 2023-08-09
Fix callout sorting

ccanos:
Goes together with https://github.com/alkem-io/client-web/pull/4520

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


techsmyth (commented):
I think starting a zero is dangerous. It should at the least start at the minimum value of the provided callouts?

src/domain/collaboration/collaboration/collaboration.service.ts
```diff
@@ -584,31 +589,24 @@ export class CollaborationService {
...
-      }
       calloutsInOrder.push(callout);
-    }
-
-    // Now update all the callouts sort order
-    for (const callout of calloutsInOrder) {
-      callout.sortOrder = minCalloutSortOrder;
-      minCalloutSortOrder += 2;
+      callout.sortOrder = index;
       await this.calloutService.save(callout);
```
me-andre: The original algo had 2 cycles for computing the order and saving. That meant if an exception is thrown by the 1st one, no partial update will be performed.

me-andre: One "cheap" optimization could be not push the callout to the array of callouts to be updated if the `sortOrder` didn't change.

src/domain/collaboration/collaboration/collaboration.service.ts
```diff
@@ -584,31 +589,24 @@ export class CollaborationService {
 
     // Get the callouts specified
     const calloutsInOrder: ICallout[] = [];
-    let minCalloutSortOrder = 10000;
+    let index = 0;
     for (const calloutID of sortOrderData.calloutIDs) {
+      index++;
       let callout;
       if (calloutID.length === UUID_LENGTH)
```
me-andre: By switching from performing `find` to having dictionaries such as `calloutsByID = _.keyBy(allCallouts, 'id'); calloutsByNameID = _.keyBy(allCallouts, 'nameID')`, the complexity can be lowered ~ from On^2 to On and the code will look simpler as well.

me-andre: This can also lower a chance of missing a Callout which nameID has a length of UUID.

valentinyanakiev (approved):
Tested locally, worked very well for me.

PR #3126, opened 2023-08-11
remove field decorators on DTO that were conflicting with actual field resolvers

techsmyth:


Comoque1:
Tested locally and works fine.

alstojanovic (approved):

PR #3153, opened 2023-08-22
Distributed Authorization Reset via RMQ Queue + Microservice

valentinyanakiev:
I have:
  - refactored the code, adjusted it to follow the microservices design of the application
  - split the `service / controller` in two modules, `subscriber / publisher`
  - added respective resets for all entities
  - replaced the `authorizationPolicyResetAll` to use the queue
  - changed the structure a bit

The circular dependency is fixed now. I have tested locally and seems to be working but I can't grasp the benefits that way... The real lacmus test is on prod. This should be reviewable now.


hero101:
You can use the sample code below to implement a retry logic.
There is a 70% chance that the message will fail and will be retried with same probability `MAX_RETRIES` amount of times.

```typescript
const MAX_RETRIES = 5;
const RETRY_HEADER = 'x-retry-count';

  @EventPattern('test-reset', Transport.RMQ)
  public testReset(
    @Payload() payload: { id: string },
    @Ctx() context: RmqContext
  ) {
    const channel: Channel = context.getChannelRef();
    const originalMsg = context.getMessage() as Message;

    const retryCount = originalMsg.properties.headers[RETRY_HEADER] ?? 0;

    try {
      if (Math.random() < 0.7) {
        throw new Error('Processing failed');
      }

      console.log(
        'Processing succeeded',
        retryCount
          ? `after ${retryCount} retr${retryCount > 1 ? 'ies' : 'y'}`
          : ''
      );

      channel.ack(originalMsg); // Acknowledge the message
    } catch (e) {
      if (retryCount >= MAX_RETRIES) {
        console.error('Max retries reached. Rejecting message');
        channel.reject(originalMsg, false); // Reject and don't requeue
      } else {
        console.warn(
          `Processing failed. Retrying (${retryCount + 1}/${MAX_RETRIES})`
        );
        channel.publish('', 'auth-reset', originalMsg.content, {
          headers: { [RETRY_HEADER]: retryCount + 1 },
          persistent: true, // Make the message durable
        });
        channel.ack(originalMsg); // Acknowledge the original message
      }
    }
  }
```

src/platform/admin/authorization/admin.authorization.resolver.mutations.ts
```diff
@@ -234,68 +236,15 @@ export class AdminAuthorizationResolverMutations {
...
-      this.organizationService.getOrganizations({}),
-      this.userService.getUsers({}),
-    ]);
+    try {
+      await this.authResetService.publishAllSpaceReset();
+      await this.authResetService.publishAllOrganizationsReset();
+      await this.authResetService.publishAllUsersReset();
+      await this.authResetService.publishPlatformReset();
+    } catch (error) {
+      return false;
```
hero101, line 241: log the error, and give some meaningful response to the user

src/core/microservices/microservices.module.ts
```diff
@@ -90,13 +92,19 @@ const subscriptionFactoryProviders = subscriptionConfig.map(
       useFactory: chatGuidanceServiceFactory,
       inject: [WINSTON_MODULE_NEST_PROVIDER, ConfigService],
     },
+    {
```
hero101, line 95: I think this should be provided by the publisher instead. It is the only place it will be used.
The current implementation relies on the microservices module to provide it and this is a very hidden dependency.

hero101 (changes requested):

hero101 (approved):

PR #3173, opened 2023-08-30
removed nameID from update space input DTO

techsmyth:


alstojanovic (approved):
Tested locally updating a space.

PR #3177, opened 2023-08-30
refactor of roles service to allow separately querying journey roles from org roles

techsmyth:
Rational for this PR is that the userRoles query always was getting both all journey roles and all organization roles - regardless of what fields were being requested. And in the latest update from Andrew the code does request the different roles separately - which leads to an inefficient api.

The graphql api has not changed, this is a pure internal optimization that is important to allow client querying of organizational roles to be fast - currently it is still getting all the journey roles in the business logic. 

Design:
- The entity that is returned from userRoles / organizationRoles is pretty much empty, storing only information that is needed for the fields on it.
- All data is now returned as fields, allowing for retrieving of journey roles separate from organization roles. 

Important is that the logic path for organization roles is now fully independent of that of journey roles, as they are different (space visibilities etc makes no sense on organization roles).


src/services/api/roles/util/get.organization.roles.for.user.entity.data.ts
```diff
@@ -0,0 +1,29 @@
...
+          displayName: true,
+        },
+      },
+    } as FindManyOptions);
+  };
+  const [organizations] = await Promise.all([
+    fetchData(Organization, organizationIds),
+  ]);
+
+  return { organizations };
```
hero101: I see that you have copy-pasted the code from the other function, but this syntax is not required here.
It's a single promise and the function is providing a single value.
Distructuring is not needed.

hero101 (changes requested):
I see a lot of async/await in your code that are not required.
You can think of the `await` basically as execution. It should only be used if you need the value right now, in your next line, otherwise, you just pass around the promise.

Overall the implementations look good. An additional optimization leveraged by resolvers.

src/services/api/roles/roles.resolver.fields.ts
```diff
@@ -20,6 +22,27 @@ export class RolesResolverFields {
...
   ) {}
 
+  @UseGuards(GraphqlGuard)
+  @ResolveField('organizations', () => [RolesResultOrganization], {
+    description: 'Details of the roles the contributor has in Organizations',
+  })
+  public async organizations(
+    @Parent() roles: ContributorRoles
+  ): Promise<RolesResultOrganization[]> {
+    return await this.rolesService.getOrganizationRolesForUser(roles.id);
```
hero101: I don't know how the build passed but `getOrganizationRolesForUser` accepts `roles` here.

src/services/api/roles/util/get.organization.roles.for.user.entity.data.ts
techsmyth: This is actually fixed in the latest code...

src/services/api/roles/roles.resolver.fields.ts
techsmyth: Also fixed in the latest code...

techsmyth: correction, it was fixed - looks like something went wrong with a merge...will fix again.

src/services/api/roles/util/get.organization.roles.for.user.entity.data.ts
techsmyth: now fixed...

hero101 (dismissed):

hero101 (approved):

PR #3192, opened 2023-09-03
added modules to support callout templates

techsmyth:
New entities / modules:
* CalloutFraming
* CalloutResponseDefaults
* CalloutResponsePolicy
* CalloutTemplate

Additions
* lookup storage bucket for the profile on calloutFraming and calloutTemplate
* added lookup entry for calloutTemplate
* Migration up / down (not fully tested)
* DTOs used added to validation
* samples for working with the api

Light testing is done (create, read). Additional validation needed. 

image


src/domain/collaboration/callout-framing/callout.framing.entity.ts
```diff
@@ -0,0 +1,48 @@
...
+  @AfterUpdate()
+  @AfterLoad()
+  async decompressContent() {
+    if (this.content && this.content !== '') {
+      this.content = await decompressText(this.content);
+    }
+  }
+
+  @Column('longtext', { nullable: false })
+  content?: string;
```
hero101: The typeorm definition says non-nullable, but the class field is nullable.

src/domain/collaboration/callout-framing/callout.framing.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+
+@ObjectType('CalloutFraming')
+export abstract class ICalloutFraming extends IAuthorizable {
+  profile!: IProfile;
+
+  @Field(() => JSON, {
+    nullable: true,
+    description: 'The whiteboard template content for this Callout Framing.',
+  })
+  content?: string;
```
hero101: I think it's a good idea to sync this field with the entity definition

src/domain/collaboration/callout-framing/callout.framing.service.ts
```diff
@@ -0,0 +1,116 @@
...
+    private calloutFramingRepository: Repository<CalloutFraming>
+  ) {}
+
+  async createCalloutFraming(
+    calloutFramingData: CreateCalloutFramingInput
+  ): Promise<ICalloutFraming> {
+    const calloutFraming: ICalloutFraming =
+      CalloutFraming.create(calloutFramingData);
+
+    calloutFraming.authorization = new AuthorizationPolicy();
```
hero101, line 36: what about profile

src/domain/collaboration/callout-response-defaults/callout.response.defaults.entity.ts
```diff
@@ -0,0 +1,40 @@
...
+  @AfterUpdate()
+  @AfterLoad()
+  async decompressContent() {
+    if (this.whiteboardContent && this.whiteboardContent !== '') {
+      this.whiteboardContent = await decompressText(this.whiteboardContent);
+    }
+  }
+
+  @Column('longtext', { nullable: false })
+  whiteboardContent?: string;
```
hero101: The entity definition is `nullable: false`

hero101: The entity definition is `nullable: false`

src/domain/collaboration/callout-response-defaults/callout.response.defaults.service.ts
```diff
@@ -0,0 +1,66 @@
...
+  async updateCalloutResponseDefaults(
+    calloutResponseDefaults: ICalloutResponseDefaults,
+    calloutResponseDefaultsData: UpdateCalloutResponseDefaultsInput
+  ): Promise<ICalloutResponseDefaults> {
+    if (calloutResponseDefaultsData.postDescription) {
+      calloutResponseDefaults.postDescription =
+        calloutResponseDefaultsData.postDescription;
+    }
+
+    return calloutResponseDefaults;
```
hero101: what about `whiteboardContent`?

hero101 (changes requested):
I have left some notes that we need to discuss and also general notes:
- passing a profile to the framing when creating a template does not feel right. Furthermore, you have to provide a display name to that profile, which feels even stranger.
- the responseDefaults input is mandatory but nothing in IS. I am guessing it needs to be also optional or some sort of validation to provide at least one or only one
- I believe the updating API can be improved. Updating the framing of a template requires it's ID, but it should be already known by the backend so it's not really required.

The create and delete work fine.

src/domain/collaboration/callout-framing/callout.framing.entity.ts
techsmyth: nullable should be true

src/domain/collaboration/callout-framing/callout.framing.service.ts
techsmyth: yikes

src/domain/collaboration/callout-framing/callout.framing.entity.ts
techsmyth: field to be called whiteboardContent

src/domain/collaboration/callout-response-defaults/callout.response.defaults.entity.ts
techsmyth: nullable: true

src/domain/collaboration/callout-response-defaults/callout.response.defaults.service.ts
techsmyth: good catch

techsmyth:
@hero101 please rename Response to be Contribution

hero101 (approved):

PR #3203, opened 2023-09-06
added type for whiteboard content

techsmyth:
Used on all DTOs, fields where whiteboard content is being handled.

### DIFF
--- a/src/domain/common/scalars/scalar.whiteboard.content.ts
+++ b/src/domain/common/scalars/scalar.whiteboard.content.ts
@@ -1,0 +1,46 @@
+import { LogContext } from '@common/enums';
+import { ValidationException } from '@common/exceptions';
+import { Scalar, CustomScalar } from '@nestjs/graphql';
+import { Kind, ValueNode } from 'graphql';
+
+const WHITEBOARD_CONTENT_LENGTH = 8388608;
+
+@Scalar('WhiteboardContent')
+export class WhiteboardContent implements CustomScalar<string, string> {
+  description = 'Content of a Whiteboard, as JSON.';
+
+  parseValue(value: unknown): string {
+    return this.validate(value);
+  }
+
+  serialize(value: any): string {
+    return value; // value sent to the client
+  }
+
+  parseLiteral(ast: ValueNode): string {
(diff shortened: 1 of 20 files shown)


src/domain/common/scalars/scalar.whiteboard.content.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  parseLiteral(ast: ValueNode): string {
+    if (ast.kind === Kind.STRING) {
+      return this.validate(ast.value);
+    }
+    return '';
+  }
+
+  validate = (value: any) => {
```
me-andre: As far as I see, this can be a method

me-andre: (or maybe a class method even)

techsmyth: You mean to move the function to be at the class level? Not sure that adds any value as this method should not be public and is not re-used.

src/domain/common/scalars/scalar.whiteboard.content.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (typeof value !== 'string') {
+      throw new ValidationException(
+        `Whiteboard content is not string: ${value}`,
```
hero101: Please don't throw the whole value in the exception.

src/domain/common/scalars/scalar.whiteboard.content.ts
hero101: `validate` is already a public class method - It can be private, but I don't see any other issue than that.

hero101 (changes requested):
I also believe the new scalar must be included in the codegen.yml of the client.

src/domain/common/scalars/scalar.whiteboard.content.ts
me-andre: Sorry for not being clear, what I meant is that I would expect that method to be `static` (or extracted outside as a utility function). That is a matter of code style though.

hero101 (approved):

me-andre (approved):

PR #3221, opened 2023-09-21
Removed HOME_0 callout location. Moved recommendations to HOME_1

valentinyanakiev:
I have encountered multiple problems with this. I believe the requirement might be incorrect.

What is done:
- Removed HOME_0 ("HOME_TOP" / "Dashboard Recommendations" on the client)
- Replaced it with HOME_1 ("HOME_RIGHT" / "Dashboard Right" on the client)

This is one to one with AC.

Problems I have seen:
- There are default callouts created in Journeys in HOME_RIGHT that are not visualized (contributor-profiles)
- I believe HOME_RIGHT is actually the mid-right section on the page, not the top-right one. I don't believe we had anything for top-right section.

@techsmyth @SimoneZaza @larssondenise - the PR follows the AC but further changes may be needed as I feel the AC may not be what is actually needed.

### DIFF
--- a/src/common/enums/callout.display.location.ts
+++ b/src/common/enums/callout.display.location.ts
@@ -1,7 +1,6 @@
 import { registerEnumType } from '@nestjs/graphql';
 
 export enum CalloutDisplayLocation {
-  HOME_TOP = 'HOME_0',
   HOME_LEFT = 'HOME_1',
   HOME_RIGHT = 'HOME_2',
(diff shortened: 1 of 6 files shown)


me-andre (changes requested):
I think it's also important to go through all HOME_RIGHT callouts and ensure that the former HOME_TOP is still at the top (has the lowest sortOrder among new siblings).

me-andre (approved):

Issue #2196, opened 2022-09-28
Review mandatory spaceID


valentinyanakiev:
## Description
As a developer 
I want to have properties set based on business logic validation rules and not the IDE
So that there is lesser risk in writing broken logic.

## Acceptance criteria

- [ ] hubID mandatory

## Additional Context

Search for //toDo make mandatory https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/alkem-io/server/2196

techsmyth:
Fixed as part of storage buckets change PR

PR #3244, opened 2023-09-30
StorageBucket per profile

techsmyth:
Adding a storage bucket to every profile, so that authorization can be managed per Profile. 

StorageResolver service moved to a separate module
New module to allow simple user lookup without taking the whole User module imported. To avoid module circular dependencies. Updated other modules to use this to avoid too many dependencies. 
Fixed issue with spaceID not being mandatory on Challenge + Opportunity (#2196)
Authorization cascaded to storage bucket per profile
Migrations up + down
Migration sets the right storage bucket for **every** profile

Note: client should be only minimally affected (user groups)
Note: there are also some bugs addressed with this PR, such as challenge was not deleting the storage bucket associated with it.

**For addressing separately**: the reporting of storage used within a space / challenge, plus the way the admin side shows documents will need to be updated. 

For testing:
- check that all profiles have the storageBucketId column with a value
- besides the storage bucket functionality, please also test that groups are working

to do:
- test, test, test


src/domain/collaboration/callout/callout.service.ts
```diff
@@ -287,8 +307,8 @@ export class CalloutService {
     publishedTimestamp?: number
   ): Promise<ICallout> {
     if (publisherID) {
-      const publisher = await this.userService.getUserOrFail(publisherID);
-      callout.publishedBy = publisher.id;
+      const publisher = await this.userLookupService.getUserByUUID(publisherID);
```
valentinyanakiev, line 311: This is not better than before imo, quite the contrary.

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
```diff
@@ -51,7 +51,11 @@ export class StorageBucketResolverMutations {
     const visual = await this.visualService.getVisualOrFail(
       uploadData.visualID,
       {
-        relations: ['profile'],
+        relations: [
+          'profile',
+          'profile.storageBucket',
+          'profile.storageBucket.authorization',
```
valentinyanakiev, line 57: 'profile.storageBucket.authorization', should be redundant here

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
```diff
@@ -118,7 +118,11 @@ export class StorageBucketResolverMutations {
     const reference = await this.referenceService.getReferenceOrFail(
       uploadData.referenceID,
       {
-        relations: ['profile'],
+        relations: [
+          'profile',
+          'profile.storageBucket',
+          'profile.storageBucket.authorization',
```
valentinyanakiev, line 124: 'profile.storageBucket.authorization' should be redundant

src/services/infrastructure/user-lookup/user.lookup.service.ts
```diff
@@ -0,0 +1,35 @@
+import { FindOneOptions, Repository } from 'typeorm';
```
valentinyanakiev, line 1: I don't understand the need of this service.

valentinyanakiev (commented):
Some small changes that need discussion. Up-down-up works now with a minor change. I have added the profileType as well + migrated / tested the data migration, all seems fine. 
There is an issue with the storageBucketIds thou - prod data, 261 are left NULL after migration. I have tried to verify manually but I am not sure whether this isn't a data integrity issue...

src/services/infrastructure/user-lookup/user.lookup.service.ts
techsmyth: The alternative is to import the whole user module, which leads to a circular dependency. And in general we are importing the whole user module in multiple places where all we need is a simple service to find a user identity, the rest is not needed. So it greatly simplifies the module dependency tree - avoiding a circular dependency.

src/domain/storage/storage-bucket/storage.bucket.resolver.mutations.ts
techsmyth: possibly, but when loading like this it sometimes does not load it - so added it to be sure. You can try removing it - but no harm in having it there.

src/domain/collaboration/callout/callout.service.ts
techsmyth: For most of the code in the server when we look up a user, we KNOW that it is a UUID. The lookup by email / nameID should only be allowed on incoming queries - not for internal lookups. And the lookup via email / nameID has a small overhead - and more importantly is simply not as safe as when you know it is a UUID. And see comment later about not importing the whole User module. Also described in the pull request description...

techsmyth:
> Some small changes that need discussion. Up-down-up works now with a minor change. I have added the profileType as well + migrated / tested the data migration, all seems fine. There is an issue with the storageBucketIds thou - prod data, 261 are left NULL after migration. I have tried to verify manually but I am not sure whether this isn't a data integrity issue...

Almost certainly there are some integrity issues - the migration navigates the current tree, so any orphaned entity with a profile will not be reached. In fact these profiles are very good way of identifying which ones are orphaned - please raise another issue to have those profile UUIDs listed + that we then do some data clean up to find out where they are used, and navigate from there to find redundant entities. I suspect quite a number come from postTemplate as that used not to be deleted when a Callout was deleted...

valentinyanakiev (approved):

PR #3261, opened 2023-10-05
IPFS logging a warning when /ipfs/{CID} is being accessed

valentinyanakiev:


techsmyth (commented):
Changes make sense. Key question is how these log entries then make it over to elastic - how does then happen?

valentinyanakiev:
> Changes make sense. Key question is how these log entries then make it over to elastic - how does then happen?

They will be send to elastic as part of the logs. I have an alternative commit with this endpoint throwing an exception when being accessed. I can also create a separate exception / log context for it. Maybe that would be better?

Either way, not the greatest behavior when the endpoint is accessed in the browser, it is either whitescreen or the exception as output. Happy to demonstrate.

techsmyth:
@valentinyanakiev I would think that throwing an error would be best - we actually do not want to have anything coming through. It might be best to do something like what Svetlo did in the storage.access.controller whereby he catches another exception + then returns a nice http not found one: 
image

So then you throw an exception within a try-catch, and return the http not found one. 

The caught exception does also get logged as an error, and the user gets the right http response. 

Work?

valentinyanakiev:
> @valentinyanakiev I would think that throwing an error would be best - we actually do not want to have anything coming through. It might be best to do something like what Svetlo did in the storage.access.controller whereby he catches another exception + then returns a nice http not found one: image
> 
> So then you throw an exception within a try-catch, and return the http not found one.
> 
> The caught exception does also get logged as an error, and the user gets the right http response.
> 
> Work?

@techsmyth that doesn't make sense in our case. I don't need to catch anything as I have nothing executed in this method. I will just throw the exception... I will just create IpfsNotFoundException in the LogContext.IPFS and throw it if we hit this endpoint - that should work  OK. I have most of that already in a branch, just wanted to clarify the types. I have added those changes now - please have a look and merge if you think it's OK.

techsmyth (approved):
Looks good!

PR #3263, opened 2023-10-05
Callout refresh part 1: Update CalloutFraming and Callout to use CalloutFraming.

alstojanovic:
Todo:
- consider adding dataloaders into to load whiteboards on framing
- test, test, test...


alstojanovic:
CalloutFraming resolver fiealds are not fully working. (profile is working but whiteboard and whiteboardRt are not)
Migration is working but needs check if all constraints are set correctly.

PR #3273, opened 2023-10-06
Mutation to transfer files from whiteboard content to documents

hero101:
Moves the files from whiteboard, whiteboard_rt, whiteboard_template to the document api.

To be used when the client can support it.


src/app.module.ts
```diff
@@ -228,6 +229,7 @@ import { IpfsLogModule } from '@services/api-rest/ipfs-log/ipfs.log.module';
     ChatGuidanceModule,
     LookupModule,
     AuthResetSubscriberModule,
+    AdminWhiteboardModule,
```
valentinyanakiev:
```suggestion
    //AdminWhiteboardModule,
```

valentinyanakiev (changes requested):
Please disable it until the client is ready

valentinyanakiev (approved):

PR #3299, opened 2023-10-12
Split of StorageAggregation from StorageBucket

techsmyth:
New entity StorageAggregator to manage the aggregation of storage usage from child storage buckets.
- It also has a directStorage field to hold the documents that were previously directly under the storageBucket on the top level entity. The top level storage bucket is re-used for this.

Note: a clear separation in terms of child storage buckets and child storage aggregators - makes the UI I think also simpler. 

First pass logic to allow getting the displayname + url for child storage aggregators.

Logic to aggregate the storage usage (size) added in for storage aggregator. 

Migration up + down


valentinyanakiev (changes requested):
Seems to be mostly working for me. Checked the migration, Space / Challenge buckets / aggregators, all types of buckets. What stood out is that on User there is no  storage bucket field resolver on API level, and on Organization there is.  Also, Global Admin could not access organization.profile.storageBucket.size. Apart from that, overall seems to work ...

techsmyth:
@valentinyanakiev the fact there was a storage bucket field on org was an error, it should have been storageAggregator. Did not get picked up by typings as using a data loader. Also added the same field also now in for users. Otherwise the size on profile now seems to be working fine: 
image

valentinyanakiev (approved):

PR #3300, opened 2023-10-13
resolve the ipfs upload duplex issue by upgrading to specific version of ipfs-http-client

techsmyth:
Needs to be exactly the following version to avoid module loading issues:
https://github.com/ipfs/js-ipfs/issues/4139


valentinyanakiev:
I can't reproduce the issue but this works for me.

PR #3318, opened 2023-10-16
Conversion service refresh 

techsmyth:
Allows converting a challenge, with child opportunities, to be a space.

This service will need additional testing, suggest to take in refresh as is and fix from here. In light initial testing it is working.


valentinyanakiev (approved):
@techsmyth Merging this, as agreed, but it requires further work in the space of policies and their validation. I couldn't migrate a challenge to space as the lead organization limit was always out of bounds - that business logic needs updating.

techsmyth:
@valentinyanakiev you mean the error of having exactly one lead org before converting? That is deliberate - otherwise we can have spaces with no host org which is not allowed. So the challenge may need to be updated before converting which is fine.

PR #3331, opened 2023-10-18
Lifecycle init

alstojanovic:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/core/authorization/authorization.service.ts
```diff
@@ -86,6 +86,12 @@ export class AuthorizationService {
         'Authorization: no definition provided',
         LogContext.AUTH_POLICY
       );
+    if (authorization.credentialRules === '') {
```
valentinyanakiev, line 89: @techsmyth this requires a discussion. In my opinion, this is not future-proof, as mandating any type of authorization rules set for an entity is not guaranteed. We may have only verifiedCredentialRules for a specific entity, or just privilegeRules. I'd rather go with logging a warning / error but not stopping the flow here. Happy to discuss this.

techsmyth: @valentinyanakiev agreed that there could be no credential rules possibly - but that is different than having the field not initialized. And this exception is about flagging when an authorization policy is checked when it has not been properly initialized. We could make this more explicit but for now checking for an empty string will do it - and I think it should be an exception so that it is highly visible to developers. These errors are otherwise costing way too much time to track down - so my view is to leave as an exception for now and if we have later authorization policies without credential rules we adapt the logic then. At the moment every authorization policy has at least one rule...

valentinyanakiev: I see it now, it is an empty array when nothing is initialized, not an empty string. That would work.

valentinyanakiev (approved):

PR #3347, opened 2023-10-20
Fixed commenting on Posts

valentinyanakiev:
tested on all journey levels


hero101 (approved):
Looks good.Should cover all the cases with the Post

PR #3350, opened 2023-10-21
ORM relations loaded with typing

techsmyth:
With help of chatGPT took not long at all...should be safe enough.


valentinyanakiev (changes requested):
Unfortunately, you fell into the classic new GPT-developer trap, in my opinion - underestimating complexity, effort and risk and taking GPT's reponse for safe and correct. There are several issues with this PR:
- typings - {} is not the same as []. I don't see any signature changes. Also, a simple search of the obsolete type `FindOptionsRelationByString` still shows 27 hits and this needs to be replaced with FindOptionsRelations<TType>
- After replacing these types there are more errors that surface
- refactoring - after changing FindOptionsRelationByString with FindOptionsRelations<TType> all places loading relations will need to be refactored, which are quite likely in the hundreds, as merging an object of type X with Y is not the same as merging an object X with Z

With all that, I can also safely say that I was correct to urge caution with this task and that it is grossly underestimated. Fixing all this can easily take more than a day and I am not sure this is a day we want to spare this sprint.

valentinyanakiev:
I have opened a [PR](https://github.com/alkem-io/server/pull/3361) with the necessary changes and I have done smoke testing of it, seems to be working. Ideally the whole thing is merged to develop once we have the API tests running again.

valentinyanakiev (approved):

PR #3359, opened 2023-10-24
rolesUser to only show child challenge + opportunities for spaces with READ access

techsmyth:
Updated the rolesUser service to not show memberships of challenges + opportunities that are inside of Spaces that the user does not have access to. 

Note: there is a second case that we can consider, whereby there are opportunities inside of private challenges - but that should be picked up separately. This PR will address the bulk of the current production errors.


techsmyth:
@Comoque1 please check this fix - and if you think it useful please raise a separate issues for accessing opportunities in a private challenge.

valentinyanakiev (changes requested):
This doesn't work at all. A GA with memberships: 

image

Also, after the straight case is fixed and smoke testing passes, tests will need to be updated as well.

valentinyanakiev (approved):

PR #3364, opened 2023-10-25
Chat guidance simple answer voting

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/services/external/elasticsearch/guidance-reporter/guidance.reporter.service.ts
```diff
@@ -67,6 +67,34 @@ export class GuidanceReporterService {
     return;
   }
 
+  public async updateAnswerRelevance(answerId: string, relevance: boolean) {
```
valentinyanakiev, line 110: Having this in the `GuidanceReporterService` couples elastic with the relevance of the answer, domain wise. I believe updating the relevance should be in the `ChatGuidanceLogService`, and the concrete implementation could be on DB level, Elastic, or, ideally, both. Currently, you have chosen `Elastic` but that makes the whole DB approach half-working, (so not working, as it doesn't give the full picture). This merits a discussion at least.

valentinyanakiev (changes requested):

PR #3392, opened 2023-10-30
added type to callout template

techsmyth:


valentinyanakiev (changes requested):
I don't believe the type should be nullable in the database. All Callouts should have a type.
We also need to migrate the existing types.

alstojanovic:
This is a new field so there isn't any existing data to migrate. It says to set type to be POST for any existing callout templates.

techsmyth:
@valentinyanakiev as Aleks says this is a new field on CalloutTemplate, basically to bring it to the same set of information as is needed for Callout. Only point is whether nullable should be false, if you think that is better then please just update the migration + merge it in...

valentinyanakiev:
> This is a new field so there isn't any existing data to migrate. It says to set type to be POST for any existing callout templates.

I am redacting this as this is no longer relevant after our discussion...

alstojanovic:
ProfileType for CalloutTemplate should be CalloutTemplate. The type field on Profile infers the Profile's parent entity type and we are talking here about a different field. Callouts have type which is of type CalloutType and Profile has also a type field which is of type ProfileType.
For instance a Callout is of type POST_COLLECTION but its Profile type is CALLOUT.

The NULL/NOT NULL point stands I think and it should be corrected.

valentinyanakiev (approved):

PR #3393, opened 2023-10-31
MyJourneys based on latest activity

valentinyanakiev:
- field resolver under `me`
- all Journeys I have credentials for returned with `IBaseChallenge` type
- credentials sorted based on my latest contributions (`Activity`)
- optional parameter for filtering
- optional parameter for taking a limited amount of Journeys

If there are multiple different types of journeys with activity on them, this will return a mix of journeys, based on the latest contributions of the user.


src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -624,9 +624,19 @@ export class ChallengeService {
     return opportunity;
   }
 
-  async getChallenges(): Promise<Challenge[]> {
-    const challenges = await this.challengeRepository.find();
-    return challenges || [];
+  public async getChallenges(
+    options?: FindOneOptions<Challenge>
+  ): Promise<IChallenge[]> {
+    let challenges: IChallenge[] | undefined = undefined;
```
hero101: `const` and early exit instead of reassigning value to `challenges`

src/domain/challenge/space/space.service.ts
```diff
@@ -399,6 +402,7 @@ export class SpaceService {
     if (args && args.IDs)
       spaces = await this.spaceRepository.find({
         where: { id: In(args.IDs) },
```
hero101, line 422: you have to also include `options.where`, because the rest operator below will override everything

src/domain/collaboration/opportunity/opportunity.service.ts
```diff
@@ -257,6 +257,21 @@ export class OpportunityService {
     return opportunity;
   }
 
+  public async getOpportunities(
+    options?: FindOneOptions<Opportunity>
+  ): Promise<IOpportunity[]> {
+    let opportunities: IOpportunity[] | undefined = undefined;
```
hero101: same comments as for the challenge service

src/domain/challenge/challenge/challenge.service.ts
```diff
@@ -624,9 +624,19 @@ export class ChallengeService {
...
+    challenges = await this.challengeRepository.find({
+      ...options,
```
hero101: just pass options normally as a variable.

src/domain/challenge/base-challenge/journey.interface.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (journey.parentSpace) return IChallenge;
+    if (journey.challenge) return IOpportunity;
+
+    throw new RelationshipNotFoundException(
+      `Unable to determine Journey type for ${journey.id}`,
+      LogContext.CHALLENGES
+    );
+  },
+})
+export abstract class IJourney extends INameable {
```
hero101: isn't it better to extend `IBaseChallenge` instead - it's 1:1?

src/domain/challenge/challenge/challenge.interface.ts
```diff
@@ -1,10 +1,13 @@
...
 import { IInnovationFlow } from '../innovation-flow/innovation.flow.interface';
 import { IStorageAggregator } from '@domain/storage/storage-aggregator/storage.aggregator.interface';
-@ObjectType('Challenge')
-export abstract class IChallenge extends IBaseChallenge {
+import { IJourney } from '../base-challenge/journey.interface';
+import { IBaseChallenge } from '../base-challenge/base.challenge.interface';
+@ObjectType('Challenge', {
+  implements: () => [IJourney],
+})
+export abstract class IChallenge extends IBaseChallenge implements IJourney {
```
hero101, line 10: `IBaseChallenge` and `IJourney` are 100% the same. 
I think we should discuss the structure further.

src/services/api/me/me.resolver.fields.ts
```diff
@@ -98,4 +100,23 @@ export class MeResolverFields {
...
+
+  @UseGuards(GraphqlGuard)
+  @ResolveField(() => [MyJourneyResults], {
+    description: 'The Journeys I am contributing to',
+  })
+  public myJourneys(
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args({
+      name: 'limit',
+      type: () => Float,
```
hero101, line 113: It's javascript `Number` ?

src/services/api/me/me.service.ts
```diff
@@ -48,4 +58,82 @@ export class MeService {
...
+      relations: { collaboration: true, innovationFlow: true, challenge: true },
+    });
+
+    const myJourneyResults: MyJourneyResults[] = [];
+
+    // Process spaces, challenges, and opportunities together
+    const entitiesToProcess = [...spaces, ...challenges, ...opportunities];
+
+    for (const entity of entitiesToProcess) {
+      if (!entity.collaboration?.id) continue;
```
hero101: this sound like a serious issue. Might be good to raise `EntityNotInitializedException` or log it.

src/services/api/me/me.service.ts
```diff
@@ -48,4 +58,82 @@ export class MeService {
...
+    }
+
+    // Sort myJourneyResults by the newest activity's createdDate in descending order
+    myJourneyResults.sort((a, b) => {
+      const dateA = a.latestActivity?.createdDate || new Date(0);
+      const dateB = b.latestActivity?.createdDate || new Date(0);
+      return dateB.getTime() - dateA.getTime();
+    });
+
+    return myJourneyResults.slice(0, limit);
```
hero101, line 139: `activityLogService.activityLog` already has a `limit` argument.

hero101 (changes requested):

src/domain/challenge/base-challenge/journey.interface.ts
valentinyanakiev: /home/valentin/work/repos/alkemio/server/src/domain/challenge/challenge/challenge.interface.ts:9
export abstract class IChallenge extends IJourney {
                                         ^
TypeError: Class extends value undefined is not a constructor or null
    at Object.<anonymous> (/home/valentin/work/repos/alkemio/server/src/domain/challenge/challenge/challenge.interface.ts:9:42)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/home/valentin/work/repos/alkemio/server/src/domain/challenge/base-challenge/journey.interface.ts:4:1)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)

That's what I get in this scenario.

PR #3394, opened 2023-10-31
Activity feed

hero101:
- How do you understand what type of journey is this activity about?
- What is the `visibility` parameter passed to the Activity?
- The current approach reuses old activity code which has some weak points below
May return incorrect paginated results due to `convertRawActivityToResults` returning `undefined` entries due to errors in processing or missing data.
May return wrong `pageInfo` data OR less amount of items than asked for


src/domain/activity-feed/activity.feed.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    const spacesWithCredentials = Array.from(
+      credentialMap.get('spaces')?.keys() ?? []
+    );
+    // get only Spaces specified in the filter
+    const filteredSpaceIds = spaceIdsFilter.length
+      ? intersection(spaceIdsFilter, spacesWithCredentials)
+      : spacesWithCredentials;
+    // check if the Spaces exist;
+    // a Space might not exist if it's deleted or broken/orphaned data was introduced
+    const trueOrList = await this.spaceService.spacesExist(filteredSpaceIds);
```
valentinyanakiev: `trueOrList` is the most confusing variable name

src/domain/activity-feed/activity.feed.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    // how do you understand what journey is this activity about
+    return this.getPaginatedActivity(collaborationIds, {
+      types,
+      userID: myActivity ? agentInfo.userID : undefined,
+      // visibility: true, // todo; what is this?
+      paginationArgs,
+    });
+  }
+
+  private async getQualifyingSpaces(
```
valentinyanakiev, line 71: This method is hard to read and very long. Something like the one below might be better (again, GPTed, so it could be further improved):

```typescript
private async getQualifyingSpaces(
  agentInfo: AgentInfo,
  options?: Pick<ActivityFeedFilters, 'spaceIds' | 'roles'>
) {
  // Extract filters from options, defaulting to empty arrays if not provided
  const { spaceIds: spaceIdsFilter = [], roles: rolesFilter = [] } =
    options ?? {};

  // Get all Spaces the user has credentials for
  const credentialMap = groupCredentialsByEntity(agentInfo.credentials);
  const spacesWithCredentials = Array.from(
    credentialMap.get('spaces')?.keys() ?? []
  );

  // Get only Spaces specified in the filter, ensuring they exist
  const filteredSpaceIds = this.filterSpacesWithCredentials(
    spaceIdsFilter,
    spacesWithCredentials
  );

  // Get only Spaces with the appropriate roles specified in the filter
  return this.filterSpacesByRoles(credentialMap, filteredSpaceIds, rolesFilter);
}

// Helper function to filter Spaces based on credentials and filter criteria
private filterSpacesWithCredentials(
  spaceIdsFilter: string[],
  spacesWithCredentials: string[]
) {
  return spaceIdsFilter.length
    ? intersection(spaceIdsFilter, spacesWithCredentials)
    : spacesWithCredentials;
}

// Helper function to filter Spaces by roles
private async filterSpacesByRoles(
  credentialMap: Map<string, Map<string, string[]>>,
  filteredSpaceIds: string[],
  rolesFilter: string[]
) {
  if (rolesFilter.length === 0) {
    return filteredSpaceIds;
  }

  const spacesWithAppropriateRoles = filteredSpaceIds.filter((spaceId) => {
    const spaceRoles = credentialMap.get('spaces')?.get(spaceId) ?? [];
    return rolesFilter.some((role) =>
      (spaceRoles as string[]).includes(role)
    );
  });

  return spacesWithAppropriateRoles;
}
```

src/domain/challenge/base-challenge/base.challenge.service.ts
```diff
@@ -284,7 +284,7 @@ export class BaseChallengeService {
     return profile;
   }
 
-  public async getCollaboration(
+  public async getCollaborationOrFail(
     challengeId: string,
     repository: Repository<BaseChallenge>
   ): Promise<ICollaboration> {
```
valentinyanakiev:
```suggestion
  ): Promise<ICollaboration | never> {
```

src/domain/challenge/space/space.service.ts
```diff
@@ -858,9 +858,15 @@ export class SpaceService {
     );
   }
 
-  public async getCollaboration(space: ISpace): Promise<ICollaboration> {
-    return await this.baseChallengeService.getCollaboration(
-      space.id,
+  public async getCollaborationOrFail(
+    spaceIdOrEntity: ISpace | string
+  ): Promise<ICollaboration> {
```
valentinyanakiev:
```suggestion
  ): Promise<ICollaboration | never> {
```

src/services/api/activity-log/activity.log.service.ts
```diff
@@ -79,6 +79,12 @@ export class ActivityLogService {
     return results;
   }
 
+  public async convertRawActivityToResults(rawActivities: IActivity[]) {
```
valentinyanakiev: Add return type

src/services/api/activity-log/activity.log.module.ts
```diff
@@ -45,6 +45,6 @@ import { CalendarEventModule } from '@domain/timeline/event/event.module';
     ActivityLogResolverQueries,
     ActivityLogResolverSubscriptions,
   ],
-  exports: [],
+  exports: [ActivityLogService], // todo
```
valentinyanakiev: todo?

src/services/api/activity-log/dto/activity.log.entry.interface.ts
```diff
@@ -50,7 +50,7 @@ import { IActivityLogEntryCalloutLinkCreated } from './activity.log.dto.entry.ca
     );
   },
 })
-export abstract class IActivityLogEntry {
+export class IActivityLogEntry {
```
valentinyanakiev, line 53: Why did you change that?

src/platform/activity/activity.service.ts
```diff
@@ -80,6 +84,40 @@ export class ActivityService {
...
+      userID,
+      paginationArgs = {},
+    } = options ?? {};
+
+    const qb = await this.activityRepository.createQueryBuilder('activity');
+
+    qb.where({
+      collaborationID: In(collaborationIDs),
+      visibility: visibility,
+    }).orderBy({ createdDate: 'DESC' });
```
valentinyanakiev: I would add an enum (or constants?) `ASC` and `DESC` that are re-shared in the application and would provide them as arguments. It may be that we need things ordered differently in the future.

valentinyanakiev (changes requested):

src/services/api/activity-log/activity.log.module.ts
hero101: I introduced it because this service was not meant to be used outisde of the module.

src/services/api/activity-log/dto/activity.log.entry.interface.ts
hero101: Because the pagination requires an object that can be instantiated. The same is done for Users, Organizations and all journeys.

I could extends the abstract class, but then the concrete class is included in the schema.
Let me know if you can think of a solution.

src/platform/activity/activity.service.ts
hero101: As a rule of thumb I use is to not over engineer things on the first try.
It's already using the type `OrderByCondition`, or a simple `'ASC' | 'DESC'` will be enough.

I have added it as an option.

src/domain/activity-feed/activity.feed.service.ts
hero101: Variables and types in Typescript are often called this way to reflect that you need to assert the type.
How would you called it? Will `successOrNonExistingSpaces` suit?

src/domain/activity-feed/activity.feed.service.ts
hero101: The function in mind is 33 lines long - I wouldn't classify it as long.
The readability I can improve.

This suggestion has some code deleted and semantically isn't good.

valentinyanakiev: > The function in mind is 33 lines long - I wouldn't classify it as long. The readability I can improve.
> 
> This suggestion has some code deleted and semantically isn't good.

33 lines you wouldn't classify as long, but best practices do so. Generally, 20 is considered to be the best practice for method and 200 for a file / service. You can look it up.

PR #3438, opened 2023-11-09
Remove Authorization requirement for Space.collaboration

### LINKED ISSUE #5158: Public VC profiles to be visible to unauthenticated users as well
## Description

As an unauthenticated user exploring Alkemio, I don't like that if I click on a VC avatar from the Contributors page, I get told to sign up first. 
As a proud Alkemio team member, I like to show our VCs, also to unauthenticated users, and I don't get why their profiles are hidden.

## Acceptance criteria

- [ ] Give anonymous users the same visibility rights as registered users for VCs. There will still be VCs not visible if they are account restricted or not visible at all
- [ ] …

## Additional Context
they have the same visibility on the server, it's a client issue.

Screen shots / UX designs / data that are relevant.

ccanos:
This fixes #5158 (client) but I'm not fully sure if this is a proper fix or if it is dangerous in any way.

- I have realized that the resolver for `community` doesn't have `@AuthorizationAgentPrivilege` either.

- With this change I can request `authorization` without errors
```graphql
{
space(ID:"eco1") {
  collaboration{
    id
    authorization{
      myPrivileges
    }
  }
}}
```
image

which is what I need, with `authorization.myPrivileges` I have enough.

And I have confirmed that the other fields: `relations`, `callouts`, `tagsetTemplates`, `timeline` throw an error and come set to `null` if I request them.

But I am concerned about that privilege `CREATE_RELATION`, having that permission may be a sign of another different bug.

### DIFF
--- a/src/domain/challenge/space/space.resolver.fields.ts
+++ b/src/domain/challenge/space/space.resolver.fields.ts
@@ -92,7 +92,6 @@ export class SpaceResolverFields {
     return context;
   }
 
-  @AuthorizationAgentPrivilege(AuthorizationPrivilege.READ)
   @UseGuards(GraphqlGuard)
   @ResolveField('collaboration', () => ICollaboration, {
     nullable: true,


valentinyanakiev:
Conceptually, that is a no-go. I have investigated thoroughly and the whole need for querying `collaboration` on Space comes from [this PR](https://github.com/alkem-io/client-web/pull/5095). `CollaborationPrivileges` are added on `SpaceContext` and later those privileges are used in `CalloutTypeSelect`. I believe that is very very wrong and probably has multiple bugs in it.

1. Collaboraiton can be on any journey, and `CalloutTypeSelect` can be executed on any level, as that is, essentially, where the `Callout` would be. 
2. At the moment `CalloutTypeSelect` gets the permissions from the `Space`.
3. That will lead to a mess on `Challenge` and `Opportunity`. I believe code similar to `useCalloutCreation` is needed (maybe this can be refactored into a provider if it is reused in multiple spaces).
4. Then we can remove the `Collaboration` from the `spaceProvider` queries (which were never intend to, logically, have anything to do with Collaboration - Callouts on conceptual level).

So I believe this needs re-design in the approach, it's not a simple straightforward fix.

ccanos:
Ok, understood. I'm going to extract the authorization check from `useCalloutCreation` and use it for this. it's a refactor but not too big. This PR can be closed then

PR #3442, opened 2023-11-10
added ability to lookup authorization policy by iD, protected by platform admin privilege

techsmyth:
This allows looking up on production the details of the authorization policy where e.g. a privilege was not provided.


alstojanovic (approved):
Pushed small typing fix.

Issue #3447, opened 2023-11-13
BUG:  authorizationReset on platform creates a new auth policy every time


techsmyth:
**Describe the bug**
Running authorization reset on platfom is generating a new authorization policy every time.

This leads to orphaned data in the production database.

**To Reproduce**
To see it run the following query, note that the authorization ID changes:
```
mutation {
  authorizationPolicyResetOnPlatform {
    id
    authorization {
      id
    }
  }
}
```

**Expected behavior**
The same authorization policy should be used and simply the rules in it updated.

PR #3448, opened 2023-11-13
Platform authorization handling; org credentials; org cred cascading privilege to user

techsmyth:
Addresses #3447 

Allows for granting + revoking credentials on organizations

Allows for users to get privileges based on being an associate of an org.

Note: after granting an org as beta tester will need to reset platform authorization. Not ideal, but will work near term. Needs more discussion.


valentinyanakiev:
I don't see the BETA_TESTER privilege propagated to the user in the org. Merging this, but needs a fix.

PR #3471, opened 2023-11-21
Fix broken linkedin avatar auth

valentinyanakiev:


techsmyth (approved):
Migration works to fix document parent storage bucket. Note: make sure do authorization reset after the migration for the document to be visible.

New identities created via LI are also now setting the documents correctly.

PR #3495, opened 2023-12-06
Add limit and order to innovationPacks query

ccanos:
- Adds `queryData: { limit, orderBy }` to innovationPacks query
- Exposes Post/Wb/IF template counts as a field of templateSet


hero101 (approved):
Tested locally.

PR #3511, opened 2023-12-12
Serving finalized invitations

me-andre:
Some other functionality may be affected, needs to be tested.
From the point of the API and the client, this is a fix: applications with state = 'approved' and invitations with state = 'rejected' couldn't be requested even if such states were specified.


hero101 (approved):
Looks good to me. May break some test that rely on a strict number of invitations.
@Comoque1 with the new introduced behaviour if you do not specify `states` all the invites will be returns (even the finalized). With the old behaviour finalized invites were never returned.

PR #3512, opened 2023-12-13
Changed the order of execution on user reset

hero101:
`clonedAnonymousReadAccessAuthorization` was cloned before the auth is reset and then `clonedAnonymousReadAccessAuthorization` is compared against.


Comoque1 (approved):

Comoque1:
Tested OK with mail registration flow on sandbox env.

PR #3539, opened 2024-01-11
Squash old migrations into one.

alstojanovic:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


techsmyth:
@alstojanovic @valentinyanakiev super to see this one coming in :) One thought though is that I would like these migrations still visible in the code base as copying from older ones is a huge timesaver, and when doing new ones then nice to be able to reference them rather than making new code / changing the head. Maybe create an "old migrations" folder?

alstojanovic:
There was an old-migrations folder while working but decided to remove it in the end. I don't see an issue with having old migrations in a separate folder. @valentinyanakiev thoughts?

valentinyanakiev (changes requested):
This doesn't work. As discussed, there are two scenarios:

1. There are no prior migrations and we have an empty `alkemio` schema. Then the new migration should create the schema.
2. There were prior migrations and. Then the new migration should NOT create the schema.

Also tests should pass.

valentinyanakiev (dismissed):
Works now.

PR #3540, opened 2024-01-11
ExcalidrawServer sends BROADCAST_READY when socket is set up

me-andre:
image

1. A socket(client) sends a request message to the server
 to request files from other sockets(clients)
2. Sockets(clients) respond with files
3. Server relays the response from (2) to the other sockets(clients)
4. Client handles the response message if needed

### DIFF
--- a/src/services/external/excalidraw-backend/excalidraw.server.ts
+++ b/src/services/external/excalidraw-backend/excalidraw.server.ts
@@ -36,6 +36,7 @@ import {
   SERVER_SAVE_REQUEST,
   SocketIoServer,
   RemoteSocketIoSocket,
+  BROADCAST_READY,
 } from './types';
 import { CREATE_ROOM, DELETE_ROOM } from './adapters/adapter.event.names';
 import {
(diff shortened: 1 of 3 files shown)


src/services/external/excalidraw-backend/excalidraw.server.ts
```diff
@@ -221,6 +223,12 @@ export class ExcalidrawServer {
         (roomID: string, data: ArrayBuffer) =>
           serverVolatileBroadcastEventHandler(roomID, data, socket)
       );
+      // separate channel for sending requests between sockets; available to all sockets, no matter the privileges
+      // socket is sending a request to the server, so the server broadcasts it to all other sockets
+      // the separate channel is required, because not all sockets can broadcast through SERVER_BROADCAST
```
me-andre: I would reword to reflect that this for the users with "Read" privilege only / "Viewers" etc. Current description can be a bit unclear if you're out of the context.

me-andre: Maybe it makes sense to rename channels to UPDATES_BROADCAST and REQUESTS_BROADCAST?

hero101: You are right about the description. But there is just one channel on which the socket is sending a request from files, which logically I called `server-request-broadcast`

me-andre: This description is still very technical and context-dependent on the current story. I could easily imagine in a couple of month reading this and not recovering the purpose of the channel. Can we use the following description?

> A channel where all viewers of the whiteboard can broadcast messages.
> Currently used to send requests for missing data, to pull from other whiteboard collaborators.
> This channel is needed because not all viewers has write privilege so they can't use the channel where updates are broadcasted.

hero101 (approved):

PR #3570, opened 2024-02-01
Fixed ordering of journeys per latest activity

valentinyanakiev:
Group by + order by together on sql level didn't work. It was grouping first, ordering later, and it should have been the other way round.

This seems to be working now and is performant. It took 67 ms for me with real dataset.

### DIFF
--- a/src/platform/activity/activity.service.ts
+++ b/src/platform/activity/activity.service.ts
@@ -162,11 +162,10 @@ export class ActivityService {
       .where('activity.triggeredBy = :triggeredBy', {
         triggeredBy,
       })
-      .groupBy('activity.collaborationId')
       .orderBy('activity.createdDate', 'DESC')
       .getMany();
 
-    // Get unique collaboration IDs from activities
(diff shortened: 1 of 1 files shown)


src/platform/activity/activity.service.ts
```diff
@@ -185,9 +184,24 @@ export class ActivityService {
...
     // Filter activities that have a corresponding collaboration
-    const activityData = activities
-      .filter(activity => collaborationMap.has(activity.collaborationID))
-      .slice(0, limit);
+    const filteredActivities = activities.filter(activity =>
+      collaborationMap.has(activity.collaborationID)
+    );
+
+    // Create a map of collaboration IDs to latest activities
+    const latestActivityMap = new Map();
```
alstojanovic: I would only rename this to be `latestActivityPerJourneyMap` or `latestActivityPerCollaborationMap` because it will be easier to understand the code when we come back to it after some time.

alstojanovic (changes requested):

alstojanovic (approved):

PR #3573, opened 2024-02-02
Timers deleted on room globally deleted 

hero101:
- removed the retry logic - introduced complexity and potential bugs
- timeout converted to interval, managed in a central place
- more verbose logging added
- each server sends an event to inform the others in the cluster that the last socket from a room is disconnected, so the room is deleted globally. This is needed in the case where no sockets are connected locally to an instance, but the room still exists globally (on the other instances), so the timers are there to handle saving and contributions.

## Please squash on merge


valentinyanakiev (dismissed):
Code looks logical.
@alstojanovic please have a look on sandbox whether it behaves as it should.

alstojanovic:
The timers are getting removed but there might be a different issue. Logs are missing when a second user is emitted to the room.
Please do not merge untill fully tested.

alstojanovic (approved):

PR #3592, opened 2024-02-07
Activity feed service fix

valentinyanakiev:
The activity feed service `getAllAuthorizedCollaborations`'s implementation is expecting Promise<string[]> but the logic returned Promise<string[] | never>.

I have altered the logic to work as expected, based on standard notation we use, and standard contracts we use. If it is `getAllAuthorizedCollaborationsOrFail`, I'd expect the behaviour observed before the fix (- the handling of the childCollaborations).


hero101 (dismissed):
Ideally you should try-catch only the throwable, but this is also fine.

PR #3608, opened 2024-02-13
allow platform admins to update email addresses, nameID

techsmyth:
The PLATFORM_ADMIN privilege is only granted to GlobalAdmins

Minor tidy up in save method of UserService

### DIFF
--- a/src/domain/community/user/dto/user.dto.update.platform.settings.ts
+++ b/src/domain/community/user/dto/user.dto.update.platform.settings.ts
@@ -1,0 +1,27 @@
+import { MID_TEXT_LENGTH } from '@common/constants/entity.field.length.constants';
+import { NameID } from '@domain/common/scalars/scalar.nameid';
+import { Field, InputType } from '@nestjs/graphql';
+import { IsEmail, MaxLength } from 'class-validator';
+
+@InputType()
+export class UpdateUserPlatformSettingsInput {
+  @Field(() => String, {
+    nullable: false,
+    description:
+      'The identifier for the User whose platform managed information is to be updated.',
+  })
+  userID!: string;
+
+  @Field(() => NameID, {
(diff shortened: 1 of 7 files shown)


src/domain/community/user/dto/user.dto.update.platform.settings.ts
```diff
@@ -0,0 +1,27 @@
...
+    description: 'Upate the URL path for the User.',
+  })
+  nameID?: string;
+
+  @Field({
+    nullable: true,
+  })
+  @IsEmail()
+  @MaxLength(MID_TEXT_LENGTH)
+  email!: string;
```
valentinyanakiev: There is disconnect between the decorator and the property, should be:
```suggestion
  email?: string;
```

src/domain/community/user/user.service.ts
```diff
@@ -545,12 +542,17 @@ export class UserService {
...
-      await this.userRepository.save(user);
+      await this.save(user);
       await this.setUserCache(user);
     }
 
     return user;
   }
+
+  async save(user: IUser): Promise<IUser> {
+    return await this.userRepository.save(user);
```
valentinyanakiev, line 553: User implements IUser, but is not the same. If a DTO has more properties from `User` than the interface , these properties won't be updated in the database.

valentinyanakiev (changes requested):
Two minor things to check.

src/domain/community/user/user.service.ts
techsmyth: I think you are correct, but this is the same pattern that we have on all entities - so we do want to fix it then we need to do it structurally - not just here. So would suggest to take it separately from this PR.

valentinyanakiev (approved):

PR #3616, opened 2024-02-15
clean up of spaces filtering / sorting to allow for choosing a different sort order

techsmyth:
Code clean up; previous code mixed up a couple of phases. 

Should also be marginally faster as filtering on visibility on db query, not in memory.

### DIFF
--- a/src/domain/challenge/space/space.resolver.queries.ts
+++ b/src/domain/challenge/space/space.resolver.queries.ts
@@ -36,7 +36,7 @@ export class SpaceResolverQueries {
     @Args({ nullable: true }) args: SpacesQueryArgs
   ): Promise<ISpace[]> {
     if (!innovationHub) {
-      return this.spaceService.getSpaces(args);
+      return this.spaceService.getSpacesSorted(args);
     }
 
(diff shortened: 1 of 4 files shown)


hero101:
I don't know if it due to this code or the existing functionality, but the sorting doesn't seem right.
With this query it produces these results. The archived one is on the top, the active in the middle.

``` json
{
  spaces(filter:{visibilities:[ACTIVE, DEMO, ARCHIVED]}) {
    license {
      visibility
    }
  }
}
```

```json
{
  "data": {
    "spaces": [
      {
        "license": {
          "visibility": "ARCHIVED"
        }
      },
      {
        "license": {
          "visibility": "ACTIVE"
        }
      },
      {
        "license": {
          "visibility": "DEMO"
        }
      }
    ]
  }
}
```

techsmyth:
With develop:
image

with new branch:

image


So results are the same, so this PR can be merged.

However, the sorting algorithm does have a bug. At the moment it first sorts to put DEMO last - but does not deal with ARCHIVED - so it is possible to get archived before some active ones - if one requests the archived spaces (which basically never happens apart from admin screens).

hero101 (approved):

PR #3622, opened 2024-02-16
tidy up of privileges around contributions

techsmyth:
Related to who created the contribution.

At Contribution level:
- privileges assigned cascading: C, R, U 
- privileges assigned NOT cascading: D

At link level:
- add in the D privilege before cascading to Profile

At Post level
- add in the D privilege *after* cascading to Comments but before cascading to Profile

At WhiteboardRt:
- removed C, R, U as that is now inherited from


valentinyanakiev (approved):
This looks logical but fixing this functionally requires changes on the client. Still, when a user creates a link in a collection of links, the client doesn't allow it to be deleted from the same user.

PR #3638, opened 2024-02-21
new module to hold agent info + associated cache service

techsmyth:
addresses a logical error in where the agentInfoCache service was located: it is not related to the Agent domain model entity.

Moved all agent info services / logic + the cache service into a separate module to avoid circular dependencies, and it is also then cleaner.


valentinyanakiev (approved):
Works for me locally and seems better than what it was before, although if we dig deeper there are more questions raised, like:
- do we need generic service to update cache per entity, or specific one
- if the above is true, do we need a separate module other than the auth module etc.

Regardless, I am merging this as it is better than what it was, but the whole caching space requires a deeper thought and design discussion.

PR #3651, opened 2024-02-24
sort challenges / opportunities fields using displayName

techsmyth:


valentinyanakiev (commented):
So we are changing in everywhere? Is that what we intend to do or we want to provide the option to do it either way and chose per case?

techsmyth:
yes lets change it everywhere, there is no business reason to prefer sorting on NamedID over displayName - and the latter is much more visible / understandable for users.

alstojanovic (approved):
I fixed the challenges retrieval. The options object wasn't passed in one case.
Couldn't find examples where we expect challenges/opportunities to explicitly come sorted by nameID.
Checked the views from the acc and it looks much more organised!

PR #3653, opened 2024-02-24
Replace usage of xstate lifecycle with a simple set of states; split defaults from templates set; simplify innovation flow creation

techsmyth:
The core of this PR is that the InnovationFlow entity is extended with a set of States, replacing its usage of Lifecycle. The set of states is managed by a new module InnovationFlowStates. 

In addition it addresses the following:
* The TemplateSet module was actually responsible for loading in the example templates into the Space Library
* The InnovationFlow entity was looking after which set of states it should be initiatlized with
* The Space Library filled two functions a) library b) default templates. This was confusing its role, and as a result also making the whole InnovationFlow module as well as the TemplatesSet very intertwined + hard to follow. 

To address the above, a new module called SpaceDefaults has been added. It looks after
* storing the default innovation flow to use for new Challenges *and* Opportunities (same one for both)
* holding the logic for creating the input for creating a new innovation flow
* populating the Space library with example content

This makes the InnovationFlow entity significantly simpler. It also makes the TemplatesSet entity much simpler. 

Other changes:
* Removed the concept of InnovationFlowType, which was only really there to allow for default flows for challnges / opportunities in the Space Library
* Moved InnovationFlow to be under Collaboration in the code base where it logically belongs
* Moved Opportunity to be under Challenge in the code base where it logically belongs
* Moved InnovationFlow to be part of the Collaboration entity
* Simplified the usage of TagsetTemplate class
* removed policy on TemplatesSet, as that was really just a way to avoid deleting the default flows on Space Library
* As part of removing policy on TemplatesSet, moved mutation for removing a InnovationFlowTemplate back to that module, away from TempaltesSet module where it had been to deal with the policy
* profile on innovation flow has some values filled in. Later this should be more formally supported by having this on the InnovationFlowTemplate entity, but simply not there at the moment. 

Note: for callouts display in a challenge / opportunity there should not be any difference as the set of states are still managed by the TagsetTemplate. 

Finally, this new setup should actually also make it easier to support having more defaults data driven, as there is now a logical place for them. E.g. being able to have the default flows in dutch instead of english

Migration:
- up: migrates the data, including mapping from xstate definition to new states setup
- down: reverts back data structure + constrainsts; no data mapping as basically not feasible

Todo: update the migration to get the db schema properly setup.


techsmyth:
@valentinyanakiev please have a look at this PR when you can. It is to unblock the wider usage of innovation flows. Consider this a prototype :) But it does address quite some bad code smells as well as a key piece of community feedback.

valentinyanakiev (approved):

PR #3693, opened 2024-03-09
Spaces entities clean up; Account entity

techsmyth:
Introduce a new Account module to hold the following information:
- library
- defaults
- license
- host
All spaces, challenges + opportunities within a given hierarchy share the same account and can use it to access shared data.

Removed Project + Agreement modules.
Removed hierarchy within challenges; to be replaced by space hierarchy.

Moved storage aggregator to be part of BaseChallenge.

Migration up + down is added; runs ok re up / down / up / down loop.

Issues are raised for test suite updates, documentation in Alkemio repo.


src/domain/challenge/space/space.entity.ts
```diff
@@ -3,13 +3,9 @@ import { ISpace } from '@domain/challenge/space/space.interface';
...
 import { Challenge } from '@domain/challenge/challenge/challenge.entity';
 import { PreferenceSet } from '@domain/common/preference-set/preference.set.entity';
-import { TemplatesSet } from '@domain/template/templates-set/templates.set.entity';
-import { StorageAggregator } from '@domain/storage/storage-aggregator/storage.aggregator.entity';
-import { License } from '@domain/license/license/license.entity';
-import { SpaceDefaults } from '../space.defaults/space.defaults.entity';
 @Entity()
 export class Space extends BaseChallenge implements ISpace {
-  @OneToMany(() => Challenge, challenge => challenge.parentSpace, {
+  @OneToMany(() => Challenge, challenge => challenge.space, {
```
valentinyanakiev, line 9: Why is `parentSpace` changed to `space`?

valentinyanakiev (commented):
There is a fundamental underlying issue:
- `Account`'s relationship to `Space` is one-to-one
- `Account`'s relationship to `Challenge` and `Opportunity` is one-to-many
- `BaseChallenge` is an abstract class, and not an Entity (so it can not be used in defining relationships)
- We can't alter the relationship that way between `Account` and `BaseChallenge` defined on both sides to have it one-to-many everywhere
- We still don't have the 3 Journeys collapsed into one

That leaves us with two options I see:
- Collapse the entities now and define the relationship in the collapsed entity
- Use one-to-one to `Space` and one-to-many in `Challenge` and `Opportunity`. Then remove that relationships one week later.

In the former we'll end up with huge and risky PR, with the latter we'll redo the work right after we have done it. I may be missing something but this requires a discussion. @techsmyth thoughts?

src/domain/challenge/space/space.entity.ts
techsmyth: it may e better as parentSpace, but lets pick this up when we do the merge...

PR #3705, opened 2024-03-11
Added extra caching

valentinyanakiev:


src/domain/agent/agent/agent.service.ts
```diff
@@ -136,18 +136,39 @@ export class AgentService {
...
+  }
+
+  private async getAgentFromCache(id: string): Promise<IAgent | undefined> {
+    return await this.cacheManager.get<IAgent>(this.getAgentCacheKey(id));
+  }
+
+  private async setAgentCache(agent: IAgent): Promise<IAgent> {
+    const cacheKey = this.getAgentCacheKey(agent.id);
+    return await this.cacheManager.set<IAgent>(cacheKey, agent, {
+      ttl: 60,
```
techsmyth: This seems a bit short, make longer?

techsmyth (dismissed):
Makes sense. Also reminds me that we need to address the agendInfo cache PR that is hanging around. So please go ahead and merge - only comment is whether the ttl should not be longer?

src/domain/agent/agent/agent.service.ts
valentinyanakiev: Increased it to 15 minutes and is configurable now.

PR #3712, opened 2024-03-14
Collaboration with CalloutGroups

techsmyth:
New entity: CalloutGroup, which allows the set of groups that can be applied to a callout to be data driven. 

Replaces usage of DisplayLocation, which is too specific to the space / challenge / opportunity hierarchy to be in the api

Migration up is in place; very basic migration down. 

to do: 
- test
- update test suite


Comoque1:
The server is not starting
image

valentinyanakiev (approved):

PR #3715, opened 2024-03-17
InnovationFlow: update single state + keep existing callouts, validation

techsmyth:
New mutation to allow updating a single state in an innovation flow, and importantly not resetting all existing callouts in that state to the default state.

Added validation to check the set of state names is unique, and test for this when updating the set of state defintions, single state, template etc.


ccanos (approved):

ccanos:
license/cla blocked

PR #3716, opened 2024-03-17
Add community guidelines entity 

techsmyth:
New entity to hold community guidelines, which at the moment is very simple: a single profile.


src/domain/collaboration/collaboration/collaboration.interface.ts
```diff
@@ -8,10 +8,6 @@ import { IInnovationFlow } from '../innovation-flow/innovation.flow.interface';
 
 @ObjectType('Collaboration')
 export abstract class ICollaboration extends IAuthorizable {
-  @Field(() => [ICallout], {
-    nullable: true,
-    description: 'List of callouts',
-  })
```
alstojanovic, line 14: Why is this removed?

techsmyth: There is a dedicated field resolver, so this is a bug

valentinyanakiev (approved):

PR #3718, opened 2024-03-18
Add grouped activity feed query.

alstojanovic:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/domain/activity-feed/activity.feed.resolver.queries.ts
```diff
@@ -41,4 +43,27 @@ export class ActivityFeedResolverQueries {
...
+  ): Promise<IActivityLogEntry[]> {
+    this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      await this.platformAuthorizationService.getPlatformAuthorizationPolicy(),
+      AuthorizationPrivilege.READ_USERS,
+      `Activity feed query: ${agentInfo.email}`
+    );
+
+    return this.feedService.getGroupedActivityFeed(agentInfo, {
+      ...args,
```
hero101: there is no need for a spread

src/platform/activity/activity.service.ts
```diff
@@ -148,6 +148,69 @@ export class ActivityService {
...
+      defaultCondition,
+      typesCondition,
+      triggeredByCondition,
+    ]
+      .filter(condition => condition !== undefined)
+      .join(' AND ');
+
+    const groupedActivities: {
+      latest: string;
+    }[] = await this.entityManager.connection.query(
```
hero101, line 185: is there a reason to not use the repository class instead? - that will reduce much of the code. Also it's too complicated for the use case.

src/platform/activity/activity.service.ts
```diff
@@ -148,6 +148,69 @@ export class ActivityService {
...
+      WHERE ${whereConditions}
+      group by activity.resourceId, activity.triggeredBy, activity.type
+      order by latest ${orderBy}
+      ${limit ? `LIMIT ${limit}` : ''};
+      `
+    );
+
+    const activityIDs = groupedActivities.map(a => a.latest);
+
+    const qb = this.activityRepository.createQueryBuilder('activity');
```
hero101: is there a reason to not use `find` instead?

src/domain/activity-feed/grouped.activity.feed.query.args.ts
```diff
@@ -0,0 +1,40 @@
+import { Field, InputType } from '@nestjs/graphql';
+import { ActivityEventType } from '@common/enums/activity.event.type';
+import { UUID } from '@domain/common/scalars';
+import { ActivityFeedRoles } from './activity.feed.roles.enum';
+
+@InputType()
+export class GroupedActivityFeedQueryArgs {
```
hero101: I would suggest to reuse the existing input type as much as possible. This is not a separate query but a workaround, and when one query changes you need to change the other one.

src/domain/activity-feed/activity.feed.resolver.queries.ts
```diff
@@ -41,4 +43,27 @@ export class ActivityFeedResolverQueries {
 
     return this.feedService.getActivityFeed(agentInfo, { ...args, pagination });
   }
+
+  @UseGuards(GraphqlGuard)
+  @Query(() => [IActivityLogEntry], {
+    nullable: false,
+    description: 'Activity events related to the current user.',
```
hero101: - Please include some text about why this is a separate query.
- change the name of  the query to something like `activityFeedGrouped`. Better for searching and grouping queries.

hero101 (changes requested):
also revert the grouping on the paginated query

src/platform/activity/activity.service.ts
alstojanovic: I tried to use the activityRepository but I was not able to add a custom select for MAX(rowId). I could not retrieve the 'latest' column as the returned value is Activity.

hero101 (approved):

PR #3730, opened 2024-03-20
Forum entity to hold Discussions

techsmyth:
Forum entity added to hold Discussions. Split out from Communication, and only used under Platform. 

This clears out a big pile of functionality that was not used within Spaces. 

+ add privacy mode to discussions in preparation for authorization control
Only adding flag for now; usage + modes depend on further discussion. Slight improvement in authorization policy effectiveness.

Moved anonymous read access to child entities; not on platform level itself. 

To TEST:
- ensure do authorization reset on platform!


me-andre (commented):
Started review

src/common/enums/forum.discussion.privacy.ts
```diff
@@ -0,0 +1,11 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum ForumDiscussionPrivacy {
+  AUTHOR = 'author',
+  AUTHENTICATED = 'authenticated',
+  PUBLIC = 'public',
+}
+
+registerEnumType(ForumDiscussionPrivacy, {
```
me-andre, line 9: Ideally, GraphQL-related calls shouldn't go along with base enum declarations since GraphQL is the API, hence, View layer and this enum is business layer or the Model in MVC terms. In practice, as a developer I may want to import this enum in an environment where GraphQL is not initialized or even not able to be initialized.
I'm posting it mainly to create awareness because it's done like that everywhere in the app anyway.

valentinyanakiev (approved):
Seems to be working to me. Verified with client:
- No regressions found on old messages / updates
- Discussions are under Forum
- Authorization policy reset had the policy reset of discussions
- Old forum discussions were accessible
- I could upload an image to a discussion

PR #3757, opened 2024-03-26
callout framing url generation on callout template

techsmyth:
Logic needed to account for CalloutFraming being used on both Callouts + CalloutTemplates.

Tidied up logic for creating callout url path

Tidied up logic for finding whiteboard urls

Note: this is a precursor to 5166.

### DIFF
--- a/src/services/infrastructure/url-generator/url.generator.service.ts
+++ b/src/services/infrastructure/url-generator/url.generator.service.ts
@@ -18,7 +18,7 @@ import { Space } from '@domain/challenge/space/space.entity';
 @Injectable()
 export class UrlGeneratorService {
   cacheOptions: CachingConfig = {
-    ttl: 300000, // milliseconds
+    ttl: 1, // milliseconds
   };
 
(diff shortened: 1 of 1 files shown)


src/services/infrastructure/url-generator/url.generator.service.ts
```diff
@@ -540,72 +579,70 @@ export class UrlGeneratorService {
       calloutId: string;
     }[] = await this.entityManager.connection.query(
       `
-        SELECT callout_contribution.id as contributionId, callout_contribution.calloutId as calloutId FROM callout_contribution
+        SELECT callout_contribution.id, callout_contribution.calloutId as calloutId FROM callout_contribution
```
valentinyanakiev: It doesn't seem we use the `callout_contribution.id`
```suggestion
        SELECT callout_contribution.calloutId as calloutId FROM callout_contribution
```

valentinyanakiev (changes requested):

valentinyanakiev (approved):

PR #3778, opened 2024-03-29
add community guidelines to all non-space communities

techsmyth:


alstojanovic (approved):
The issue was for orphan communities with no parent space/challenge/opportunity. The prune-orphans script help validate the results.

PR #3781, opened 2024-03-31
revert error handling change that is breaking errors cascading

techsmyth:
Reverting this change + the errors are again being properly cascaded
Not sure what the issue is / what causes the problem, also not sure how this extra code works - so good if you can look at this @hero101


hero101:
If you return and object instead of a class it produces the problem.
My whole reasoning was to show a minimal error to the API client and log everything available for internal use.
I have also added a third exception filter which should catch all the other errors that are usually missed by the other two filters.

hero101 (approved):

PR #3800, opened 2024-04-10
check license flag for access to VCs, assign privilege in auth policies + use in mutations

techsmyth:
Not yet tested, code is compiling

### DIFF
--- a/src/domain/communication/room/room.service.events.ts
+++ b/src/domain/communication/room/room.service.events.ts
@@ -32,6 +32,7 @@ import { RoomSendMessageReplyInput } from '@domain/communication/room/dto/room.d
 import { SubscriptionPublishService } from '@services/subscriptions/subscription-service/subscription.publish.service';
 import { RoomService } from './room.service';
 import { VirtualContributorService } from '@domain/community/virtual-contributor/virtual.contributor.service';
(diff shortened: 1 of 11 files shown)


src/domain/communication/room/room.service.events.ts
```diff
@@ -52,10 +53,19 @@ export class RoomServiceEvents {
...
     question: Pick<IMessage, 'id' | 'message'>,
     agentInfo: AgentInfo,
-    room: IRoom
+    room: IRoom,
+    accessToVirtualContributors: boolean
   ) {
     for (const mention of mentions) {
       if (mention.type === MentionedEntityType.VIRTUAL_CONTRIBUTOR) {
+        // Only throw exception here for the case that there is an actual mention
+        if (!accessToVirtualContributors) {
```
me-andre, line 62: Do we really need this? If VCs are unaccessible, you can't add them.

techsmyth: Yes we do, as otherwise we are relying on the client logic. If someone sends in a comment mutation with a vc tagged in there the code will respond. The logic has to be enforced server side, not rely on client interaction. The alternative would be to check if the tagged vc is a member of the community (also valid) but this is simpler and actually stronger.

PR #3818, opened 2024-04-12
Allow for description on innovation flow state to be left undefined

techsmyth:
The api now allows for the description on an innovation flow state to not be specified.

The client may also need updating?


ccanos (approved):

techsmyth:
@ccanos think this can be just merged right? Is there also client work needed?

ccanos:
@techsmyth  Client changes are just finished now. https://github.com/alkem-io/client-web/pull/5912 we are about to merge

PR #3861, opened 2024-04-23
If user exists, external invitation creates a normal one instead

me-andre:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/domain/community/community/community.resolver.mutations.ts
+++ b/src/domain/community/community/community.resolver.mutations.ts
@@ -1,5 +1,5 @@
 import { Inject, UseGuards } from '@nestjs/common';
-import { Args, Mutation, Resolver } from '@nestjs/graphql';
+import { Args, createUnionType, Mutation, Resolver } from '@nestjs/graphql';
 import { IUserGroup } from '@domain/community/user-group';
 import { CommunityService } from './community.service';
 import { CurrentUser, Profiling } from '@src/common/decorators';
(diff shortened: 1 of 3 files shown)


me-andre:
Please review this one as well https://github.com/alkem-io/client-web/pull/5986

src/domain/community/community/community.resolver.mutations.ts
```diff
@@ -382,40 +393,60 @@ export class CommunityResolverMutations {
       `create invitation community: ${community.id}`
     );
 
-    const invitations: IInvitation[] = [];
+    return Promise.all(
```
hero101, line 396: I believe some handling is required here. I am not sure what will happen if one invitation fails.

src/domain/community/community/community.resolver.mutations.ts
```diff
@@ -424,7 +455,7 @@ export class CommunityResolverMutations {
     @CurrentUser() agentInfo: AgentInfo,
     @Args('invitationData')
     invitationData: CreateInvitationExternalUserOnCommunityInput
-  ): Promise<IInvitationExternal> {
+  ): Promise<IInvitation | IInvitationExternal> {
```
hero101, line 458: It bugs me that the mutation is named `inviteExternalUser...` but you are allowed to use it on anyone.

hero101 (changes requested):

src/domain/community/community/community.resolver.mutations.ts
me-andre: Same as before: an exception is thrown
Not same as before: all non-failed invitation would attempt to complete (which is better imo because we may not want to end in undefined state)

src/domain/community/community/community.resolver.mutations.ts
me-andre: Yepp, would you suggest a rename? It's actually invites by email regardless of whether a user exists, so maybe `inviteForCommunityMembershipByEmail`?

hero101: Yes! That's even better.

hero101 (approved):

PR #3871, opened 2024-04-29
Account auth fixes

valentinyanakiev:


techsmyth (commented):
This rule should already be inherited from the root policy. What was the symptom that made you add this here?

PR #3874, opened 2024-04-29
License engine

techsmyth:
Multi role setup for license enforcement, as preparation for moving to a much more comprehensive license enforcement.

Design: https://github.com/alkem-io/alkemio/blob/develop/docs/license-entitlements.md

**LicenseEngine module**: new module that is responsible for checking whether a given LicensePrivilege is granted.:
- this module has basically no dependencies, so that it can later be isolated out into a separate microservice.

**LicensePolicy module**: new module for managing the definition of the applicable license rules
- only one at the moment, stored under Platform entity
```
query {
  platform {
    licensePolicy {
       id
      authorization {
        id
        myPrivileges
      }
      featureFlagRules {
        name
        grantedPrivileges
        featureFlagName
        
      }
    }
  }
}
```
Note: design should allow for multiple LicensePolicies to be in use, even if there is only one at the moment. Otherwise we could end up with just ginormous sets of rules etc over time. It is also feasible to think that there may be different licenses applied in different contexts etc.

**License module**: updated
- License.privileges field added for easy checking what LicensePrivileges are accruing to a License.
```
query {
  spaces {
    account {
      license {
        privileges
      }
    }
  }
}
```
- removed isFeatureFlagEnabled method on service (replaced by LicenseEngineService logic)

Migration up + down added.

All license checks updated to be via checking for a LicensePrivilege using the engine

Note: this is a pure internal change for the server, the client is not impacted at all, nor are the tests.


valentinyanakiev (approved):
Haven't tested all edge cases, migration passed, seems logical, the field resolver / API works fine, I haven't experienced any regression.

Design-wise, looks logical, but seems like a lot of this code is redundant (we can re-use a service with authorization - something we can fix after this month...).

Merging this.

PR #3897, opened 2024-05-09
Space query to only lookup top level spaces

techsmyth:
+ fix from Carlos for where clause on space query


valentinyanakiev:
This doesn't work with real data, nor does the last change (space returning only root). There are 60 graphql files on the client using `space` and the application is broken in many places with it. Reverted the change on develop until we decide what to do with it.

This change here doesn't lookup correctly either:
image

I suggest we don't waste time on this for now with dozens of other things on the agenda.

valentinyanakiev:
Now works with the latest set of client changes merged.

PR #3905, opened 2024-05-09
Create & Read - Community Guidelines template

EmiliaPavlova:
### The final implementation
When a template is created from input that will create a new entity in it's table and the template will mantain a 1:1 relation to that NEW entry.
When a template is created from entity, a new entity from the same type will be duplicated and a 1:1 relation to the COPIED entry will be maintain from the template.
### Note
When a duplicate is made, in the case of creating a template from an entity, that duplicate MUST NOT maintain the same relations as the original, meaning a Post must not be a part of a Callout, etc - a pure entry not appearing anywhere on the platform, ONLY used for that single template.

**This approach will take care of domain updates to the entity and will reduce the custom code and fields that have to be mainated with every new template type.**


src/common/enums/callout.type.ts
```diff
@@ -6,6 +6,7 @@ export enum CalloutType {
   WHITEBOARD = 'whiteboard',
   WHITEBOARD_COLLECTION = 'whiteboard-collection',
   LINK_COLLECTION = 'link-collection',
+  MEMBER_GUIDELINES = 'member-guidelines',
```
techsmyth: Why does the callout type enum get extended?

src/common/enums/profile.type.ts
```diff
@@ -20,6 +20,7 @@ export enum ProfileType {
   INNOVATION_FLOW_TEMPLATE = 'innovation-flow-template',
   POST_TEMPLATE = 'post-template',
   WHITEBOARD_TEMPLATE = 'whiteboard-template',
+  MEMBER_GUIDELINES_TEMPLATE = 'member-guidelines-template',
```
techsmyth: community-guidelines-template?

src/domain/template/member-guidelines-template/dto/member.guidelines.template.dto.create.ts
```diff
@@ -0,0 +1,26 @@
...
+  MID_TEXT_LENGTH,
+  VERY_LONG_TEXT_LENGTH,
+} from '@common/constants/entity.field.length.constants';
+import { Markdown } from '@domain/common/scalars/scalar.markdown';
+import { CreateTemplateBaseInput } from '@domain/template/template-base/dto';
+import { Field, InputType } from '@nestjs/graphql';
+import { MaxLength } from 'class-validator';
+
+@InputType()
+export class CreateMemberGuidelinesTemplateInput extends CreateTemplateBaseInput {
```
techsmyth: General comment: the name should be CommunityGuidelinesTemplate - unless Simone instructed otherwise?

src/domain/template/member-guidelines-template/dto/member.guidelines.template.dto.create.ts
```diff
@@ -0,0 +1,26 @@
...
+  @Field({
+    nullable: false,
+    description: 'The type of Member Guidelines created from this Template.',
+  })
+  @MaxLength(MID_TEXT_LENGTH)
+  type!: string;
+
+  @Field(() => Markdown, {
```
techsmyth: Please add a direct CommunityGuidelines entry here, that is the pattern we should move towards.  So not the defaultDescription field.

src/domain/template/member-guidelines-template/dto/member.guidelines.template.dto.create.ts
```diff
@@ -0,0 +1,26 @@
...
+import { MaxLength } from 'class-validator';
+
+@InputType()
+export class CreateMemberGuidelinesTemplateInput extends CreateTemplateBaseInput {
+  @Field({
+    nullable: false,
+    description: 'The type of Member Guidelines created from this Template.',
+  })
+  @MaxLength(MID_TEXT_LENGTH)
+  type!: string;
```
techsmyth: Not relevant?

src/domain/template/member-guidelines-template/member.guidelines.template.entity.ts
```diff
@@ -0,0 +1,33 @@
...
+    templatesSet => templatesSet.memberGuidelinesTemplates,
+    {
+      eager: false,
+      cascade: false,
+      onDelete: 'NO ACTION',
+    }
+  )
+  templatesSet?: TemplatesSet;
+
+  @Column('text')
```
techsmyth: ditto as on entity

techsmyth (commented):
Nice to see this PR! 

A few general comments:
- the name should align with the entity from the Domain e.g. CommunityGuidelines (not MemberGuidelines)
- the fields that are specific to this template type should be just a single one for CommunityGuidelines - makes clear that the template is wrapping another entity
- I am missing the authorization service being extended to pass through the authorization to child COmmunityGuidelineTemplates, and also that when a TemplatesSet is deleted that it also removes child CommuityGuidelineTemplates

All straightforward ones to fix I think. A couple of other comments.

src/domain/template/member-guidelines-template/dto/member.guidelines.template.dto.create.ts
EmiliaPavlova: Discussed with Simone, should stay MemberGuidelinesTemplate.

src/domain/template/member-guidelines-template/dto/member.guidelines.template.dto.create.ts
valentinyanakiev: @techsmyth  can you please elaborate? Because if we have `CommunityGuidelines` reference, we change the reference, we change the template. This will create kinda dynamic template that changes all the time when the reference changes, if I understand correctly.

techsmyth: @valentinyanakiev the idea is to use the full CommunityGuidelines entity to store the template information - but that will be cloned / copied when using the template. It is not shared across usages. The idea is to get towards having an approach where each template type simply wraps around a domain model entity - not duplicating the setup with a selection of the fields that would ultimately be used. Makes both maintenance easier + I think later we can actually move to a single template type that has one of a number of types (like the CalloutContribution entity)

hero101: The current impl copies the content of the entity at that moment in time and creates a public version of it (template).
If i am understanding it correctly, you suggest to maintain a direct relation to the actual entity which is fundamentally different. Please contact me to discuss.

techsmyth: It should clone/copy the entity wrapped by the template at the moment in time when the template is used to instantiate a new entity. The new entity should **not** maintain a direct relation.

hero101: What do you mean by wrap. The implementation will follow the guideline entity, so the fields from the profile will be  copied to the fields in the profile of the template.
Be aware that you are commenting a PR not ready for a review and this is not the final implementation.

I guess you suggest the input object to pont to a guideline entry, so the whole copy (wrapping) will be done serverside, instead of manually tranferring it from the client.

hero101: ### The final implementation will be:
When a template is created from input that will create a new entity in it's table and the template will mantain a 1:1 relation to that NEW entry.
When a template is created from entity, a new entity from the same type will be duplicated and a 1:1 relation to the COPIED entry will be maintain from the template.
### Note
When a duplicate is made, in the case of creating a template from an entity, that duplicate MUST NOT maintain the same relations as the original, meaning a Post must not be a part of a Callout, etc - a pure entry not appearing anywhere on the platform, ONLY used for that single template.

**This approach will take care of domain updates to the entity and will reduce the custom code and fields that have to be mainated with every new template type.**

src/common/enums/callout.type.ts
EmiliaPavlova: Removed, my mistake.

src/domain/template/member-guidelines-template/dto/member.guidelines.template.dto.create.ts
EmiliaPavlova: All renamed :)

ccanos (approved):

PR #3912, opened 2024-05-13
Prototype authorization accessible from the outside

hero101:
- The server uses a message queue to receive and respond synchronously to authorization requests for whiteboards


valentinyanakiev (approved):
Looks like what we need, haven't tested it. Happy to merge it and give it a go.

PR #3925, opened 2024-05-16
Extended mySpaces to have a flag to show only myCreatedSpaces

valentinyanakiev:
To use, e.g. 

```gql
query {
  me {
    mySpaces(showOnlyMyCreatedSpaces:true) {
      space {
        nameID
      }
    }
  }
}
```

### DIFF
--- a/src/services/api/me/me.service.ts
+++ b/src/services/api/me/me.service.ts
@@ -14,6 +14,8 @@ import { ActivityService } from '@platform/activity/activity.service';
 import { LogContext } from '@common/enums';
 import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
 import { sortSpacesByActivity } from '@domain/space/space/sort.spaces.by.activity';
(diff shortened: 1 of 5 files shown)


src/services/api/me/me.service.ts
```diff
@@ -93,10 +97,35 @@ export class MeService {
...
-        space: activityLog.space,
-        latestActivity: activityLog,
-      });
+      if (!showOnlyMyCreatedSpaces) {
+        mySpaceResults.push({
+          space: activityLog.space,
+          latestActivity: activityLog,
+        });
+      } else {
+        if (activityLog?.space) {
```
bobbykolev, line 106: this check seems redundant

bobbykolev (approved):
Looking good @valentinyanakiev , thanks!

src/services/api/me/me.service.ts
valentinyanakiev: It is not, because activities do not have reference to spaces, and when a space is deleted, this is being hit.

PR #3931, opened 2024-05-16
Invitations to allow also inviteToParent

techsmyth:
Migration to add field to invitation entities

To test:
- enable flag that subspace admins can invite
- ensure that user to be invited is not a member of space
- invite the user to the subspace

Client works for accepting, but does not give any specific message.

Note as the client is not yet updated the only way to invite to subspace is via graphql calls


techsmyth:
@Comoque1 this one needs some good testing...

valentinyanakiev (approved):

PR #3933, opened 2024-05-16
smarter generation of nameIDs 

techsmyth:
The generation of nameIDs now is smart in how it generates the nameID from a displayName:
- generate a candidate nameID from the displayName
- if it is already taken in a given context, add '-1" as suffix + try again, and if that is taken, increase the number + try again etc

Tried with subspaces with subsubspace both using the same title 'test':
http://localhost:3000/eco1/challenges/**test**/opportunities/**test-1**

Used everywhere we were generating nameIDs: user, discussion, callout, callout contributions, calendars. 

Note: pattern used has changed. It is the responsibility of the containing entity to check / create the nameID, not the child. E.g. a Collaboration should look after nameID for a Callout, not the Callout itself which should not know its context.

### DIFF
--- a/src/services/infrastructure/naming/naming.service.ts
+++ b/src/services/infrastructure/naming/naming.service.ts
@@ -23,6 +23,8 @@ import { Collaboration } from '@domain/collaboration/collaboration';
 import { SpaceLevel } from '@common/enums/space.level';
 import { User } from '@domain/community/user/user.entity';
 import { Communication } from '@domain/communication/communication/communication.entity';
+import { Calendar } from '@domain/timeline/calendar/calendar.entity';
+import { Library } from '@library/library/library.entity';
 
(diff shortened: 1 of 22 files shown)


src/services/infrastructure/naming/naming.service.ts
```diff
@@ -12,106 +12,184 @@ import { IPost } from '@domain/collaboration/post/post.interface';
...
+  ): Promise<string[]> {
+    const callout = await this.entityManager.findOne(Callout, {
+      where: {
+        id: calloutID,
+      },
+      relations: {
+        contributions: {
+          whiteboard: true,
+          post: true,
+          link: true,
```
ccanos: not sure if this is needed

ccanos (approved):

PR #3948, opened 2024-05-18
Innovation Hub --> Account Association

valentinyanakiev:
- one side of the relationship defined only (so innovationHub has accountID)
- accountID provided in the `CreateInnovationHub / UpdateInnovationHub` payloads. No full Account entity / payload added, as this should have been created before the `InnovationHub`
- lookup of the Account based on ID + setting it in the `Create / Update`
- No data migration (it's impossible to guess it correctly)
- added field resolver for innovationHub --> account

Note: if `account` is queried and hadn't been associated with this `innovationHub`, an exception is thrown (which has misleading text). This is the current behavior for all dataloaders having exception in their logic, I have not altered that.


techsmyth (approved):
Good to get this moving. More to be done later, will track in a separate issue.

PR #4035, opened 2024-06-03
add VC to root space when created on account

techsmyth:
Code is added but not yet tested. Logically it should be ok - if any issues please just ping me.


Comoque1:
With this branch, the creation of VC fails:
image

Comoque1:
In addition to that, by Space Admins don't have rights to assign users to space, which is also the case for VC. I suspect ones the above comment is resolved, error will be thrown on the execution of the mutation: `AddVirtualContributorToCommunity` 
image

Comoque1:
@techsmyth the above comments are resolved. I get authorization error when I try to delete VC with SpaceAdmin:
image

valentinyanakiev (approved):

PR #4047, opened 2024-06-04
ensure that feature flag to enable interaction with VCs is enabled for user hosted spaces; 

techsmyth:
to be replaced later with proper licensing setup


valentinyanakiev (changes requested):
Assuming this works (haven't tested it), it will fix it for new ones. What about existing ones?

PR #4048, opened 2024-06-04
VCs + AI Server split

techsmyth:
With this PR:
1) in the Collaboration platform domain, we have `VritualContributor` and `AiPersona` entities - currently in 1:1 relation; the VC is the entity which is added to a space and handles the interactions, and the Persona is the entity which handles the actual AI part
2) we've introduced an `AiServer` module (to be extracted in a separate app) which has an `AiPersonaService` entity - which powers the `AiPersona` entity in the main domain model
3) Interactions between the two a handler through the `AiServerAdapter`, which currently is used inside the main domain and loads the `AiServerModule` and invokes its methods (next step is to communicate over HTTP)

All data should be properly migrated. All previous functionality should be preserved.


valentinyanakiev (commented):
I can make this work, but looking more and more into it, it makes less and less sense to me, because:

- We haven't landed on the purpose / division between an `engine` and a `persona`
- Having a `defaultPersona`, equating to `engine`, might not be the path forward
- If we decide to go with `defaultPersona`, there may be a need of `defaultPersonaS`, as they may need to be per `engine`, with the current domain model. I am not sure a default persona of type Expert is indeed THE default persona.

valentinyanakiev:
I have updated the PR with develop so we can easily manage it, but won't spend more time on it until we are firmly committed to the path forward - I don't want to have a short-term solution replaced by another short-term solution.

valentinyanakiev:
[3:34:46 PM] Starting compilation in watch mode...

src/domain/community/ai-persona/ai.persona.service.ts:52:31 - error TS2551: Property 'Knowledge' does not exist on type 'typeof SpaceIngestionPurpose'. Did you mean 'KNOWLEDGE'?

52         SpaceIngestionPurpose.Knowledge
                                 ~~~~~~~~~

  src/services/infrastructure/event-bus/commands/ingest.space.command.ts:5:3
    5   KNOWLEDGE = 'knowledge',
        ~~~~~~~~~
    'KNOWLEDGE' is declared here.

src/domain/community/community/community.resolver.mutations.ts:267:29 - error TS2551: Property 'Context' does not exist on type 'typeof SpaceIngestionPurpose'. Did you mean 'CONTEXT'?

267       SpaceIngestionPurpose.Context
                                ~~~~~~~

  src/services/infrastructure/event-bus/commands/ingest.space.command.ts:6:3
    6   CONTEXT = 'context',
        ~~~~~~~
    'CONTEXT' is declared here.

src/services/ai-server/ai-persona-service/ai.persona.service.service.ts:65:33 - error TS2551: Property 'Knowledge' does not exist on type 'typeof SpaceIngestionPurpose'. Did you mean 'KNOWLEDGE'?

65           SpaceIngestionPurpose.Knowledge
                                   ~~~~~~~~~

  src/services/infrastructure/event-bus/commands/ingest.space.command.ts:5:3
    5   KNOWLEDGE = 'knowledge',
        ~~~~~~~~~
    'KNOWLEDGE' is declared here.

[3:34:53 PM] Found 3 errors. Watching for file changes.

valentinyanakiev (approved):
Will validate on ACC. Migrations  + Schema works.

PR #4066, opened 2024-06-06
VC privileges for Account Host

valentinyanakiev:
- Split Account and Account Host module because of circular dependencies
- Give READ, UPDATE, DELETE privileges to ACCOUNT_HOST

This is caused by the ACCOUNT --> SPACE --> ACCOUNT loop in inheriting the authorization policy. Because the VIRTUAL_CONTRIBUTOR needs the rules from SPACE_ADMIN, which is on SPACE level, and is not initialized at ACCOUNT creation time, no privileges are inherited from that rule.

I have added an explicit HOST rule on the VIRTUAL_CONTRIBUTOR, which fixes the problem, but the root cause is still there (the root cause also caused the patch needed yesterday). There is no easy fix, it's a design flaw.

@techsmyth this fixes the flow for now. Also, after auth reset GRANT and CREATE are added. To be consistent, we may add them, although they don't make any difference.

P.S. I have tried an extra .save just before applying authorization policy on VCs, but that didn't help (and will make things significantly slower as well).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "alkemio-server",
-  "version": "0.78.8",
+  "version": "0.78.9",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "alkemio-server",
-      "version": "0.78.8",
+      "version": "0.78.9",
       "license": "EUPL-1.2",
       "dependencies": {
         "@alkemio/matrix-adapter-lib": "^0.3.6",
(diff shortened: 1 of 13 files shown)


techsmyth (dismissed):
Logic is clear. The fact the whole setup is complex is a sign we need to think through more the relationships + roles in this space. To work through next week in BG.

I have not run it but will leave that to Evgeni.

Comoque1 (approved):

PR #4068, opened 2024-06-07
Add COMMUNITY_ADD_MEMBER on Root Space

valentinyanakiev:


bobbykolev (dismissed):

techsmyth:
@valentinyanakiev this may need a discussion. An absolute No that a space admin any user can add to the root space community. Cannot check full code now to check what is implemented.

valentinyanakiev:
> @valentinyanakiev this may need a discussion. An absolute No that a space admin any user can add to the root space community. Cannot check full code now to check what is implemented.

That's why I tagged you for review @techsmyth. _[On a Space level Add button shows all accosiated VCs to that Account.](https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/client-web/6311) - how do we visualize the section for VCs - I was assuming it is COMMUNITY_ADD_MEMBER (works for subspaces), but it could be no privilege check for that section, or a new privilege (e.g. COMMUNITY_ADD_VIRTUAL_CONTRIBUTOR)?

techsmyth:
@valentinyanakiev this was a tricky one. The core of what we want to check is that an admin can add a VC as a member when the VC is in the same account. Your suggestion re a separate privilege is I think the right path. I have added that privilege, specific to VC member addition - and also updated the logic in the mutation for adding a VC to a community role to change the privilege required based on whether the VC is in the same account. Please look at what has been done + see if you agree. 

Also when looking at it there was redundant rule being added related to invitation that could now be removed.

bobbykolev:
Hi @techsmyth, with this privilege in place the client now applies the following logic:
1. if COMMUNITY_ADD_MEMBER_VC_FROM_ACCOUNT or COMMUNITY_ADD_MEMBER is present - show both **Add** & **Add External** button (the second one showing msg for non-PA); 

If we're controlling the **Add** button only with COMMUNITY_ADD_MEMBER_VC_FROM_ACCOUNT, then it's hidden for PlatformAdmins. If we're controlling the **Add External** button only with COMMUNITY_ADD_MEMBER, then it's hidden for Space Admins. 

Let me know if you expect something different from the client, please.

valentinyanakiev:
> @valentinyanakiev this was a tricky one. The core of what we want to check is that an admin can add a VC as a member when the VC is in the same account. Your suggestion re a separate privilege is I think the right path. I have added that privilege, specific to VC member addition - and also updated the logic in the mutation for adding a VC to a community role to change the privilege required based on whether the VC is in the same account. Please look at what has been done + see if you agree.
> 
> Also when looking at it there was redundant rule being added related to invitation that could now be removed.

I looked at it, thought about it, looked at it again, thought again, and I don't have a better solution without massive refactoring. This looks odd, because there should be one privilege assigned to all that need it, and then one privilege checked, in my opinion (e.g. COMMUNITY_ADD_MEMBER_VC). But we can't really assign it that easily to PLATFORM_ADMINS with our current setup, so that should work, for the time being.

PR #4069, opened 2024-06-10
usage of space type to determine default contents of a space

techsmyth:
additional space type: virtual contributor

created defaults for different space types:
- root space
- challenge
- opportunity
- virtual contributor
- blank slate (requested by Denise)

for each space type, there is now a sub-folder with:
- callout groups to use
- default callouts to add
- states in the innovation flow
- settings

This PR also needed to fix a couple of places where we were using SpaceType when it should have been space level.

Creation of a Subspace now takes an optional parameter allowing to specify the type of the subspace to be created + from there the relevant defaults are picked up.  

The default library innovation flow is only picked up for now for challenge + opportunity types, to explicitly allow for picking up the states for VC. To be looked at better later. 

To check:
- search results are properly grouped / returned
- storage aggregators at the different space levels are properly returning

Separate comment: we need to tidy up usage of type on Community (should go)

**To test:**
- do a normal create subspace mutation via client
- do the same via graphql but with the type of VIRTUAL_CONTRIBUTOR specified:
- 
image
- check that the callouts are in the right state

Plus please explicitly check the search results related to subspaces.


techsmyth:
@valentinyanakiev I cannot test this out as local services are not starting up for me - can you please give it a try?

techsmyth:
@Comoque1 can you please try this one out?

valentinyanakiev (changes requested):
```gql
mutation createAccount($input: CreateAccountInput!) {
  createAccount(accountData: $input) {
    id
    spaceID
    __typename
  }
}
```
```json
{
  "input": {
    "hostID": "001bad6d-b200-4695-8997-459f0ca1d317",
    "spaceData": {
      "nameID": "test-r-s-1",
      "profileData": {
        "displayName": "test-r-s-1",
        "tagline": "",
        "location": {
          "city": "Pancharevo",
          "country": "AG"
        }
      },
      "tags": [
        "test-r-s-1"
      ]
    }
  }
}
```

"message":"Root space must have a type of SPACE: 'undefined'"

techsmyth:
@valentinyanakiev think I had missed pushing a couple of local commits. Just pushed. Will try myself in a moment.

valentinyanakiev:
> @valentinyanakiev think I had missed pushing a couple of local commits. Just pushed. Will try myself in a moment.

Still fails with the same error.

techsmyth:
@valentinyanakiev my mistake, had order wrong way around. Just tried flow myself now + it is working locally.

ccanos (approved):

valentinyanakiev (approved):

Comoque1:
Tested with all types on dev env.
The callouts are created successfully and visible on the corresponding flow states.

The search will be verified on ACC env.

There is something else to be considered:
 - when we have different from default innovation flow template, default states are still applied on challenge / opportunity

techsmyth:
@Comoque1 Can you please clarify the last point above, namely "There is something else to be considered:

when we have different from default innovation flow template, default states are still applied on challenge / opportunity"

This sounds like a bug...so it may need to be raised...

PR #4082, opened 2024-06-10
Admin gets COMMUNITY_ADD_MEMBER_VC_FROM_ACCOUNT

valentinyanakiev:
I am not finding a good way to do this. PLATFORM_ADMIN is not available on COMMUNITY, GRANT is sufficient for all admins and maybe all we needed is a policy rule with GRANT --> COMMUNITY_ADD_MEMBER_VC_FROM_ACCOUNT?

Personally, I couldn't find a better way.


techsmyth (approved):
If the only change is the privilege rule then I think that is fine: it is logical enough that being able to GRANT to a community also be able to ADD_VCs?

PR #4084, opened 2024-06-10
split org updates into regular admin updates + platform admin updates

techsmyth:
allow updating of nameID only for users with PLATFORM_ADMIN privilege


valentinyanakiev (changes requested):
I have created a separate module for all those platform settings under `src/platform/settings`. Haven't refactored it, but over time a lot of the code there can be refactored and re-used in that service with way less lines of code.

valentinyanakiev (approved):

Comoque1:
Tested OK on DEV env with GA and Support - can update.
Org Admin and Org Owner fails, as they don't have PLATFORM_ADMIN privilege.

PR #4092, opened 2024-06-12
Fixed missing save on update account platform settings

### LINKED ISSUE #4089: BUG: When UpdateAccountPlatformSettings mutation is triggered the changes are applied, but not reachable until policy reset
**Describe the bug**

**To Reproduce**
1. Go to local env
2. Create a space
3. For the space enable `Create template from callout` 

4. Check what the server will return: correct results are returned

5. Check from the client, that the option "Save as template" is available on a callout settings - it is not

**Expected behavior**
When UpdateAccountPlatformSettings, the changes are preserved and reachable from the client.

**Additional context**
Triggering policy reset resolves the issue.

valentinyanakiev:
Fixes #4089

### DIFF
--- a/src/domain/space/account/account.resolver.mutations.ts
+++ b/src/domain/space/account/account.resolver.mutations.ts
@@ -180,9 +180,9 @@ export class AccountResolverMutations {
     );
 
     // Update the authorization policy as most of the changes imply auth policy updates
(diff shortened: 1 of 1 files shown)


Comoque1 (approved):

Comoque1:
Tested OK locally.

PR #4101, opened 2024-06-17
Documents auth

hero101:
- added `bearer_token` in ory `authenticators`, which transforms a `session token` to JWT


valeksiev (approved):
Checked out locally and works nicely with the new ingestion service. cc @valentinyanakiev

valentinyanakiev:
> Checked out locally and works nicely with the new ingestion service. cc @valentinyanakiev

worked for me as well

PR #4117, opened 2024-06-20
Add whiteboard last saved date subscription

ccanos:
[# whiteboard-collaboration-service/8
](https://github.com/alkem-io/whiteboard-collaboration-service/issues/8)

Goes together with some client changes: https://github.com/alkem-io/client-web/pull/6435

### DIFF
--- a/src/domain/common/whiteboard/whiteboard.saved.resolver.subscription.ts
+++ b/src/domain/common/whiteboard/whiteboard.saved.resolver.subscription.ts
@@ -58,7 +58,9 @@ export class WhiteboardSavedResolverSubscription {
         `[User (${agentInfo.email}) Whiteboard Saved] - Filtering whiteboard id '${payload.whiteboardID}' - match=${isMatch}`,
         LogContext.SUBSCRIPTIONS
       );
-
+      
+      // Something is changing the Date from the payload into a string. GraphQL needs it to be a Date or a serialization error occurs
+      // So we do this conversion here
       payload.updatedDate = new Date(payload.updatedDate);
       return isMatch;
     },
--- a/src/domain/common/whiteboard/whiteboard.saved.resolver.subscription.ts
+++ b/src/domain/common/whiteboard/whiteboard.saved.resolver.subscription.ts
@@ -1,0 +1,88 @@
+import { Inject, LoggerService, UseGuards } from '@nestjs/common';
+import { Args, Resolver } from '@nestjs/graphql';
+import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import { CurrentUser } from '@common/decorators/current-user.decorator';
+import { AgentInfo } from '@core/authentication.agent.info/agent.info';
+import { GraphqlGuard } from '@core/authorization';
+import { LogContext } from '@common/enums/logging.context';
+import { AuthorizationService } from '@core/authorization/authorization.service';
+import { AuthorizationPrivilege } from '@common/enums/authorization.privilege';
+import { TypedSubscription } from '@src/common/decorators';
+import {
+  WhiteboardSavedSubscriptionArgs,
+  WhiteboardSavedSubscriptionResult,
+} from '@domain/common/whiteboard/dto/subscription/';
+import { SubscriptionReadService } from '@services/subscriptions/subscription-service';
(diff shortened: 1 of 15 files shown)


src/domain/common/whiteboard/whiteboard.saved.resolver.subscription.ts
```diff
@@ -0,0 +1,90 @@
...
+      const agentInfo = context.req?.user;
+      const isMatch = variables.whiteboardID === payload.whiteboardID;
+      this.logger.verbose?.(
+        `[User (${agentInfo.email}) Whiteboard Saved] - Filtering whiteboard id '${payload.whiteboardID}' - match=${isMatch}`,
+        LogContext.SUBSCRIPTIONS
+      );
+      
+      // Something is changing the Date from the payload into a string. GraphQL needs it to be a Date or a serialization error occurs
+      // So we do this conversion here
+      payload.updatedDate = new Date(payload.updatedDate);
```
hero101: move this to `resolve` instead

src/domain/common/whiteboard/whiteboard.saved.resolver.subscription.ts
```diff
@@ -0,0 +1,90 @@
...
+    const logMsgPrefix = `[User (${agentInfo.email}) Saved Whiteboard] - `;
+    this.logger.verbose?.(
+      `${logMsgPrefix} Subscribing to the following whiteboard: ${whiteboardID} for saved events`,
+      LogContext.SUBSCRIPTIONS
+    );
+
+    const whiteboard = await this.whiteboardService.getWhiteboardOrFail(
+      whiteboardID
+    );
+    await this.authorizationService.grantAccessOrFail(
```
hero101: this is not async

src/domain/common/whiteboard/whiteboard.service.ts
```diff
@@ -222,6 +224,11 @@ export class WhiteboardService {
 
     this.eventEmitter.emit(WHITEBOARD_CONTENT_UPDATE, savedWhiteboard.id);
 
+    this.subscriptionPublishService.publishWhiteboardSaved(
+      whiteboard.id,
+      new Date(Date.now())
```
hero101: The behavior you described in the PR description is misleading to the end user and can easily be avoided.
Persisting a large whiteboard is not ideal, so try to update only the update date, or if nothing else works, persist the whole whiteboard.
We don't want users to be afraid of using the feature, because the last save doesn't change consistently.

hero101 (changes requested):

src/domain/common/whiteboard/whiteboard.service.ts
ccanos: good good, just updating the `updateDate` manually works, so we have that weird behavior fixed now. I'll change the description of the PR, thanks.

src/domain/common/whiteboard/whiteboard.service.ts
```diff
@@ -217,11 +219,16 @@ export class WhiteboardService {
     );
 
     whiteboard.content = JSON.stringify(newContent);
-
+    whiteboard.updatedDate = new Date(); // Make sure that at least the updatedDate is updated (if the content is the same)
```
hero101: if the content is the same, the same whiteboard will be returned at `line 206`, without an update date, and no subscription will be triggered.
On the other hand, If you have reached this line, then the `updatedDate` will be updated regardlessly.

Try to set the `updateDate` in the database regardless of any whiteboard content changes.

hero101 (changes requested):

src/domain/common/whiteboard/whiteboard.service.ts
ccanos: The thing is, just a bit later we are doing JSON.parse and JSON.stringify to the whiteboard content so the whiteboard content is never the same. (stringify removes new line characters that excalidraw is sending), I've debugged it a bit and that `if` is always `false`, it was typeorm the one skipping the save, after the stringify. 
that's why I saw it working well while testing. 
no worries, I'll restructure a bit the code.

hero101 (approved):

ccanos:
please merge together with https://github.com/alkem-io/client-web/pull/6435 which I think it's ready

PR #4123, opened 2024-06-21
My invitations: also show relevant VC + Org invitations

techsmyth:
the me query fields related to community membership are now:
- communityApplications
- communityInvitations

The communityInvitations then returns invitations a) directly for the user b) for VCs on accounts hosted by the user c) on orgs owned by the user

Testing that has been done:
- able to retrieve accounts for users / orgs
- able to create an invitation to a community for a contributor
- that invitation is showing up in my list of invitations as I created the VC


src/services/infrastructure/url-generator/url.generator.service.ts
```diff
@@ -288,14 +296,39 @@ export class UrlGeneratorService {
...
+    let path = this.PATH_VIRTUAL_CONTRIBUTOR;
+    switch (type) {
+      case CommunityContributorType.USER:
+        path = this.PATH_USER;
+        break;
+      case CommunityContributorType.ORGANIZATION:
+        path = this.PATH_ORGANIZATION;
+        break;
+      case CommunityContributorType.VIRTUAL:
+        path = this.PATH_VIRTUAL_CONTRIBUTOR;
```
bobbykolev, line 310: already assigned

src/services/adapters/notification-adapter/notification.payload.builder.ts
```diff
@@ -506,51 +505,87 @@ export class NotificationPayloadBuilder {
...
+        });
+      }
+    } else {
+      // look up based on nameID
+      contributor = await this.entityManager.findOne(User, {
+        ...options,
+        where: { ...options?.where, nameID: contributorID },
+      });
+      if (!contributor) {
+        contributor = await this.entityManager.findOne(Organization, {
```
bobbykolev, line 568: Do we have unique ids/nameIDs for all entities? Is it possible a name ID of Org to match the nameID of VC but the logic to retrieve the ORG first even if you're looking for the VC?

src/services/infrastructure/user-lookup/user.lookup.service.ts
```diff
@@ -32,4 +39,107 @@ export class UserLookupService {
...
+      },
+    });
+    if (organizations.length > 0) {
+      contributorsManagedByUser.push(...organizations);
+    }
+
+    return contributorsManagedByUser;
+  }
+
+  private async getContributorsManagedByAccount(
```
bobbykolev: This should be getVirtualContributorsManagedByAccount as it queries only this type of contributors.

bobbykolev (dismissed):
A couple of comments / questions were added. Feel free to merge and if something is required we can add it later.

valentinyanakiev:
Fixed the test mocks, CI needs to be kept green.

src/services/infrastructure/url-generator/url.generator.service.ts
techsmyth: yes, but better to also have here for completeness - not just rely on the original assignment. The other option would be make the original assignment an empty string.

src/services/adapters/notification-adapter/notification.payload.builder.ts
techsmyth: The users, orgs + VCs are all in different namespaces i.e. you could have a user with nameID "john" and also an org and a vc with the same nameID.

So yes this logic has an issue. Can you please raise a separate issue for this one?

src/services/infrastructure/user-lookup/user.lookup.service.ts
techsmyth: Agreed.

src/services/adapters/notification-adapter/notification.payload.builder.ts
valentinyanakiev: You are right, this is potentially error-prone. If we have baseContributor, basically nameID should be in its scope, and it is not at the moment.

src/services/infrastructure/user-lookup/user.lookup.service.ts
valentinyanakiev: Addressed this.

src/services/adapters/notification-adapter/notification.payload.builder.ts
valentinyanakiev: Added a toDo - https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/server/4126 - to fix, and referenced it in the code.

PR #4125, opened 2024-06-22
LicensePolicy with credentialRules; FeatureFlags migrated

techsmyth:
Add new License Plans for the features that used to be feature flags

Migrate data

Replace previous featureFlagRules on LicensePolicy with credentialRules

LicensePlan extended to allow specifying per rule if it should be assigned to new org or user accounts, replacing previous usage of a "basePlan".


src/common/enums/license.credential.ts
```diff
@@ -6,6 +6,9 @@ export enum LicenseCredential {
   LICENSE_SPACE_PLUS = 'license-space-plus',
   LICENSE_SPACE_PREMIUM = 'license-space-premium',
   LICENSE_SPACE_ENTERPRISE = 'license-space-enterprise',
+  FEATURE_CALLOUT_TO_CALLOUT_TEMPLATE = 'feature-callout-to-callout-template',
```
me-andre, line 9: It seems to me those credentials are of different type/purpose and would never sit together in one field.
It could be better if License types and Features were different types.

techsmyth: This is key to understanding credential based frameworks: This is just the name of a credential, the usage is separate. They must be in the same list, as this is the set of credentials understood by the license engine. I can imagine multiple other types of license related credentials in this list, each of which gets interpreted different.

me-andre (approved):

PR #4159, opened 2024-06-26
moved external invitations to be under platform and framed as platform invitations

techsmyth:
Added new mutation to allow creating a platform invitation for a platform role

Added a new platform role: VC Campaign


src/migrations/1719431685862-platformInvitations.ts
```diff
@@ -0,0 +1,29 @@
...
+export class platformInvitations1719431685862 implements MigrationInterface {
+  name = 'platformInvitations1719431685862';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE invitation_external RENAME TO platform_invitation`
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE \`platform_invitation\` ADD \`platformId\` varchar(36) NULL`
```
valentinyanakiev:
```suggestion
      `ALTER TABLE \`platform_invitation\` ADD \`platformId\` char(36) NULL`
```

valentinyanakiev (changes requested):
One type comment in the migration, yet to run it end-to-end.

valentinyanakiev (approved):

PR #4162, opened 2024-06-27
await saving of account

techsmyth:


valentinyanakiev (approved):
Both approaches are actually the same code with different coding style, it is down to coding style preferences / standards only:

From ChatGPT:

In terms of performance, the difference between the two approaches is negligible. Both async/await and promise chaining (with .then) achieve the same end result and run in effectively the same time, as they are both built on JavaScript's Promise system. The choice between them typically comes down to code readability and maintainability rather than execution speed.

However, to compare them again briefly:

First Snippet (Mix of async/await and .then):
```typescript
const account = await this.accountService.createAccount(
  accountData,
  agentInfo
);

const accountUpdated = await this.accountAuthorizationService
  .applyAuthorizationPolicy(account)
  .then(account => this.accountService.save(account));
const space = await this.accountService.getRootSpace(accountUpdated);

await this.namingReporter.createOrUpdateName(
  space.id,
  space.profile.displayName
);
return accountUpdated;
```

Second Snippet (Pure async/await):
```typescript
let account = await this.accountService.createAccount(
  accountData,
  agentInfo
);

account = await this.accountAuthorizationService.applyAuthorizationPolicy(
  account
);
account = await this.accountService.save(account);
const space = await this.accountService.getRootSpace(account);

await this.namingReporter.createOrUpdateName(
  space.id,
  space.profile.displayName
);
return account;
```

Analysis
Similarities
Purpose: Both snippets create an account, apply authorization policies, save the updated account, retrieve the root space, and update the name using namingReporter.
Sequence: The sequence of operations is identical in both snippets: create account -> apply authorization -> save account -> get root space -> update name.
Differences
Variable Assignment:

In the first snippet, account is defined as a constant const and a new variable accountUpdated is introduced to hold the result after applying the authorization policy and saving the account.
In the second snippet, account is defined with let and is reassigned after each step (applying authorization and saving).
Usage of .then:

The first snippet uses a .then chain to save the account after applying the authorization policy. This approach is a mix of async/await and promise chaining.
The second snippet uses only async/await syntax without mixing in promise chaining.
Conclusion
Both snippets perform the same operations and achieve the same result. The choice between them can come down to coding style preference:

First Snippet: Might be preferred if you like using .then chains for subsequent operations after an async call.
Second Snippet: Might be preferred for consistency if you want to stick with async/await syntax throughout the function.
Both approaches are correct and will result in the same functionality, so it ultimately depends on your or your team's coding style preference.

Performance Considerations
Async/Await: This syntax is generally more readable and easier to understand, especially when dealing with multiple sequential asynchronous operations. It helps in writing code that looks synchronous, thus reducing cognitive load.

Promise Chaining with .then: While this is also a valid approach, it can sometimes be harder to read, especially with deeply nested asynchronous operations.

Execution Speed
Both approaches use promises under the hood and will be executed in the same way by the JavaScript engine. The overhead introduced by using .then is minimal and generally not significant enough to impact performance in a noticeable way for most applications.

Conclusion
Readability: The second snippet (pure async/await) is often preferred for readability and maintainability, especially in modern JavaScript codebases.
Performance: There is no significant performance difference between the two approaches. Any difference would be so small that it would typically not matter in real-world applications.
Given these points, you should choose the approach that you and your team find more readable and maintainable. For most developers, this would be the second snippet using pure async/await.

PR #4163, opened 2024-06-27
platform invitations naming; community new member for contributors

techsmyth:
New member of a space is not generic and works for virtual contributors as well as users:
image

Community invitations for VCs also now send over the message that was written in the invitaiton:
image

Code naming + module dependencies cleaned up to avoid duplicate code. In particular the UserLookupModule is now ContributorLookupModule + has the generic contributor lookups, which was needed for notificaitons. 

VirtualContributor is now able to retrieve the host contributor.


valentinyanakiev (approved):
Works for me locally, tested with VCs and normal members.

PR #4168, opened 2024-06-28
Simplify account + root space creation, logic around auth

techsmyth:
account auth reset no longer uses space authorization

account save no longer cascades to space save (maybe this was the bug)

generate host credentials on account host service

split creation of account from creation of root space;

 reduced knowledge of host in VC; 

simplified VC rules; removed root space cascade; 

added storage aggregator to account

Migration to add storage aggregator to existing accounts + set the root spaces storage aggregator to be that of the account

Space auth reset logic is that it looks after everything **except** for subspaces and then saves. And only then does it cascade so subspaces - as each subspace has the responsibility to update + save itself.

Simple account creation, with root space, is working locally.


Comoque1:
UpdateAccountPlatformSettings: fails:       "message": "Cannot read properties of undefined (reading 'privacy')",
image

Comoque1:
When create a space - account is created successfully, but fails "MyAccount" query: `"message": "AuthorizationPolicy without credential rules provided: 071735ed-0b12-4a46-9880-725598107fe6",`
image

Comoque1:
Create Subspace fails: `"message": "Cannot read properties of undefined (reading 'privacy')",`
image

valentinyanakiev:
> UpdateAccountPlatformSettings

I think I fixed this.

valentinyanakiev:
> Create Subspace fails: `"message": "Cannot read properties of undefined (reading 'privacy')",` image

To narrow it down 
```
query MyAccount {
  me {
    spaceMemberships {
      id
      profile {
        id
        displayName
        tagline
        url
        __typename
      }
    }
  }
}
```

It is the `profile` that fails.

valentinyanakiev:
```
mutation createSubspace($input: CreateSubspaceInput!) {
  createSubspace(subspaceData: $input) {
    ...SubspaceCard
    __typename
  }
}

fragment SubspaceCard on Space {
  id
  authorization {
    id
    anonymousReadAccess
    __typename
  }
  metrics {
    id
    name
    value
    __typename
  }
  profile {
    id
    url
    tagline
    displayName
    description
    cardBanner: visual(type: CARD) {
      ...VisualUri
      __typename
    }
    tagset {
      ...TagsetDetails
      __typename
    }
    url
    __typename
  }
  context {
    id
    vision
    __typename
  }
  community {
    id
    myMembershipStatus
    __typename
  }
  __typename
}

fragment VisualUri on Visual {
  id
  uri
  name
  __typename
}

fragment TagsetDetails on Tagset {
  id
  name
  tags
  allowedValues
  type
  __typename
}
```

```
{
  "input": {
    "spaceID": "eco1",
    "context": {
      "vision": "test-subspace\n"
    },
    "profileData": {
      "displayName": "test-subspace",
      "tagline": "test-subspace",
      "description": "test-subspace\n"
    },
    "tags": [],
    "collaborationData": {
      "addDefaultCallouts": true
    }
  }
}
``` 

fails

valentinyanakiev (approved):

src/core/bootstrap/bootstrap.service.ts
```diff
@@ -292,12 +293,20 @@ export class BootstrapService {
...
         hostID: DEFAULT_HOST_ORG_NAMEID,
       };
-      return this.accountService
-        .createAccount(spaceInput)
-        .then(account =>
-          this.accountAuthorizationService.applyAuthorizationPolicy(account)
-        )
-        .then(account => this.accountService.save(account));
+
+      let account = await this.accountService.createAccount(spaceInput);
```
hero101, line 297: here you are passing `NAMEID` to the `hostID` field, which cause the code at `account.service.ts:102` to fail, because it`s trying to read a contributor specifically by UUID, but you are passing a NAMEID instead.

**In cases where you enforce a certain format to the data, you need to validate it's the proper format.**

**This changeset breaks the bootstrap process.**

PR #4180, opened 2024-06-28
filter also by public status on VC

techsmyth:


Comoque1:
Tested OK locally:
image

Comoque1 (approved):

Issue #4176, opened 2024-06-28
BUG: space admin doesn't have rights to import / CRUD ANY templates on a space


Comoque1:
**Describe the bug**

**To Reproduce**
1. Go to any env
2. Create a space with GA
3. Assign clean user as admin to the space
4. Using the assigned user - try to create post template or to import a post template - auth error is thrown
<img width="1529" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNG1lQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5520b0e64de82c45d29491a292cb0d52459b3005/image.png" alt="image.png" />

**Expected behavior**
Space admin doesn't have rights to import / create ANY templates on a space

PR #4181, opened 2024-06-28
updated vc removal from community to extend auth policy with account host credentials in memory

techsmyth:
This PR enables the mutation on the server to run,. key here is that the authorization policy has to be extended in memory, based on the information from the incoming request. This is similar to how user self removal works.

In addition the authorization policy around account / space has been tidied up, with the result that the VC provider will now have the GRANT privilege + so will see the button to leave. 

Other updates:
removed CREATE_DISCUSSION privilege assignment from community; 
trigger authorization reset after assign license plan to account; 
removed spaceIngest mutation that logically should not be there


Comoque1:
The PR doesn't resolve leaving community.

I also Identified that host / VC provider doesn't have CRUD rights to own VC
After policy reset is the same
image

techsmyth:
@Comoque1 "I also Identified that host / VC provider doesn't have CRUD rights to own VC" ==> this was resolved with #4176

PR #4182, opened 2024-06-28
ensure root space admins can work with child entities

techsmyth:
Added rule to give all space admins CRUD on library, VCs, defaults.

Screen shot below after auth reset for a space admin on the library authorizaiton:

image


Comoque1:
Callout api template tests are passing with this PR.
image

I would rather leave the PR review to Valentin.

PR #4189, opened 2024-07-01
virtual receives auth extended for child entities

techsmyth:


src/domain/space/account/account.resolver.mutations.ts
```diff
@@ -279,10 +279,16 @@ export class AccountResolverMutations {
       virtualContributorData
     );
 
+    const clonedAccountAuth =
+      await this.accountAuthorizationService.getClonedAccountAuthExtendedForChildEntities(
+        virtual.account
+      );
+
+    // Need
```
me-andre, line 287: Can we have a better comment here if needed?

me-andre (changes requested):
Just one minor comment about a comment - it doesn't seem informational to me

PR #4197, opened 2024-07-01
remove type on community

techsmyth:
Clears up some very old, legacy code

Also cleaned up some legacy lookups in activity dapter


hero101 (approved):

hero101:
Tested locally.
Ran the `journey` suit. Only one test fails, but most probably is not caused by this change
image

PR #4204, opened 2024-07-02
updated credentials to accept invitations to include host creds when org

techsmyth:


bobbykolev (approved):
Tested locally, as admin in org, I manged to accept the invite to a VC with org host.

PR #4221, opened 2024-07-07
Detecting of replies to VC mentions

techsmyth:
Detects the first VC that was mentioned in a thread, and sends a follow up question to that VC

Clean up of logic

Moved mentions to a separate service.

Working from the room service point of view, to be cascaded to AI server + into expert engine.


valentinyanakiev (approved):
Tested OK locally

PR #4226, opened 2024-07-08
add myRoles on platform

techsmyth:
split out platform roles management into a separate module to keep the logic separate


bobbykolev (approved):
`myRoles` works as expected returning the platform roles assigned to the user.
It would be used to control the My Account block (not only) based on the VC Campaign role. 

The assignment of platform roles through the administration was tested as part of the PR. 

```
query platform {
  platform {
    myRoles
  }
}

PR #4233, opened 2024-07-08
allow callouts to have the same displayName

techsmyth:


Comoque1:
There is still validation on client level, so it must be removed as well in order the purpose of this change o work:
image

Comoque1 (approved):

Issue #4223, opened 2024-07-08
BUG: When you invite a VC to subspace and it's already part of the parent community the server throws an error


bobbykolev:
**Describe the bug**
Invite an external VC to a subspace, then invite the same vc to another subspace of the same space. The server throws an error and the user is unable to accept the invite.

**To Reproduce**
1. Invite externla VC to subspace.
2. Accept this invite (should be successful and the VC is now part of the space and subspace).
3. Invite the same VC to another subspace from the same space.
4. Try to accept the invte
4. See the error (server)

**Expected behavior**
The host of a VC should be able to accept all invites and a VC should be able to be part of multiple subspaces under the same space.

**Screenshots**
<img width="1222" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOEs4QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7c940640fed67bcc56d6f159cedd9131942f31c0/afbeelding%20%281%29.png" alt="afbeelding (1).png" />

**Additional context**

Local server logs. It looks like the logic for assiging the VC to the parent space is failing as it's already part of it.

```

C:\Users\Vision\Downloads\work\Alkemio\project\server\src\common\exceptions\user.input.error.ts:6   
    super(message, {
    ^
GraphQLError: Agent (virtual-vctoinvite) already has assigned credential: space-member
    at new UserInputError (C:\Users\Vision\Downloads\work\Alkemio\project\server\src\common\exceptions\user.input.error.ts:6:5)
    at new ValidationException (C:\Users\Vision\Downloads\work\Alkemio\project\server\src\common\exceptions\validation.exception.ts:7:5)
    at AgentService.grantCredential (C:\Users\Vision\Downloads\work\Alkemio\project\server\src\domain\agent\agent\agent.service.ts:193:15)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at CommunityService.assignContributorAgentToRole (C:\Users\Vision\Downloads\work\Alkemio\project\server\src\domain\community\community\community.service.ts:1071:12)
    at CommunityService.assignVirtualToRole (C:\Users\Vision\Downloads\work\Alkemio\project\server\src\domain\community\community\community.service.ts:710:32)
[...]

PR #4238, opened 2024-07-09
Prevent an error on accepting VC invites

bobbykolev:
https://github.com/alkem-io/server/issues/4223

check whether the VC already has the role before assigning it


techsmyth (approved):
Indeed, nice fix.

PR #4273, opened 2024-07-17
Fix 4251: Get my created spaces in a proper way

ccanos:
I've just removed the param `showOnlyMyCreatedSpaces` and have added a field to the `me` query that returns directly my created spaces instead of passing through activities.

### DIFF
--- a/src/domain/space/space/space.service.ts
+++ b/src/domain/space/space/space.service.ts
@@ -1173,7 +1173,7 @@ export class SpaceService {
 
   public async getSpaceForAccountOrFail(accountId: string): Promise<ISpace> {
     const space = await this.spaceRepository.findOne({
-      where: { account: { id: accountId } },
+      where: { level: 0, account: { id: accountId } },
       relations: {
         account: true,
       },
--- a/src/domain/space/space/space.service.ts
+++ b/src/domain/space/space/space.service.ts
@@ -1171,6 +1171,24 @@ export class SpaceService {
     return space.account;
   }
 
+  public async getSpaceForAccountOrFail(accountId: string): Promise<ISpace> {
+    const space = await this.spaceRepository.findOne({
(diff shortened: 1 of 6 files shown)


src/services/api/me/me.service.ts
```diff
@@ -142,40 +144,33 @@ export class MeService {
...
 
+  public async getMyCreatedSpaces(agentInfo: AgentInfo): Promise<ISpace[]> {
+    const user = await this.userService.getUserOrFail(agentInfo.userID);
+    if (!user) {
+      throw new EntityNotFoundException(
+        `User not found ${agentInfo.userID}`,
+        LogContext.COMMUNITY
+      );
+    }
+    const accounts =
```
bobbykolev: In case of no accounts hosted by the user, can we just make sure the server returns an empty array?

src/domain/space/space/space.service.ts
```diff
@@ -1171,6 +1171,24 @@ export class SpaceService {
     return space.account;
   }
 
+  public async getSpaceForAccountOrFail(accountId: string): Promise<ISpace> {
+    const space = await this.spaceRepository.findOne({
```
bobbykolev: Currently, every account has a spaceID. So if we have the account then it would be more optimal to directly use its spaceID instead of iterating all the spaces.

src/services/api/me/me.service.ts
ccanos: yeah, accounts was already an empty array, but now `compact` will make sure of that

src/domain/space/space/space.service.ts
ccanos: I have removed this one, good idea. Just passing `includeSpace` to getAccountsHostedByContributor is going to make it more efficient

bobbykolev (approved):

PR #4283, opened 2024-07-19
Rename interactions to vcInteractions in resolvers

valeksiev:
Needed for https://github.com/alkem-io/client-web/issues/6520


techsmyth (approved):
Made a couple of simplifications:
- did not need field decorator on interface as have a field resolver 
- retrieve the room as part of typeorm loading instead of setting explicitly (cleaner)

Only other thought is if it helps to add a field resolver for the contributor instead of just returning the UUID?

But suggest this one goes in as is if you are ok with the changes in the last two commits.

valeksiev:
I was having trouble with the decorators/resolver functions and was not sure which one we prefer.

Re the room - didn’t like the _circular_ loading but agreed it’s cleaner. 

VC uuid works fine for now and if we want to change it I’d suggest a bigger change but it’s outside the scope of this PR for sure.

I think you can go ahead and merge it.

techsmyth:
@valeksiev thanks for comments. Actually now this is merged I presume the client also needs merging?

valeksiev:
I need to address some comments on the client first - either today or tomorrow.

techsmyth:
Great, Monday of course also fine - just team needs to be aware that client is out of sync with server.

PR #4284, opened 2024-07-19
Ory Kratos updated to v1.2; Session Refresh fixed by proxying fresh 🍪

me-andre:
…ed cookies

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/core/authentication/authentication.service.ts
+++ b/src/core/authentication/authentication.service.ts
@@ -18,12 +18,14 @@ import { SessionExtendException } from '@common/exceptions/auth';
 
 import { AgentInfoCacheService } from '@core/authentication.agent.info/agent.info.cache.service';
 import { getSession } from '@common/utils';
+import ConfigUtils from '@config/config.utils';
+
 @Injectable()
 export class AuthenticationService {
   private readonly kratosPublicUrlServer: string;
   private readonly adminPasswordIdentifier: string;
   private readonly adminPassword: string;
-  private readonly extendAfter: number;
+  private readonly extendSessionMinRemainingTTL: number | undefined; // min time before session expires when it's already allowed to be extended (in milliseconds)
   private readonly kratosIdentityClient: IdentityApi;
   private readonly kratosFrontEndClient: FrontendApi;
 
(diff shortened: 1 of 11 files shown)


src/core/authentication/authentication.service.ts
```diff
@@ -176,48 +177,78 @@ export class AuthenticationService {
...
+      return true; // Set to -1 if specified as lifespan in config, meaning it can be extended at any time
+    }
 
     const expiry = new Date(session.expires_at);
-    return Date.now() >= expiry.getTime() - this.extendAfter;
+    return Date.now() >= expiry.getTime() - this.extendSessionMinRemainingTTL;
+  }
+
+  private parseEarliestPossibleExtend(
+    earliestPossibleExtend: unknown
```
hero101, line 236: can you narrow down the type?

src/core/middleware/session.extend.middleware.ts
```diff
@@ -98,24 +87,12 @@ export class SessionExtendMiddleware implements NestMiddleware {
...
-        return next();
-      }
-
-      res.cookie(this.SESSION_COOKIE_NAME, orySessionId, {
-        path: '/',
-        sameSite: 'lax',
-        expires: new Date(newExpiry),
-        httpOnly: true,
-        encode: val => val, // skip encoding; pass the value as is
+      const { headers } = await this.kratosClient.toSession({
```
me-andre, line 98: ⚠️Add a comment why proxying headers is needed and we can't just extend the old cookie.

src/core/middleware/session.extend.middleware.ts
```diff
@@ -58,9 +58,7 @@ export class SessionExtendMiddleware implements NestMiddleware {
 
     let session: Session;
     try {
-      session = await getSession(this.kratosClient, {
-        authorization,
-      });
+      session = getSessionFromJwt(authorization);
```
hero101, line 61: this function requires the token, and you are passing the whole value of the header.
`getSession` was able to parse the header, but this one doesn't

hero101 (changes requested):
Will need to test locally.

src/core/authentication/authentication.service.ts
me-andre: It is narrowed down already (config gives `any`). From here on we operate on the narrowest possible type (hence the `typeof` below).

src/core/middleware/session.extend.middleware.ts
hero101: Turns out that it actually can parse the whole header.

src/core/middleware/session.extend.middleware.ts
```diff
@@ -70,22 +68,13 @@ export class SessionExtendMiddleware implements NestMiddleware {
...
-        });
-
-      if (!newSession) {
-        this.logger.warn?.(
-          'New session cookie not sent: new session not found',
-          LogContext.AUTH_TOKEN
+      try {
+        await this.authService.extendSession(session);
+      } catch (error) {
+        this.logger.error(
```
hero101, line 74: when this fails the code continues and sends a cookie everytime

src/core/authentication/authentication.service.ts
```diff
@@ -176,48 +177,78 @@ export class AuthenticationService {
...
+  ): number | undefined {
+    /**
+     * If you need high flexibility when extending sessions, you can set earliest_possible_extend to lifespan,
```
hero101, line 239: I believe this must be a description to the variable in the config.

hero101 (approved):

src/core/authentication/authentication.service.ts
me-andre: done

PR #4290, opened 2024-07-20
Drop transferring of space settings via community policy

techsmyth:
This was always not a nice solution, and had lead to some subtle authorization reset errors in the past.

It is also important to drop this link in order to make the communityPolicy module more generically a role manager module, so this is a step in that direction. 

No api or external changes, pure business logic refactoring

Risk area: authorization reset

To test: please do an authorization reset on all accounts on production data.


valentinyanakiev (approved):
Code looks good and I didn't see an obvious regression when doing auth reset on account. Community policies seemed to be working as well.

PR #4296, opened 2024-07-23
Resolve room in the the room event subscription result and forward replies to threads on posts to the VC service

valeksiev:
fixes:
https://github.com/alkem-io/client-web/issues/6649
https://github.com/alkem-io/client-web/issues/6650

### DIFF
--- a/src/services/subscriptions/subscription-service/subscription.publish.service.ts
+++ b/src/services/subscriptions/subscription-service/subscription.publish.service.ts
@@ -15,7 +15,7 @@ import {
   RoomEventSubscriptionPayload,
   WhiteboardSavedSubscriptionPayload,
 } from './dto';
-import { Room } from '@domain/communication/room';
+import { IRoom } from '@domain/communication/room/room.interface';
 
 @Injectable()
 export class SubscriptionPublishService {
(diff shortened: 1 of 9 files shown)


src/services/subscriptions/subscription-service/subscription.publish.service.ts
```diff
@@ -47,11 +48,13 @@ export class SubscriptionPublishService {
...
     type: MutationType,
     data: IMessage | IMessageReaction,
-    messageID?: string
+    messageID?: string,
+    room?: Room
   ): Promise<void> {
     const payload: RoomEventSubscriptionPayload = {
       eventID: `room-event-${randomInt()}`,
       roomID: roomId,
+      room: room || {},
```
hero101: I am not convinsed this won't throw if it's an empty object. Why is `room` optional to begin with - it's mendatory in `RoomEventSubscriptionResult`

hero101 (changes requested):

src/services/subscriptions/subscription-service/subscription.publish.service.ts
valeksiev: I will update the signature.

hero101 (approved):

PR #4298, opened 2024-07-23
Move InnovationHubs under account

techsmyth:
InnovationHub now under accounts, being created / deleted + auth reset there

Additional fields on InnovationHub to match VC/InnovationPack

Lookup of InnovationHub now via UUID where feasible. 

Added field on lookup for InnovationHub

Platform.InnovationHub field now only returns the current innovationHub.

Migration added to add extra felds

Todo:
- check database constraints after the migration


ccanos (approved):

src/library/library/library.service.authorization.ts
```diff
@@ -4,7 +4,7 @@ import { ILibrary } from './library.interface';
 import { IAuthorizationPolicy } from '@domain/common/authorization-policy';
 import { LibraryService } from './library.service';
 
-@Injectable()
+Injectable();
```
hero101, line 7: This change removes the class from the DI tree. As a result dependencies to and from the class won't be resolved, breaking the code.

PR #4303, opened 2024-07-24
Account-Space: all spaces to directly know L0 space ID, not rely on account

techsmyth:
A new field so that all spaces can directly know the L0 space ID without having to go via account.

This update is required to be able to move to having multiple spaces per account, which in turn is needed to be able to have a single account per user / org. 

It should also be a little bit more efficient in terms of lookup of levelzerospaceID for reporting.


valentinyanakiev (approved):
Tested with real data, migration passed, auth reset passed, L0, L1, L2 spaces loaded + callouts loaded.

I have done only exploratory testing, and as a pattern I would do only that for merging to the feature branch. Before we merge the feature branch I will run all test suites to find potential regressions.

PR #4309, opened 2024-07-24
add simple check for user being host of an account before deleting

techsmyth:
to be revisited later when we give every user an account but for now this will prevent this problem


Comoque1:
Tested OK locally with an API test
image

Comoque1 (approved):

PR #4326, opened 2024-07-26
Safe migration for new db

valentinyanakiev:
- refactored migration to be more readable
- lookup of default accountHost only in case there are innovationPacks to be migrated
- extended accountHost lookup to search 1. For `Alkemio` org, 2. For any Org that is accountHost, in case there is data to be migrated

Tested locally, creates DB from scratch. We can NOT rely that a certain dataset exists in a migration, unless we have just created it.


ccanos (approved):
tested on acc db, with and without an Alkemio Org, and on an empty db

PR #4335, opened 2024-07-31
Template set fixes

ccanos:
Fixes two errors on the InnovationHubs and InnovationPacks 
[BUG: Creating WB template on existing innovation pack throws an error #6691](https://github.com/alkem-io/client-web/issues/6691)
[BUG: When try to create innovation hub, error is thrown #6692](https://github.com/alkem-io/client-web/issues/6692)


src/domain/innovation-hub/innovation.hub.service.ts
```diff
@@ -73,11 +73,19 @@ export class InnovationHubService {
     hub.authorization = new AuthorizationPolicy();
     hub.listedInStore = true;
     hub.searchVisibility = SearchVisibility.ACCOUNT;
+    hub.account = account;
+
+    if (!account.storageAggregator) {
```
hero101: i would fail sooner, no point of doing all that extra work if you already know it will fail.

src/services/infrastructure/storage-aggregator-resolver/storage.aggregator.resolver.service.ts
```diff
@@ -122,12 +113,16 @@ export class StorageAggregatorResolverService {
       return await this.getStorageAggregatorOrFail(space.storageAggregator.id);
     }
 
-    const query = `SELECT \`id\` FROM \`innovation_pack\`
+    const query = `SELECT \`account\`.\`storageAggregatorId\` FROM \`innovation_pack\`
```
hero101: could you please add some comments, describing what and why this happens.
Why don't you use directly the aggregator of the account?

hero101 (changes requested):

hero101 (approved):

PR #4336, opened 2024-08-02
Configuration service with typings

hero101:
### Description
- The `AlkemioConfig` is a Typescript type dublicating the `alkemio.yml`. It is used to provide strict typing to the config and a dot notation to the `get` method. 
- All the usages in the code base are updated.
- Added an entry for graphql upload max file size, as it was previously missing.

#### Pros
- all the fields and their types are well-known

#### Cons
- `alkemio.yml` and `AlkemioConfig` have to be in sync.

---

### Old style
```typescript
constructor(private configService: ConfigService) {
    this.kratosPublicUrlServer = this.configService.get(
      ConfigurationTypes.IDENTITY
    ).authentication.providers.ory.kratos_public_base_url_server;
}
```

### New style
```typescript
constructor(private configService: ConfigService<AlkemioConfig, true>) {
    this.kratosPublicUrlServer = this.configService.get('identity.authentication.providers.ory.kratos_public_base_url_server', { infer: true });
}
```


alkemio.yml
```diff
@@ -278,6 +278,9 @@ communications:
 ## storage ##
 # Alkemio uses multiple types of persistent storage, including SQL database and IPFS.
 storage:
+  graphql_upload:
+    # 20MB
+    max_file_size: 1048576
```
valentinyanakiev: 1048576 is 10, not 20
```suggestion
    max_file_size: 20971520
```

valentinyanakiev (changes requested):
Overall application seems to be working from the client, but:
- default max file size should be double on what it is set here
- the full config schema should be accessible from the `configuration` api, and currently it returns errors, e.g.

```gql
query configuration {
  platform {
    configuration {
      storage {
        file {
          mimeTypes
          maxFileSize
        }
      }
    }
  }
}
```

I suggest testing whether the typings work with the GQL types when accessing the full configuration in gql, otherwise this won't work in some cases.

valentinyanakiev (commented):
I also see { infer: true } everywhere, ideally this will be a default and not passed on every usage.

valentinyanakiev (approved):

PR #4370, opened 2024-08-07
Organization AccountTab - make spaceId nullable

ccanos:
Part of: https://github.com/alkem-io/client-web/issues/6440
and linked with: https://github.com/alkem-io/client-web/pull/6723

### DIFF
--- a/src/domain/space/account/account.resolver.mutations.ts
+++ b/src/domain/space/account/account.resolver.mutations.ts
@@ -106,6 +106,12 @@ export class AccountResolverMutations {
         community: true,
       },
     });
+    if (!rootSpace) {
+      throw new EntityNotInitializedException(
+        `Unable to load root space for account ${account.id}`,
+        LogContext.ACCOUNT
+      );
+    }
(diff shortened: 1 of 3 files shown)


src/domain/space/account/account.resolver.mutations.ts
```diff
@@ -106,6 +106,12 @@ export class AccountResolverMutations {
         community: true,
       },
     });
+    if (!rootSpace) {
```
bobbykolev, line 109: If we're supporting Accounts without spaceId, do we need to throw an error here?

bobbykolev (approved):

src/domain/space/account/account.resolver.mutations.ts
ccanos: yes, we have just created a space a few lines ago, this should return the space for sure, but now that `getRootSpace` was nullable typescript was complaining. 
But this code is gone in the Neil's PR. I'm going to merge it for now.

PR #4372, opened 2024-08-07
added type field to agent; removed parentDisplayID field from Agent

techsmyth:
Addressing some technical debt, makes tracking of issues easier.


valentinyanakiev (changes requested):
This is still WIP, the definitions of the entity don't match the DB schema after running the migration, and even bootstrap fails.

techsmyth:
@valentinyanakiev somehow I missed deleting the field in what I checked in, my bad. Update column definition to match the migration, good catch.

valentinyanakiev (approved):

PR #4375, opened 2024-08-07
added type to storage aggregator, with migration

techsmyth:
Type addition to make it easier to track down storage related errors.

Code also uses the type to correctly return the parent entity information, especially for organizations / users

### DIFF
--- a/src/migrations/1723036109999-storageAggregatorType.ts
+++ b/src/migrations/1723036109999-storageAggregatorType.ts
@@ -1,0 +1,61 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class StorageAggregatorType1723036109999 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // Add the type column to the storage aggregator table
+    await queryRunner.query(
+      'ALTER TABLE `storage_aggregator` ADD `type` varchar(36) NULL'
+    );
+    await this.updateStorageAggregatorTypeForEntity(
+      queryRunner,
+      'account',
+      'account'
+    );
+    await this.updateStorageAggregatorTypeForEntity(
+      queryRunner,
+      'space',
+      'space'
+    );
+    await this.updateStorageAggregatorTypeForEntity(
+      queryRunner,
+      'user',
+      'user'
+    );
+    await this.updateStorageAggregatorTypeForEntity(
+      queryRunner,
+      'organization',
+      'organization'
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {}
+
+  private async updateStorageAggregatorTypeForEntity(
+    queryRunner: QueryRunner,
+    entityType: string,
+    storageAggregatorType: string
+  ) {
+    const entities: {
+      id: string;
+      storageAggregatorId: string;
+    }[] = await queryRunner.query(
+      `SELECT id, storageAggregatorId FROM \`${entityType}\``
+    );
+    for (const entity of entities) {
+      const [storageAggregator]: {
+        id: string;
+      }[] = await queryRunner.query(
+        `SELECT id FROM storage_aggregator WHERE id = '${entity.storageAggregatorId}'`
+      );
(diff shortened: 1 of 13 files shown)


src/services/infrastructure/storage-aggregator-resolver/storage.aggregator.resolver.service.ts
```diff
@@ -62,18 +66,32 @@ export class StorageAggregatorResolverService {
...
+  ): Promise<IAccount> {
+    const account = await this.entityManager.findOne(Account, {
+      where: {
+        storageAggregator: {
+          id: storageAggregator.id,
+        },
+      },
+    });
+    if (!account) {
+      throw new NotImplementedException(
```
valentinyanakiev: this is not `NotImplementedException`, more of an `EntityNotFoundException`

src/services/infrastructure/storage-aggregator-resolver/storage.aggregator.resolver.service.ts
```diff
@@ -82,17 +100,55 @@ export class StorageAggregatorResolverService {
     });
     if (!space) {
       throw new NotImplementedException(
```
valentinyanakiev: this is not NotImplementedException, more of an EntityNotFoundException

src/services/infrastructure/storage-aggregator-resolver/storage.aggregator.resolver.service.ts
```diff
@@ -82,17 +100,55 @@ export class StorageAggregatorResolverService {
...
+        storageAggregator: {
+          id: storageAggregator.id,
+        },
+      },
+      relations: {
+        profile: true,
+      },
+    });
+    if (!organization) {
+      throw new NotImplementedException(
```
valentinyanakiev: this is not NotImplementedException, more of an EntityNotFoundException

src/services/infrastructure/storage-aggregator-resolver/storage.aggregator.resolver.service.ts
```diff
@@ -82,17 +100,55 @@ export class StorageAggregatorResolverService {
...
+        storageAggregator: {
+          id: storageAggregator.id,
+        },
+      },
+      relations: {
+        profile: true,
+      },
+    });
+    if (!user) {
+      throw new NotImplementedException(
```
valentinyanakiev: this is not NotImplementedException, more of an EntityNotFoundException

src/migrations/1723036109999-storageAggregatorType.ts
```diff
@@ -0,0 +1,61 @@
...
+    await this.updateStorageAggregatorTypeForEntity(
+      queryRunner,
+      'user',
+      'user'
+    );
+    await this.updateStorageAggregatorTypeForEntity(
+      queryRunner,
+      'organization',
+      'organization'
+    );
```
valentinyanakiev: I don't see logic about the `Platform` `StorageAggregator`

valentinyanakiev (changes requested):
Some comments inline.
I also suggest this is covered with API tests before we take it in, then we will make sure it works and it will continue working in the future.

techsmyth:
@valentinyanakiev  comments are I think addressed. Re api test, would not hold this up for that as parentEntity field is not really used outside of spaces for now, so this was more about stopping getting a ton of errors in the logs rather than getting a field we depend on robust.

valentinyanakiev (changes requested):
One final change before we merge that - level on `ParentEntity` field resolver returns `SPACE` for `Platform`, `Organization`, `Account`, `User`. That doesn't make any sense - it should be nullable and the field should be null, in case it is not relevant at all for the parent entity.

Raised [a story](https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/test-suites/353) to add tests. Looks to be working functionally with data from acceptance, tested only on API level.

valentinyanakiev (approved):

PR #4393, opened 2024-08-11
Authorization reset: only process policies data; remove Relation; policy type

techsmyth:
Idea is that auth reset should only save authorizationPolicies, not touch any other data on the platform. The current authorization reset process ends up saving **all the data basically on the platform". 

Design used:
- Authorization services all only return authorization policies that are updated
- There should / must not be any saving of actual domain data from within these services
- After an authorization reset has completed, it is the responsibility of the calling service to actually commit all the updated authorization policies to the database

NOTE: The intention / expectation is that this will be significantly faster than the current approach. This approach is also simply safer: the authorization resets no longer touch / update / modify in any way the non-authorization policy entities. 

Also removed the Relation module; no point in updating its authorization etc code. Also includes a migration in this PR. 

Key testing that is needed is the test suite that validates assigned privileges. This is a lot of code, so a reasonable chance of regression.


valentinyanakiev:
> Idea is that auth reset should only save authorizationPolicies, not touch any other data on the platform. The current authorization reset process ends up saving **all the data basically on the platform".
> 
> Design used:
> 
> * Authorization services all only return authorization policies that are updated
> * There should / must not be any saving of actual domain data from within these services
> * After an authorization reset has completed, it is the responsibility of the calling service to actually commit all the updated authorization policies to the database
> 
> NOTE: The intention / expectation is that this will be significantly faster than the current approach. This approach is also simply safer: the authorization resets no longer touch / update / modify in any way the non-authorization policy entities.
> 
> Also removed the Relation module; no point in updating its authorization etc code. The remainder of the work (e.g. to remove the data from the database) is to be done under server-3669 + should be done very near term. This PR does not do that to avoid having a migration in the PR.
> 
> Key testing that is needed is the test suite that validates assigned privileges. This is a lot of code, so a reasonable chance of regression.

It is not amazingly faster (left is new algorithm, right is old one, test dataset is ACC data):
image

hero101 (approved):
Tested locally with test-suits branch server-3669 - passing with flying colors
image

PR #4397, opened 2024-08-12
"flow-state" tagset template is  created in advance

hero101:


hero101:
drafting because I found a bug while using the client

hero101:
Turns out it was due to corrupted data/schema - ready for review again.

valentinyanakiev (changes requested):
Doesn't work for me, tested with createing a subspace of an existing space.

I get a lot of "Unable to load tagsetTemplate for Tagset:...".
When I create a subspace with the code from `develop`, it works. This will introduce a regression.

valentinyanakiev (approved):

PR #4400, opened 2024-08-13
NameID simplified & tested

hero101:
### NameID generation algorithm
  0. a form of a name is provided and is used as a base for the nameID. For example user's first and lastname, callout's display name, etc.
  1. any [diacritics](https://www.google.com/search?q=diacritics) are replaced by their coresponding letter in the English alphabet
  2.  the product from the previous step is then stripped out of any none alphanumeric characters. Dashes are allowed.
  3. the product from the previous step is lower cased
  4. if the product from the previous step does not meet the minimum length (currently 5), it will be extended with random alphanumeric characters until the minimum length (currently 5) is met
  5. the product from the previous step is clamped at max length (currently 25) and returned

### NamingService.createNameID
- made private
- extracted to a testable function
- rearranged the implementation
---

- added tests for  `NamingService.createNameIdAvoidingReservedNameIDs`


techsmyth (changes requested):
I tried creating a new user using the data below:

{
  "input": {
    "email": "asdf@asdf.com" ,
    "profileData": {
      "displayName": "?????.....?..-",
      "description": "",
      "tagline": ""
    }
  }
}

which gave the following error:

image


Also I tried to do the same with organization and it seems the new logic is not applied for organizations, I get a nameID that was too short (one character).

hero101:
> I tried creating a new user using the data below:
> 
> { "input": { "email": "[asdf@asdf.com](mailto:asdf@asdf.com)" , "profileData": { "displayName": "?????.....?..-", "description": "", "tagline": "" } } }
> 
> which gave the following error:
> 
> image
> 
> Also I tried to do the same with organization and it seems the new logic is not applied for organizations, I get a nameID that was too short (one character).

@techsmyth
- There was never a minimum limit to the nameID
- The screenshot you provided is also reproducible on develop
- The code is propagated to the same nameID generation functions as before. Only the underlying core function is refactored

techsmyth (dismissed):
Approving based on my testing. Svetlo: if you are ok with my updates then feel free to merge.

hero101:
@techsmyth I have added `CreateNameableOptionalInput` for consistency reasons, and it needs to be used where the same behaviour is required.
The two classes can go out of sync - let me know if you have ideas how to keep them in sync.

Still need to fix the tests after your latest change, and run some integration tests, since the changes started to pile up and they touch  important code.
It indeed takes more time than anticipated, but I am not confidant in the changes.

techsmyth (approved):

PR #4412, opened 2024-08-16
Avatar generation + storage for contributors

techsmyth:
moved avatar URL creation to a separate module; 
streamlined creation of avatars on contributors; 
moved code to retrieve + store avatars to single location; 
removed dependency on contributor module from account host module


valentinyanakiev:
Simplified the code a bit, removed the dependency on `file-type` package - just setting the MimeType to PNG for the avatars. I have one workaround (see the comment added in `UserService`). I have left `uploadAvatarFromExternalURL` in `ContributorService`. Whether that's the best place, is up for debate. @techsmyth any thoughts? This works, and is tested.

techsmyth:
Important I think to only do the move etc after the contributor is fully created, so moved the actual image retrieval etc to be in a separate method that is called. The also means we can use the actual contributor ID for the Document creation. Hit this issue as bootstrap was failing.

src/domain/community/contributor/contributor.service.ts
```diff
@@ -15,23 +14,99 @@ import { IAgent } from '@domain/agent/agent/agent.interface';
...
-      limit
+  public async addAvatarVisualToContributorProfile(
+    profile: IProfile,
+    profileData: CreateProfileInput,
+    agentInfo?: AgentInfo,
+    firstName?: string,
+    lastName?: string
+  ): Promise<void> {
+    let avatarURL: string | undefined = undefined;
+    if (profileData.avatarURL && profileData.avatarURL !== '') {
```
bobbykolev: `&& profileData.avatarURL !== ''` is redundant

src/domain/community/user/user.service.ts
```diff
@@ -117,7 +117,7 @@ export class UserService {
 
     await this.validateUserProfileCreationRequest(userData);
 
-    const user: IUser = User.create(userData);
+    let user: IUser = User.create(userData);
```
bobbykolev, line 120: I'm not sure that mutating this response is better than keeping a separate state for `createUserResponse` and `updateUserResponse` below. 

Please, double-check.

src/domain/storage/storage-bucket/storage.bucket.service.ts
```diff
@@ -395,4 +397,38 @@ export class StorageBucketService {
...
+    }
+
+    // Not stored on Alkemio, download + store
+    const imageBuffer = await this.avatarCreatorService.urlToBuffer(avatarURL);
+
+    const document = await this.uploadFileAsDocumentFromBuffer(
+      storageBucketId,
+      imageBuffer,
+      'avatar',
+      MimeTypeVisual.PNG,
```
bobbykolev: Not sure how this would work with different mime types.

Note that the avatars from https://eu.ui-avatars.com/ weren't looking very good once downloaded directly from the link.
As the downloaded images were low res, I've added the following size matching the big contributor cards in the contributors' layout - `&size=145` to the end of the avatar generation link. However, it still doesn't look sharp enough, we should consider retina > 2 as well.

src/services/external/avatar-creator/avatar.creator.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import axios, { AxiosResponse } from 'axios';
+import { LogContext } from '@common/enums';
+
+@Injectable()
+export class AvatarCreatorService {
+  constructor(
+    @Inject(WINSTON_MODULE_NEST_PROVIDER) private readonly logger: LoggerService
+  ) {}
+
+  public generateRandomAvatarURL(firstName: string, lastName?: string): string {
```
bobbykolev, line 13: Not required in this PR, but I've noticed that the generation breaks in case of symbols passed in the firstName/lastName. We should sanitize them before generation.

bobbykolev:
I got an error after email registration (on the login step, locally):

image

src/domain/community/contributor/contributor.service.ts
techsmyth: the alternative is to check that it is an actual URL, which may be the better check. But we do need to guard that there is a meaningful avatar URL given here

src/domain/storage/storage-bucket/storage.bucket.service.ts
techsmyth: Good point, so we need to update the image link that is created.

src/services/external/avatar-creator/avatar.creator.service.ts
techsmyth: Actually makes a lot of sense to sanitize here.

src/domain/community/user/user.service.ts
techsmyth: does not really make a difference. It is generally messy to have user, userSavedOnce, userSavedTwice etc. So easier to just reuse...

techsmyth:
@bobbykolev thanks for the review. Just pushed changes that a) check for valid URL b) sanitize the provided first + last name. c) add size. 

Re error, please try again with login setup you used above. I suspect the extra check now added re detecting valid URL will avoid this case.

PR #4413, opened 2024-08-16
Limit history size sent to engine to the last 10 messages && send vc name and description to the engine

valeksiev:


techsmyth (commented):
Overall makes sense. Main question I have is why both a threadID and an interactionID? Surely one one of the two is needed i.e. we could never have an intereaction ID in two different thread IDs? Or is that me thinking too much in the current implementation + that later they may be both needed?

valentinyanakiev (changes requested):
Please add the N interactions limit in the config, and provide the number of the interactions, not N+1 (as discussed in the daily).

valeksiev:
> Overall makes sense. Main question I have is why both a threadID and an interactionID? Surely one one of the two is needed i.e. we could never have an intereaction ID in two different thread IDs? Or is that me thinking too much in the current implementation + that later they may be both needed?
We can eventually read the history from the engine itself through GQL. Right now, adding data to the message works, but we will eventually need to change that as the message size will become a blocker.

The main focus of this PR is to send the name and description of the VC to the engine, but I decided to limit the history as it was generating a lot of noise in the logs. In the engine itself, we have a similar cleanup - use the last N (set to 10 currently) messages from the history in order to stay withing the tokens limitations for the LLMs - so no point in processing and sending the messages and then discarding them at the last possible moment.

Also, the conversion from Matrix messages to _AI messages_ is of complexity 'configured limit' instead of 'messages in the room'.

valentinyanakiev (approved):
Haven't tested locally, but the changes I requested are in.
I suggest we test on ACC, I have configured Mistral API there.

PR #4418, opened 2024-08-16
whiteboard optimized auth reset fix; moved auth policy update of whiteboard to mutation for update

techsmyth:
updated logic so only update auth policy if actually change the settings

Note: only important to update the auth policy when change the content policy of a whiteboard via the update mutation on whiteboard. It is not needed when do an update on a whiteboard from a template.

Note: also fixes whiteboard auth reset in new optimized loading setup.


valentinyanakiev (approved):
Code looks fine, tested manually:
- creating callouts
- opening existing callouts
- changing content update policies of a whiteboard and accessing it with another user
- creating a subspace, creating callouts inside it

additionally ran:

- npm run test:preferences
- npm run test:roles
- npm run test:usermanagement
- npm run test:callouts
- npm run test:storage
- npm run test:journey

All passed, code looks fine.

PR #4434, opened 2024-08-20
License credentials on Space, not Account

techsmyth:
Migration for merging of accounts updated to move license credentials to be on Space agent, not account Agent

Fields related to license plan that is active moved to be on Space, not License

Mutations for assigning + revoking licenses now on Space entity


ccanos (approved):

Comoque1:
@techsmyth there sis something suspicious with `account`, `space` and `credential` agents
While I was testing the story locally with latest develop branch, I identified the following (in case the query is correct): 
1. creating a new space from User account creates 2 similar credentials: one for the space and one for the account
   
2. querying locally the old data, doesn't seem migrated the agent credential to the space credential (again, if the query is correct)
```
SELECT distinct(cr.resourceID),   sp.nameID,sp.agentId as spaceAgent,cr.agentId as credentialAgent, 
acc.agentId as accountAgent, sp.id as spaceId, cr.resourceID, acc.id as accountId ##, cr.type 
FROM alkemio.account as acc
join alkemio.space as sp on (acc.id = sp.accountId)
join alkemio.credential as cr on (cr.agentId = acc.agentId or sp.agentId = cr.agentId)

where   sp.level = 0 ##sp.agentId = cr.agentId 
```

Comoque1:
Issue 1: fixed:



Issue 2: fixed:

PR #4440, opened 2024-08-21
first pass typeorm definitions & code changes

hero101:
# On the same topic, I have created a draft list of rules to follow while writing such definitions
### Defaults
- no defaults in entity definitions; set defaults through business logic
- no defaults to the class fields
### Relations
- relations that are not eager need to be defined as optional, since they need to be loaded explicitly
### Lengths
- always define length with constants
- stick to the existing length constants OR in the rear case - create one if nothing fits

|data|what to use|
|----------|----------|
| Alkemio identifiers | `char(UUID_LENGTH)` |
| enums | `varchar(ENUM_LENGTH)` |
| short text | `varchar(your length)` |
| long text | `text` |
| URI, URL | `varchar(URI_LENGHT)` |

### Important
- **generate a domain change migration via `migration:generate`**


src/common/constants/entity.field.length.constants.ts
```diff
@@ -1,14 +1,19 @@
+// text
 export const TINY_TEXT_LENGTH = 16;
 export const ALT_TEXT_LENGTH = 120;
+export const MINI_TEXT_LENGTH = 64;
```
techsmyth: Need to see where this is used, instinct is that the save in space is not worth it + just makes it confusing to have it. So maybe where this is used just use SMALL

src/common/constants/entity.field.length.constants.ts
```diff
@@ -1,14 +1,19 @@
...
+// ids
 export const UUID_LENGTH = 36;
+export const MESSAGEID_LENGTH = 44;
+// nameID
 export const NAMEID_MAX_LENGTH = 25;
 export const NAMEID_MIN_LENGTH = 5;
-export const MESSAGEID_LENGTH = 44;
-
+// others
+export const ENUM_LENGTH = 128;
```
techsmyth, line 17: Nice

src/domain/collaboration/innovation-flow/innovation.flow.entity.ts
```diff
@@ -15,6 +15,6 @@ export class InnovationFlow
   @JoinColumn()
   profile!: Profile;
 
-  @Column('text')
-  states: string = '[]';
+  @Column('simple-array', { nullable: false })
```
techsmyth, line 18: This may also fix the issue with comments in state names or not?

src/domain/collaboration/post/post.entity.ts
```diff
@@ -2,14 +2,15 @@ import { Column, Entity, JoinColumn, OneToOne } from 'typeorm';
...
 
 @Entity()
 export class Post extends NameableEntity implements IPost {
   @Column('text')
   type!: string;
 
-  @Column('char', { length: 36, nullable: true })
-  createdBy!: string;
+  @Column('char', { length: UUID_LENGTH, nullable: true })
+  createdBy?: string;
```
techsmyth: So standardise across the codebase that createdBy etc is optional? Triggers also the idea that all nameable entities should have createdBy seeing how often we have this field. If you agree then please raise a separate issue to make this change i.e. that all nameable entities track who they are created by, similar to created date.

src/domain/common/form/form.entity.ts
```diff
@@ -4,9 +4,9 @@ import { IForm } from './form.interface';
 
 @Entity()
 export class Form extends BaseAlkemioEntity implements IForm {
-  @Column('text')
+  @Column('text', { nullable: false })
   questions!: string;
 
   @Column('text', { nullable: true })
-  description = '';
+  description?: string;
```
techsmyth: We may want this to be mandatory. And actually later I could imagine that a form also gets a profile if it becomes more widely used. But for now I would suggest mandatory.

src/domain/common/location/location.entity.ts
```diff
@@ -1,26 +1,27 @@
...
-  stateOrProvince = '';
+  @Column('varchar', { length: SMALL_TEXT_LENGTH, nullable: true })
+  stateOrProvince?: string;
 
-  @Column('varchar', { length: 255, nullable: true })
-  postalCode = '';
+  @Column('varchar', { length: SMALL_TEXT_LENGTH, nullable: true })
+  postalCode?: string;
 
   constructor() {
```
techsmyth: remove the empty constructor?

src/domain/common/preference/preference.definition.entity.ts
```diff
@@ -36,7 +40,7 @@ export class PreferenceDefinition
   valueType!: PreferenceValueType;
 
   @Column({
-    length: SMALL_TEXT_LENGTH,
+    length: ENUM_LENGTH,
```
techsmyth: nullable false?

src/domain/common/tagset-template/tagset.template.entity.ts
```diff
@@ -18,20 +19,21 @@ export class TagsetTemplate
   implements ITagsetTemplate
 {
   @Column('varchar', {
-    default: RestrictedTagsetTemplateNames.DEFAULT,
-    length: 255,
+    length: ENUM_LENGTH,
```
techsmyth: SMALL as it is not an enum?

src/domain/common/tagset-template/tagset.template.entity.ts
```diff
@@ -18,20 +19,21 @@ export class TagsetTemplate
...
     nullable: false,
   })
   type!: TagsetType;
 
-  @Column('simple-array')
+  @Column('simple-array', {
+    nullable: false,
+    default: [], // todo check value
```
techsmyth: no default value?

src/domain/common/visual/visual.entity.ts
```diff
@@ -9,7 +9,7 @@ export class Visual extends AuthorizableEntity implements IVisual {
   @Column()
   name!: string;
 
-  @Column('text', { default: '' })
+  @Column('text', { nullable: false })
```
techsmyth: Is text the right type for this? maybe varchar with long text length? But am not sure of the trade offs text vs varchar.

src/domain/community/ai-persona/ai.persona.entity.ts
```diff
@@ -3,28 +3,23 @@ import { IAiPersona } from './ai.persona.interface';
...
 @Entity()
 export class AiPersona extends AuthorizableEntity implements IAiPersona {
   // No direct link; this is a generic identifier
-  @Column('varchar', { nullable: false, length: 128 })
+  @Column('varchar', { nullable: false, length: SMALL_TEXT_LENGTH })
   aiPersonaServiceID!: string;
 
   @Column('text', { nullable: true })
-  description = '';
+  description?: string;
```
techsmyth, line 15: I think description should be mandatory + be set in code to empty string if not provided. Or we require it in the create DTO.

src/domain/community/ai-persona/ai.persona.entity.ts
```diff
@@ -3,28 +3,23 @@ import { IAiPersona } from './ai.persona.interface';
...
 
-  @Column('varchar', {
-    length: 255,
-    default: AiPersonaDataAccessMode.SPACE_PROFILE_AND_CONTENTS,
-  })
+  @Column('varchar', { length: MINI_TEXT_LENGTH, nullable: false })
```
techsmyth: This is an enum, so suggest to use enum length

src/domain/community/ai-persona/ai.persona.entity.ts
```diff
@@ -3,28 +3,23 @@ import { IAiPersona } from './ai.persona.interface';
...
-  @Column('simple-array', {
-    nullable: false,
-    default: [AiPersonaInteractionMode.DISCUSSION_TAGGING],
-  })
+  @Column('simple-array', { nullable: false })
   interactionModes!: AiPersonaInteractionMode[];
 
   @Column('text', { nullable: true })
-  bodyOfKnowledge = '';
+  bodyOfKnowledge?: string;
```
techsmyth, line 24: I think this is an identifier, so should be varchar + probably small text length. Please check the usage. We may need to discuss this one.

src/domain/community/invitation/invitation.entity.ts
```diff
@@ -21,18 +23,18 @@ export class Invitation extends AuthorizableEntity implements IInvitation {
   })
   community?: Community;
 
-  @Column('char', { length: 36, nullable: true })
+  @Column('char', { length: UUID_LENGTH, nullable: false })
   invitedContributor!: string;
+  // todo: can we create an invite without a creator?
+  @Column('char', { length: UUID_LENGTH, nullable: true })
+  createdBy?: string;
```
techsmyth: No, needs to have createdBy I think here. So mandatory.

src/domain/community/user/user.entity.ts
```diff
@@ -25,22 +26,22 @@ export class User extends ContributorBase implements IUser {
   rowId!: number;
 
   @Column()
```
techsmyth: All these columns really should also have lenghts / nullable false

src/domain/community/virtual-contributor/virtual.contributor.entity.ts
```diff
@@ -27,7 +30,7 @@ export class VirtualContributor
   listedInStore!: boolean;
 
   @Column('varchar', {
-    length: 36,
+    length: MINI_TEXT_LENGTH,
```
techsmyth: Enum length

src/domain/space/space/space.entity.ts
```diff
@@ -100,9 +104,8 @@ export class Space extends NameableEntity implements ISpace {
   level!: number;
 
   @Column('varchar', {
-    length: 36,
+    length: MINI_TEXT_LENGTH,
```
techsmyth: Enum length

src/domain/storage/document/document.entity.ts
```diff
@@ -4,13 +4,19 @@ import { StorageBucket } from '../storage-bucket/storage.bucket.entity';
...
+  UUID_LENGTH,
+} from '@common/constants';
 
 @Entity()
 export class Document extends AuthorizableEntity implements IDocument {
   // omitting OneToOne decorator for createdBy to avoid circular dependency
   // needs a redesign to avoid circular dependency
-  @Column('char', { length: 36, nullable: true })
-  createdBy!: string;
+  @Column('char', { length: UUID_LENGTH, nullable: true })
```
techsmyth, line 18: createdBy is mandatory in teh code, and I think it should also be here. We need to track who uploaded a document.

src/domain/timeline/event/event.entity.ts
```diff
@@ -27,19 +28,19 @@ export class CalendarEvent extends NameableEntity implements ICalendarEvent {
...
+  @Column('datetime', { nullable: false })
   startDate!: Date;
 
-  @Column('boolean', { default: true })
-  wholeDay = true;
+  @Column('boolean', { nullable: false })
+  wholeDay!: boolean;
 
-  @Column('boolean', { default: true })
+  @Column('boolean', { nullable: false })
```
techsmyth, line 38: No default value?

src/platform/activity/activity.entity.ts
```diff
@@ -12,27 +16,27 @@ export class Activity extends BaseAlkemioEntity implements IActivity {
...
 
-  @Column('boolean', { default: true })
-  visibility = true;
+  @Column('boolean', { nullable: false })
+  visibility!: boolean;
 
   @Column({
-    // TODO: It's 255, migrate it to something else
-    length: SMALL_TEXT_LENGTH,
+    length: 255, // already 255 in db
```
techsmyth: Length via constant? Medium text length? This is a field we use in the code, not user provided data.

src/platform/forum-discussion/discussion.entity.ts
```diff
@@ -18,8 +19,8 @@ export class Discussion extends NameableEntity implements IDiscussion {
   @JoinColumn()
   comments!: Room;
 
-  @Column('char', { length: 36, nullable: true })
-  createdBy!: string;
+  @Column('char', { length: UUID_LENGTH, nullable: true })
+  createdBy?: string;
```
techsmyth: Mandatory I think

src/platform/invitation/platform.invitation.entity.ts
```diff
@@ -42,11 +44,11 @@ export class PlatformInvitation
   @Column('varchar', { length: 255, nullable: true })
```
techsmyth: length via costant

src/services/ai-server/ai-persona-service/ai.persona.service.entity.ts
```diff
@@ -17,11 +18,11 @@ export class AiPersonaService
...
 
-  @Column({ length: 128, nullable: false })
+  @Column({ length: ENUM_LENGTH, nullable: false })
   engine!: AiPersonaEngine;
 
   @Column({
-    length: 64,
+    length: ENUM_LENGTH,
     nullable: false,
     default: AiPersonaDataAccessMode.SPACE_PROFILE,
```
techsmyth: No default value?

src/services/ai-server/ai-persona-service/ai.persona.service.entity.ts
```diff
@@ -30,11 +31,11 @@ export class AiPersonaService
   @Column('text', { nullable: false })
   prompt!: string;
 
-  @Column({ length: 64, nullable: true })
-  bodyOfKnowledgeType!: AiPersonaBodyOfKnowledgeType;
+  @Column({ length: ENUM_LENGTH, nullable: true })
+  bodyOfKnowledgeType?: AiPersonaBodyOfKnowledgeType;
 
   @Column({ length: 255, nullable: true })
```
techsmyth: length via constant?

techsmyth (commented):
Super update, cleans up a lot. 

Basically I would remove the mini text length, it seemed to be mostly used for enums anyway so can use that length. Also do remove all usage of default field so we have a consistent state. 

And in looking at this code it does seem to me that we want to have createdBy on Nameable entity, a separate issue but one to track.

src/common/constants/entity.field.length.constants.ts
hero101: it was used for enums, so I replaced it with small

src/domain/collaboration/innovation-flow/innovation.flow.entity.ts
hero101: I don't know to what are you referring to, but this will parse the text to an array everytime it's read, and parse it back to CSV when written.

src/domain/collaboration/post/post.entity.ts
hero101: I have left `createdBy` to be optional since the code creates a lot of things and we don't have a standardized "creator" in these cases.
Furthermore, it would be great to be an FK instead of a string value - that would only be possible if the ID is coming from a single table, i.e. `user`.

src/domain/common/preference/preference.definition.entity.ts
hero101: `false` is by default

src/domain/common/visual/visual.entity.ts
hero101: https://stackoverflow.com/a/417184/11207901

> ...and the reality
That's what the standards say. For the reality, there was an article on [boutell.com](https://web.archive.org/web/20190902193246/https://boutell.com/newfaq/misc/urllength.html) (link goes to Internet Archive backup) that discussed what individual browser and server implementations will support. The executive summary is:
>> Extremely long URLs are usually a mistake. URLs over 2,000 characters will not work in the most popular web browsers. Don't use them if you intend your site to work for the majority of Internet users.

src/domain/community/ai-persona/ai.persona.entity.ts
hero101: A description is usually not mandatory, but we can enforce it in the API layer.
About the column type - the API layer enforces `65568`. We can either match it or standardize it to `DESCRIPTION_LENGTH` of some length.
 What do you think?

src/domain/community/ai-persona/ai.persona.entity.ts
hero101: This is `text`, limited to `65568` in the API layer.
There is another `bodyOfKnowledgeID` which is the identifier.

src/domain/community/user/user.entity.ts
hero101: Let's discuss them additionally. They are currently `varchar(255)`.

src/domain/storage/document/document.entity.ts
hero101: We need to standardize the "creator". Currently, any 36-long string can be a creator.
In addition, if the code creates something who would be the creator?

hero101:
@techsmyth 
Some highlights after the first review pass
- description - mandatory vs optional; length; enforcement in API layer?
- createdBy - I have left `createdBy` to be optional since the code creates a lot of things and we don't have a standardized "creator" in these cases. Furthermore, it would be great to be an FK instead of a string value - that would only be possible if the ID is coming from a single table, i.e. user.
- User columns need further discussion.

src/domain/collaboration/post/post.entity.ts
techsmyth: To consider a FK, but it cannot in many cases be a TypeORM relation  as that leads to all kinds of circular dependencies. For later.

src/platform/forum-discussion/discussion.entity.ts
techsmyth: Ok

src/domain/common/form/form.entity.ts
techsmyth: Agreed that this should then also be enforced on the api layer

src/domain/community/ai-persona/ai.persona.entity.ts
techsmyth: On reconsideration, we probably want this description to match the definition on the Profile entity - which would make it nullable.

src/domain/community/ai-persona/ai.persona.entity.ts
techsmyth: Ok, so it is actually the bodyOfKnowledge description. Then ok as is you have it above.

src/domain/common/preference/preference.definition.entity.ts
techsmyth: we have it explicit in may places, maybe also we just state it always?

src/domain/common/visual/visual.entity.ts
techsmyth: Ok, then varchar with LONG_TEXT_LENGTH?

src/domain/storage/document/document.entity.ts
techsmyth: If the code creates something then we have to find a reasonable identifier to put in. E.g. for the user creation flow in bootstrap I put the actual user id in. But it should be mandatory for documents so we know who created it.

techsmyth:
@hero101 I have added the suggested guidelines into the docs folder for permanent reference

valentinyanakiev (approved):

PR #4453, opened 2024-08-23
tidied up account authorization so that host credentials give CRUD on resources in account + cascades

techsmyth:
I was able as org admin to create a VC.

Other case I suspect is related to the PR raised with picking up the wrong agent for auth policy reset.


bobbykolev:
Ok, the straight case where an admin creates a space (vc license applied) and then creates VC is now covered.

However, when creating a space through the Org Account Tab doesn't set the vc license resulting into the following bug:
https://github.com/alkem-io/server/issues/4463
And if an Org Admin tries to create a VC in a space that that user did not create, then we're facing the following:
https://github.com/alkem-io/server/issues/4464

Comoque1 (approved):

PR #4454, opened 2024-08-23
Client 5405 sort collections

EmiliaPavlova:
The PR is now ready. 

Additionally to what was implemented the following changes were applied:
 - [x] Migration updated;
 - [x] The sorting changes onSave were removed;
 - [x] The DB update with the sorting was added;
 - [x] The failing response of the mutation was fixed (took a lot of time) - missing brackets around the [ICalloutContribution] return type;

https://github.com/alkem-io/server/pull/4454/commits/fdf19fd5bd0e769273bdac7e56fb92ce9b3da335
https://github.com/alkem-io/server/pull/4454/commits/afaf2498994cebce2389b6f7f0b8d9f2c5255f98


bobbykolev (dismissed):

src/domain/collaboration/callout/callout.service.ts
```diff
@@ -466,7 +468,7 @@ export class CalloutService {
   ): Promise<ICalloutContribution> {
     const calloutID = contributionData.calloutID;
     const callout = await this.getCalloutOrFail(calloutID, {
-      relations: { contributionPolicy: true },
+      relations: { contributions: true },
```
EmiliaPavlova, line 471: Are you sure we don't need contributionPolicy anymore?

src/domain/collaboration/callout/dto/callout.dto.event.post.created.ts
```diff
@@ -7,14 +7,25 @@ export class CalloutPostCreated {
...
 
+  @Field(() => String, {
+    nullable: false,
+    description: 'The identifier of the Contribution.',
+  })
+  contributionID!: string;
+
+  @Field(() => Number, {
+    nullable: false,
+    description: 'he sorting order for this Contribution.',
```
bobbykolev: Typo in the description, missing 'T' at the beginning.

src/domain/collaboration/callout/callout.service.ts
bobbykolev: ah, good catch it seems this was there for the creation and it should stay, I'll double-check and push it back

bobbykolev (approved):

src/domain/collaboration/callout/callout.service.ts
bobbykolev: It looks like it works flawlessly without it.

PR #4458, opened 2024-08-24
pick up the level zero space agent for authorization logic

techsmyth:
Previous fix only helped level zero spaces; with this fix all subspaces now also pick up the correct licensing.


Comoque1:
Fix works for newly created subspaces L1 and L2.
For existing, it doesn't (after running policy reset.).

Comoque1:
Fix works for newly created subspaces L1 and L2.
For existing, it doesn't (after running policy reset.).

Comoque1 (approved):

PR #4461, opened 2024-08-25
Templates

techsmyth:
Move to one Template entity, which can have one entity type that it is wrapping set...

Post entity has the type column removed, it was no longer used.

Added new service InputCreator, that takes an entity and returns the relevant input DTO to create a new one. Rational is to centralize the way we essentially close entities on the platform:
image

Removed usage of template IDs as part of entity create DTOs (especially collaboration related), that was mixing two different concepts. Now the client uses the input creator to create essentially input for a clone, and then does the create. We could later put up some kind of "cloning service" that does both steps, but that I think is not needed. 

Risk areas:
- URLs
- Creation of spaces + collaboration in particular based on other collaborations.
- Creation + updating of templates
- Notification related to post created: the payload no longer sends the type, the notifications service may need it.


techsmyth:
When creating a user via bootstrap I am now getting the following error:

image


This is not happening on develop, something with loading has gone wrong...

valentinyanakiev:
@coderabbitai review

bobbykolev (approved):

PR #4468, opened 2024-08-26
Server 4401 ingest space feedback

### LINKED ISSUE #4401: Ingest service feedback on finished ingestion
## Description

As a user I want to know when I can use my VC and not get errors.

## Acceptance criteria
- [ ] VC creation flow to show the screen to allow users to ask a question to their newly created VC
- [ ] Client display to take into account the status field, and subscribe to the updates

## Additional context
- [ ] Add field to the persona service to expose the last ingested time. If ingestion has never completed,  then this field is undefined. 
- [ ] VC entity to have a calculated "status" field with the following values: INITIALIZING, READY. 
- [ ] INITIALIZING = last ingested date is undefined; READY = last ingested date is available
- [ ] Subscription to receive updates on the change in status for a given VC
- [ ] Users using the web client can not use the VC mentions until the VC is Ready

## Additional Context
- Vertical story, client, server, ingest service

- client -  mentions
- server - mentions 
- server - new additional business logic (e.g. subscription) to be able to subscribe to with event when the ingestion is finished
- ingest service - raise an event when ingestion is finished

valeksiev:

Fixes: https://github.com/alkem-io/server/issues/4401

### DIFF
--- a/src/common/enums/virtual.contributor.status.enum.ts
+++ b/src/common/enums/virtual.contributor.status.enum.ts
@@ -1,0 +1,10 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum VirtualContributorStatus {
+  NITIALIZING = 'initializing',
+  READY = 'ready',
+}
+
+registerEnumType(VirtualContributorStatus, {
+  name: 'VirtualContributorStatus',
+});
--- a/src/common/enums/virtual.contributor.status.enum.ts
+++ b/src/common/enums/virtual.contributor.status.enum.ts
@@ -1,7 +1,7 @@
 import { registerEnumType } from '@nestjs/graphql';
 
 export enum VirtualContributorStatus {
-  NITIALIZING = 'initializing',
+  INITIALIZING = 'initializing',
   READY = 'ready',
 }
 
--- a/src/domain/community/virtual-contributor/dto/virtual.contributor.updated.subscription.args.ts
+++ b/src/domain/community/virtual-contributor/dto/virtual.contributor.updated.subscription.args.ts
@@ -7,5 +7,5 @@ export class VirtualContributorUpdatedSubscriptionArgs {
     description: 'The Virtual Contributor to receive the events for.',
     nullable: false,
   })
-  virtualContributorId!: string;
+  virtualContributorID!: string;
 }
--- a/src/domain/community/virtual-contributor/dto/virtual.contributor.updated.subscription.args.ts
+++ b/src/domain/community/virtual-contributor/dto/virtual.contributor.updated.subscription.args.ts
@@ -1,0 +1,11 @@
+import { ArgsType, Field } from '@nestjs/graphql';
+import { UUID_NAMEID } from '@domain/common/scalars';
+
+@ArgsType()
+export class VirtualContributorUpdatedSubscriptionArgs {
+  @Field(() => UUID_NAMEID, {
+    description: 'The Virtual Contributor to receive the events for.',
+    nullable: false,
+  })
+  virtualContributorId!: string;
+}
--- a/src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
+++ b/src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
@@ -1,0 +1,76 @@
+import { Args, Resolver, Subscription } from '@nestjs/graphql';
+import { CurrentUser } from '@src/common/decorators';
+import { VirtualContributorService } from './virtual.contributor.service';
+import { AgentInfo } from '@core/authentication.agent.info/agent.info';
+import { AuthorizationService } from '@core/authorization/authorization.service';
+import { AuthorizationPrivilege, LogContext } from '@common/enums';
+import { PlatformAuthorizationPolicyService } from '@platform/authorization/platform.authorization.policy.service';
+import { Inject, LoggerService, UseGuards } from '@nestjs/common';
+import { GraphqlGuard } from '@core/authorization';
+import { SubscriptionReadService } from '@services/subscriptions/subscription-service';
+import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import { VirtualContributorUpdatedSubscriptionArgs } from './dto/virtual.contributor.updated.subscription.args';
+import { VirtualContributorUpdatedSubscriptionResult } from './dto/virtual.contributor.updated.subscription.result';
+
+@Resolver()
+export class VirtualContributorResolverSubscriptions {
+  constructor(
+    private virtualContributorService: VirtualContributorService,
+    private authorizationService: AuthorizationService,
+    private platformAuthorizationPolicyService: PlatformAuthorizationPolicyService,
+    private subscriptionService: SubscriptionReadService,
+    @Inject(WINSTON_MODULE_NEST_PROVIDER)
+    private readonly logger: LoggerService
+  ) {}
+
+  @UseGuards(GraphqlGuard)
+  @Subscription(() => VirtualContributorUpdatedSubscriptionResult, {
+    nullable: false,
+    description: 'Receive updates on virtual contributors',
+    async filter(
+      this: VirtualContributorResolverSubscriptions,
+      payload,
+      variables
+    ) {
+      const isMatch =
+        variables.virtualContributorId === payload.virtualContributor.nameID;
+
+      this.logger.verbose?.(
+        `[Filtering VirtualContribuor updated event id '${payload.eventID}'; VC id ${payload.virtualContributor.nameID}- match=${isMatch}`,
+        LogContext.SUBSCRIPTIONS
+      );
+      return isMatch;
+    },
+  })
+  async virtualContributorUpdated(
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args({ nullable: true })
+    { virtualContributorId }: VirtualContributorUpdatedSubscriptionArgs
+  ) {
+    const platformPolicy =
+      await this.platformAuthorizationPolicyService.getPlatformAuthorizationPolicy();
+
+    const hasAccess = this.authorizationService.isAccessGranted(
+      agentInfo,
+      platformPolicy,
+      AuthorizationPrivilege.PLATFORM_ADMIN
+    );
+
+    if (!hasAccess) {
+      return {};
+    }
+    const vc =
+      await this.virtualContributorService.getVirtualContributorOrFail(
+        virtualContributorId
+      );
+
+    this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      vc.authorization,
+      AuthorizationPrivilege.READ,
+      `subscription to Virtual Contributor updates on: ${vc.id}`
+    );
+
+    return this.subscriptionService.subscribeToVirtualContributorUpdated();
+  }
+}
--- a/src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
+++ b/src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
@@ -1,5 +1,5 @@
-import { Args, Resolver, Subscription } from '@nestjs/graphql';
-import { CurrentUser } from '@src/common/decorators';
+import { Args, Resolver } from '@nestjs/graphql';
+import { CurrentUser, TypedSubscription } from '@src/common/decorators';
 import { VirtualContributorService } from './virtual.contributor.service';
 import { AgentInfo } from '@core/authentication.agent.info/agent.info';
 import { AuthorizationService } from '@core/authorization/authorization.service';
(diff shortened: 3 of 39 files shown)


src/common/enums/virtual.contributor.status.enum.ts
```diff
@@ -0,0 +1,10 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum VirtualContributorStatus {
+  NITIALIZING = 'initializing',
```
techsmyth: Typo :)

src/domain/community/virtual-contributor/dto/virtual.contributor.updated.subscription.args.ts
```diff
@@ -0,0 +1,11 @@
+import { ArgsType, Field } from '@nestjs/graphql';
+import { UUID_NAMEID } from '@domain/common/scalars';
+
+@ArgsType()
+export class VirtualContributorUpdatedSubscriptionArgs {
+  @Field(() => UUID_NAMEID, {
+    description: 'The Virtual Contributor to receive the events for.',
+    nullable: false,
+  })
+  virtualContributorId!: string;
```
techsmyth: For DTOs used in the server we tend to use capital ID for fields

src/services/ai-server/ai-server/ai.server.module.ts
```diff
@@ -13,17 +13,20 @@ import { AiPersonaEngineAdapterModule } from '../ai-persona-engine-adapter/ai.pe
...
+import { SubscriptionServiceModule } from '@services/subscriptions/subscription-service';
+import { VirtualContributor } from '@domain/community/virtual-contributor/virtual.contributor.entity';
 
 @Module({
   imports: [
     AuthorizationModule,
     AuthorizationPolicyModule,
     AiPersonaServiceModule,
-    TypeOrmModule.forFeature([AiServer]),
+    TypeOrmModule.forFeature([AiServer, VirtualContributor]),
```
techsmyth, line 24: This is dangerous linkage, need to understand the rational...

src/common/enums/virtual.contributor.status.enum.ts
hero101: typo in `NITIALIZING`

src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
```diff
@@ -0,0 +1,76 @@
...
+    private virtualContributorService: VirtualContributorService,
+    private authorizationService: AuthorizationService,
+    private platformAuthorizationPolicyService: PlatformAuthorizationPolicyService,
+    private subscriptionService: SubscriptionReadService,
+    @Inject(WINSTON_MODULE_NEST_PROVIDER)
+    private readonly logger: LoggerService
+  ) {}
+
+  @UseGuards(GraphqlGuard)
+  @Subscription(() => VirtualContributorUpdatedSubscriptionResult, {
```
hero101: Please use the `TypedSubscription` decorator instead.

src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.logger.verbose?.(
+        `[Filtering VirtualContribuor updated event id '${payload.eventID}'; VC id ${payload.virtualContributor.nameID}- match=${isMatch}`,
+        LogContext.SUBSCRIPTIONS
+      );
+      return isMatch;
+    },
+  })
+  async virtualContributorUpdated(
+    @CurrentUser() agentInfo: AgentInfo,
+    @Args({ nullable: true })
```
hero101: not nullable

src/domain/community/virtual-contributor/virtual.contributor.resolver.subscriptions.ts
```diff
@@ -0,0 +1,76 @@
...
+    const platformPolicy =
+      await this.platformAuthorizationPolicyService.getPlatformAuthorizationPolicy();
+
+    const hasAccess = this.authorizationService.isAccessGranted(
+      agentInfo,
+      platformPolicy,
+      AuthorizationPrivilege.PLATFORM_ADMIN
+    );
+
+    if (!hasAccess) {
```
hero101: I don't believe you need to have `PLATFORM_ADMIN` to subscribe

src/migrations/1724333243087-addBoKLastUpdateToAiPersonaService.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddBoKLastUpdateToAiPersonaService1724333243087
+  implements MigrationInterface
+{
+  name = 'AddBoKLastUpdateToAiPersonaService1724333243087';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE \`ai_persona_service\` ADD \`bodyOfKnowledgeLastUpdated\` datetime DEFAULT NULL`
```
hero101, line 10: this does not match the column definition in `ai.persona.service.entity.ts`

techsmyth (changes requested):
Looks good!

Main concern is the linkage back from AI Server to VC, to be chatted through...

src/services/infrastructure/event-bus/messages/ingest.space.result.event.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class IngestSpaceResult implements IEvent {
+  constructor(
+    public readonly spaceId?: string,
+    public readonly purpose?: SpaceIngestionPurpose,
+    public readonly personaServiceId?: string,
+    public readonly timestamp?: number,
+    public result?: SpaceIngestionResult,
+    public error?: string
+  ) {}
```
hero101, line 30: An object as input would make it way more flexible and future-proof.

hero101, line 30: Are all of the inputs optional - what is the purpose of the class then?

src/services/ai-server/ai-persona-service/ai.persona.service.interface.ts
```diff
@@ -39,4 +39,10 @@ export class IAiPersonaService extends IAuthorizable {
     description: 'The body of knowledge ID used for the AI Persona Service',
   })
   bodyOfKnowledgeID!: string;
+
+  @Field(() => Date, {
+    nullable: true,
+    description: 'When wat the body of knowledge of the VC last updated.',
+  })
+  bodyOfKnowledgeLastUpdated?: Date;
```
hero101: doesn't this field need to be set to `undefined/null` at some point? - if not, then it's never null.

src/migrations/1724333243087-addBoKLastUpdateToAiPersonaService.ts
```diff
@@ -0,0 +1,23 @@
...
+    );
+
+    await queryRunner.query(
+      `UPDATE \`ai_persona_service\` SET \`bodyOfKnowledgeLastUpdated\`= NOW()`
```
hero101, line 14: I don't think this is a good idea. It will make every vc `ready`.
Maybe a better approach is to conditionally set it if a body of knowledge is defined.

hero101 (changes requested):

src/services/ai-server/ai-server/ai.server.service.ts
```diff
@@ -44,26 +46,76 @@ export class AiServerService {
...
+      where: {
+        aiPersona: {
+          aiPersonaServiceID: personaService.id,
+        },
+      },
+      relations: { aiPersona: true },
+    });
+
+    if (virtualContributor) {
+      await this.subscriptionPublishService.publishVirtualContributorUpdated(
```
hero101, line 124: I think this should be done in another place, higher in the hierarchy, to not have a dependency between this service and `SubscriptionPublishService`

src/migrations/1724333243087-addBoKLastUpdateToAiPersonaService.ts
valeksiev: Running the generator, I don't get any differences. What are you referring to?

src/migrations/1724333243087-addBoKLastUpdateToAiPersonaService.ts
valeksiev: Having a VC without a BoK would be a data inconsistency error. I've added that for backward compatibility - if you have a VC you have used all of a sudden, it'd pop up as initializing on the FE.

src/services/ai-server/ai-server/ai.server.service.ts
valeksiev: Any ideas where to put it? The stack here is `Nests CQRS -> IngestResultHandler -> This method`.

src/services/ai-server/ai-persona-service/ai.persona.service.interface.ts
valeksiev: I don't understand that completely - the question mark gives me the `undefined`. For newly created VCs before the ingestion is completed this will undefined.

src/migrations/1724333243087-addBoKLastUpdateToAiPersonaService.ts
hero101: it's because BoK is nullable in the database but probably enforced with other means.

src/migrations/1724333243087-addBoKLastUpdateToAiPersonaService.ts
hero101: my bad

src/services/ai-server/ai-persona-service/ai.persona.service.interface.ts
hero101: What about on re-ingestion? Shouldn't it be set to `undefined`? The state is based on this field having a value.

valeksiev: Do you mean to set it to undefined while the BoK is being re-ingested? Does that make sense? The old one is available while the BoK is being updated. I agree we can have a state like `UPDATING` or something, but I'd say this is outside this PR's scope and can't be handled with a single column.

hero101: The problem is that it does not display the right behavior to the user.

Let's imagine this scenario:
I have just re-ingested, and my VC is already ready but has no clue about the latest knowledge I gave it, it could even not work, because it's currently initializing.

---
I agree that the issue was about newly created VCs but I see the same problem happening for old VCs and I think it can be handled by switching the state back to `initializing`.

src/services/ai-server/ai-server/ai.server.service.ts
hero101: At this point, my instinct tells me to have a service in between the handler and the `AiServerService`, which wraps simple functionality from multiple services, without coupling them together. 
A couple of real examples are having to send an email (notification service), that something VC-related happened, or that a new VC was created, reaching the quota (licensing service).
But if you believe it's too much feel free to leave it as is for now.

valeksiev:
@techsmyth @hero101 We all dislike the dependency of the AiServer on the the VC repository (or some other part of the collaboration server), but I can't think of a way around it. This will need reworking when we split out the AiServer anyway.

Ignoring this I think I've addressed all comments and the implementation is complete.

techsmyth (dismissed):
Understood re linkage back from AI server to collaboration server. To be addressed later.

techsmyth (approved):

ccanos:
Tested locally and it worked fine
- Latest `server-4401-ingest-space-feedback` on both client and server.
- Added `AZURE_OPENAI_API_KEY=...` to `.env.docker`
- Modified `AUTH_ADMIN_EMAIL` and `AUTH_ADMIN_PASSWORD` in `.env.docker` - careful with special characters, you'll need to escape the `$` sign, and maybe put the password in between double quotes.
- Delete Exchange `event-bus` in RabbitMQ if the type is not `direct`. If the type is `fanout` get in, delete the exchange and restart services
- Set to null all the `bodyOfKnowledgeLastUpdated` cells in the table `ai_persona_service` - careful with MySQLWorkbench, if safe updates is enabled you need to include an id column in all the WHERE clauses.
- Navigate to `http://localhost:3000/vc/<yourVC>/settings/settings`
- See that the status says Initializing:
image
- Open browser console and look for the graphql WebSocket
image
- Click on the button Update Knowledge
- See the new message coming that updates the status
image
- See that in MySQL table `ai_persona_service` the VC's `bodyOfKnowledgeLastUpdated` column has been updated
- I don't see any errors in the server or in the services logs

ccanos:
@valeksiev One note I just noticed, maybe it's okay, maybe you can't/don't want to do anything about it,
but the dates normally are stored in the database in UTC, and for the column `bodyOfKnowledgeLastUpdated` I have noticed it is storing current date as is. Or at least MySQLWorkbench is translating it to my timezone, not sure.
image

hero101 (approved):

PR #4470, opened 2024-08-26
Fix CREATE_SPACE permission for VC Campaign and Beta testers

ccanos:
The only real "fix" is this:
image

Before this change, all users with Admin, VCCampaign and BetaTester creds were getting the `CREATE` privilege was is already there.  Note that `CREATE` is not `CREATE_SPACE`.
Then, everybody was getting `CREATE_SPACE` and `CREATE_VIRTUAL_CONTRIBUTOR` in the function at the end `appendPrivilegeRules`.
I have commented the call to `appendPrivilegeRules` for now and have added CREATE_VIRTUAL_CONTRIBUTOR privilege to the users with those credentials


techsmyth (approved):
Super to see :) Yes this is I think the correct approach.

PR #4482, opened 2024-08-27
delete accounts for user + org after they are removed

techsmyth:


Comoque1:
Tested OK locally.

valentinyanakiev (approved):
Approving as it looks to solve current issues, but DDD is not properly designed in my opinion. Adding a [story](https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/server/4486) to review that.

PR #4489, opened 2024-08-28
Optimization of data loading on auth reset

techsmyth:
Optimizes the data that is loaded around Profiles.

Also done for CalloutFraming, CalloutContribution and User entity. 

Further optimization possible but this relatively small set of changes is showing some good results. On my local machine: 
- simple account with one space: auth reset went from 4.1s to 1.7s
- simple user: auth reset went from 380ms to 80ms

Suspect the win is across the board. More to be had, but given the above results thought it already interesting to push this, even if is just a very rough measure of performance impact.

### DIFF
--- a/src/domain/common/authorization-policy/authorization.policy.service.ts
+++ b/src/domain/common/authorization-policy/authorization.policy.service.ts
@@ -1,6 +1,6 @@
 import { Inject, Injectable, LoggerService } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import { Repository } from 'typeorm';
+import { FindOptionsSelect, Repository } from 'typeorm';
 import {
   EntityNotFoundException,
   ForbiddenException,
@@ -36,7 +36,7 @@ export class AuthorizationPolicyService {
     private readonly logger: LoggerService
   ) {}
 
-  public authorizationSelectOptions = {
+  public authorizationSelectOptions: FindOptionsSelect<AuthorizationPolicy> = {
     id: true,
     anonymousReadAccess: true,
     credentialRules: true,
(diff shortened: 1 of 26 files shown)


src/domain/common/profile/profile.service.authorization.ts
```diff
@@ -18,25 +17,73 @@ export class ProfileAuthorizationService {
...
+            authorization:
+              this.authorizationPolicyService.authorizationSelectOptions,
+            tagset: {
+              id: true,
+              authorization:
+                this.authorizationPolicyService.authorizationSelectOptions,
+            },
+          },
+        },
+      },
```
hero101, line 72: do you need all other fields except `authorization` and the `id`?

src/domain/common/authorization-policy/authorization.policy.service.ts
```diff
@@ -36,6 +36,14 @@ export class AuthorizationPolicyService {
     private readonly logger: LoggerService
   ) {}
 
+  public authorizationSelectOptions = {
```
hero101: Use the proper type here to be sure that you can react when `Authorization` changes.
It's a good practise to use `as const` at the end of the object declaration to make it read-only.

src/domain/common/profile/profile.service.authorization.ts
hero101: I see that it's actually `false`. `False` won't select them, but the default coming from the class will seem like they are loaded.

```typescript
  @Column('text', { nullable: false })
  displayName = '';

  @Column('text', { nullable: true })
  tagline = '';
```

**It is `false` by default for member columns of the table.**

hero101 (changes requested):
The code was not building.

I have fixed the comments myself. I will do light testing before approval.

hero101 (approved):
No regressions after running test suits

PR #4493, opened 2024-08-29
added mutations to assign + remove license plans to accounts

techsmyth:
plus: additional logic check re license plan type


valentinyanakiev (commented):
I am a bit confused not by this particular code, but by the logic in general:
- we have 1 platform, and we refrain from using its ID anywhere, as its implicitly only one
- it seems we also have only 1 licensing (is there a case in the future where we would have 2+, does it make sense in our current setup?)
- the licensing, albeit being one (One-to-One with Platform), is being queried / passed on. Isn't that redundant? E.g. 
```typescript
    if (planData.licensingID) { <----- Why do we need that?
      licensing = await this.licensingService.getLicensingOrFail(
        planData.licensingID
      );
    } else {
      licensing = await this.licensingService.getDefaultLicensingOrFail();
    }
```
- It seems like 10-20% of this code is defensive / optional for something we might have in the future, albeit we can't really have it, on database level. Happy to discuss.

valentinyanakiev (approved):
Approving following a discussion on the design and potential future redesign considerations.

Flows assign / revoke for space / account are working for me locally.

PR #4503, opened 2024-09-02
Whiteboard picking the right authorization

techsmyth:
Please test locally. 

And also merge if ok into develop


hero101 (approved):
Tested locally with two users working on the same whiteboard.

PR #4511, opened 2024-09-04
ensure authorizations policy is reset for newly created document

techsmyth:
When using `adminUpdateContributorAvatars` make sure auth reset is done so the new document is accessible.


bobbykolev:
I tested it locally with the https://github.com/alkem-io/cli/pull/71 script, and it looks good now.

bobbykolev (approved):

PR #4514, opened 2024-09-05
Backup scripts updated

valentinyanakiev:
- scripts updated to download backups from S3
- updated the script to validate hard dependency on aws cli v2 and install it if not present
- updated the script to use the new base paths of the backups in s3, and to use the different buckets
- to use, the AWS credentials in .env need to be set (will be sent separately)
- if a ~/.aws/config is already present, the user is prompted whether to override it

- draft until prod backups are deployed / validated
- validated locally with acc

### DIFF
--- a/.scripts/backups/restore_latest_backup.sh
+++ b/.scripts/backups/restore_latest_backup.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-# The storage is the first argument passed to the script. Default to 'mariadb' if not provided.
+# The storage is the first argument passed to the script. Default to 'mysql' if not provided.
 STORAGE=${1:-mysql}
 
 # Validate the storage
 if [[ "$STORAGE" != "mysql" && "$STORAGE" != "postgres" ]]; then
-    echo "Invalid storage '$STORAGE'. Please specify 'mysql', or 'postgres'."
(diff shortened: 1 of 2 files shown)


ccanos:
I had a few ideas about these scripts that I've been writing the last few weeks, feel free to address whichever you want, the script is already super useful for me as is. Haven't tested this PR though.

- drop and recreate alkemio database before restoring: I already have that changed and never commit it, it's just a duplicate of the line 43 in `restore_latest_backup.sh` (183 after your changes) to import first a sql file that has a DROP and a CREATE in it.
- Check yq version. Only >4 works and my Ubuntu (Kubuntu 24.04) installed with apt version 3.1.0-3. Luckily the version in `snap` is `v4.44.2` and got it to work quite quickly anyway.
- Check if the db file is already downloaded, if so, skip the download. At the moment everytime I run it it re-downloads both, which may be useful for someone, but for me I don't need it.
- Pass the filename of the backup you want to restore as an argument instead of downloading it, not sure if this can be done already calling `restore_latest_backup.sh` directly.
- Do not restart services automatically, at least not both... or maybe yeah.. This is a matter of personal preferences, I normally have services and server running in separate terminals. If it can be parametrized it would be great.

PR #4518, opened 2024-09-07
remove extra index on storage bcket to authorization

techsmyth:
Found this when tracking down a TypeORM issue


valentinyanakiev:
will be fixed as part of the index cleanup

PR #4534, opened 2024-09-12
Streamline migrations - both prod / local

valentinyanakiev:
- works with clean setup
- works with prod backup
- altered the approach - dropped all dropable Foreign Keys, Unique Indexes, Indexes and re-created them 
- removed the index / FK / unique index (relation) matching via @Index decorator - now all works with defaults

This is the alternative approach to what I did yesterday, it is not aligning the code with the DB but the other way round. It should remove all redundancies that way. 

It is impossible to outright remove all indexes as there are  dependencies, like AutoIncrement for instance mandates an Index etc.

This should fix most REL_ IDX_ FK_ mismatches as well

## Summary by CodeRabbit

- **Refactor**
	- Removed unnecessary index decorators from the `InnovationFlow` class, simplifying the database schema.
	- Updated the relationship handling to set profile references to `NULL` upon deletion, enhancing data integrity.

- **Chores**
	- Introduced migration files to remove foreign keys, unique constraints, and regular indexes from the database, facilitating schema optimization.
	- Modified foreign key constraints for the `innovation_flow` table to improve handling of deletions in the `profile` table.

### DIFF
--- a/src/domain/collaboration/collaboration/collaboration.entity.ts
+++ b/src/domain/collaboration/collaboration/collaboration.entity.ts
@@ -1,11 +1,4 @@
-import {
-  Column,
-  Entity,
-  Index,
-  JoinColumn,
-  OneToMany,
-  OneToOne,
-} from 'typeorm';
+import { Column, Entity, JoinColumn, OneToMany, OneToOne } from 'typeorm';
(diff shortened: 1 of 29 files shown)


hero101 (approved):
generated a clean migration on local and acceptance.

PR #4535, opened 2024-09-12
Active Subscriptions (license plans) exposed under Account

bobbykolev:
As part of: https://github.com/alkem-io/client-web/issues/6844

This PR provides the required `subscriptions` on the Account Entity (to be used by Org/Usr).
Similar to what we have on Spaces, we get the subscriptions from the agent's credentials.  

This implementation was required to allow the Admins of an Org to be granted relevant privileges for entity creation under that Organization's Account (spaces, VCs, etc).

+ One unused method was deleted (Neil spotted it) 🙏

## Summary by CodeRabbit

- **New Features**
	- Introduced a new abstract class for managing organization subscriptions, including fields for subscription name and expiry date.
	- Added a resolver field to retrieve active subscriptions associated with an organization.
	- Enhanced subscription management capabilities within the organization service, allowing for improved account and subscription retrieval.
	- Added a new abstract class for managing account subscriptions, with fields for subscription name and expiry date.
	- Introduced a resolver method to fetch active subscriptions for accounts.

- **Bug Fixes**
	- Removed unused imports and improved logic for fetching subscriptions to ensure accurate data retrieval.


techsmyth (changes requested):
Thanks Bobby. The code in itself looks fine, but the conceptual model is different. The subscription is on the Account, not on the Organization. So it subscriptions field is on Account entity. Rational is that a User could also have a subscription. It is the Account that is common and has this.

Also I would suggest that we there can just be one LicenseSubscription type defined, replacing the SpaceSubscription / AccountSubscription. The information returned is essentially just a LicenseCredential + expiry date which is the same in both situations. The actual setup will need to change more as not great to return a Credential - but will need to address that later (so not part of this PR).

bobbykolev:
> Thanks Bobby. The code in itself looks fine, but the conceptual model is different. The subscription is on the Account, not on the Organization. So it subscriptions field is on Account entity. Rational is that a User could also have a subscription. It is the Account that is common and has this.
> 


Thanks for the suggestions and clarifications @techsmyth. 
On the subscription under account and not under Org, I agree, it's taken from the Account, but it's exposed under the organization as we need it only there. I'm afraid that if we expose it under the Account, we might end up with the situation where we can't retrieve Account under Organization - https://github.com/alkem-io/server/issues/4537

If you insist I can move the subscriptions under the Account and retrieve them from there, although we don't have access in the Orgs paginated list for now. 


_____
UPDATE: after the fix in https://github.com/alkem-io/server/issues/4537
The PR was reworked to expose subscriptions under Account as it was suggested.

techsmyth (approved):
Working nicely :)

PR #4581, opened 2024-10-02
Cleanup roles indexes

valentinyanakiev:
- npm run migration:generate was creating those differences

- assuming it was from the roles PR.

- we should always make sure schema is synced

## Summary by CodeRabbit

- **Chores**
	- Cleaned up specific database indexes to enhance performance and maintainability.
	- Implemented a migration that allows for easy rollback of changes if needed.


valentinyanakiev:
This doesn't pass but there is clearly a problem with entity definitions + schema. Needs to be fixed asap.

ccanos (approved):

PR #4589, opened 2024-10-04
AuthenticationType enum added and resolver added to User

valentinyanakiev:
- Added `AuthenticationType` under user
- Refactored `createAgentInfo` - strongly typed, smaller methods, more readable

## Summary by CodeRabbit

- **New Features**
	- Introduced an `AuthenticationType` enum for various authentication methods.
	- Added a new field resolver to retrieve the user's authentication method.
	- Implemented methods to retrieve authentication type and identity by email in the user service.
	- Launched the `KratosAdapter` module for enhanced identity management integration.

- **Improvements**
	- Enhanced `AgentInfo` class with an optional `authenticationType` property.
	- Streamlined the `AuthenticationService` with new helper methods for better maintainability.
	- Improved organization of the `OryDefaultIdentitySchema` by introducing the `OryTraits` interface.

These changes enhance user experience by providing clearer authentication options and improving data retrieval related to user authentication methods.

### DIFF
--- a/src/common/enums/authentication.type.ts
+++ b/src/common/enums/authentication.type.ts
@@ -1,0 +1,12 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum AuthenticationType {
+  LINKEDIN = 'linkedin',
+  MICROSOFT = 'microsoft',
+  EMAIL = 'email',
+  UNKNOWN = 'unknown',
+}
+
+registerEnumType(AuthenticationType, {
+  name: 'AuthenticationType',
+});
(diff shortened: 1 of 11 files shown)


techsmyth (changes requested):
Logic is working when applied to the current user. But not for retrieving the authentication method for another user - which is the primary driver i.e. I want to be able to get the authentication method used by all Alkemio org associates.

Right now if I run a query across all users it returns the auth method for the current user (admin@alkem.io) for every user:

image


But the neil user in the above is LinkedIn. 

This almost by definition cannot be anchored on the current session. It has to either a) be stored with the user profile or b) retrieve dynamically from Kratos. Latter is clearly better - and immediately covers all existing users.

techsmyth (approved):

PR #4592, opened 2024-10-04
Organization verified status filter

hero101:


bobbykolev (approved):
Looks and works fine. 👏
As communicated, organizations that are unverified after they were verified are still returned as verified. This will be fixed separately.

PR #4604, opened 2024-10-11
Optionally delete identity with user profile

valentinyanakiev:
- deleted KratosAdapter - it really is not an adapter, but an external service (it does not adapt anything, we don't have a business interface to adapt to). Replaced with KratosService in Infrastructure.
- refactored Kratos client usage to be via KratosService - for extending sessions, for listing login types, for deleting accounts now
- added optional deletion of kratos identity when the user profile is deleted - defaults to true. Added the option so that if there are users that mismatch the current kratos DB we may want to skip this option

## Summary by CodeRabbit

- **New Features**
	- Introduced a new exception for user identity not found.
	- Added a new module and service for improved session management.
	- Enhanced user deletion functionality with an optional identity deletion feature.
	- Added new methods for session validation and identity management.

- **Bug Fixes**
	- Updated session management methods to improve reliability and performance.

- **Chores**
	- Removed deprecated adapter module and associated methods to streamline authentication processes.


techsmyth (approved):
Working nicely

PR #4641, opened 2024-10-25
fix subspace-admin credential removal + small refactor

valentinyanakiev:
- optimized a redundant check on figuring out whether the user is a subspace-admin (done with one DB call)
- fixed a typo
- fixed wrongly passed roleSetId to credential removal logic

## Summary by CodeRabbit

- **New Features**
	- Improved role management functionality, enhancing the efficiency of role set operations.
  
- **Bug Fixes**
	- Corrected the logic for removing contributors from roles, ensuring proper handling of admin roles. 

- **Refactor**
	- Renamed method for clarity and simplified return statements for better performance.


Comoque1:
When a user has been only a member of space/subspace/subsubspace and leave them, he successfully leaves the communities.
When a user has been an admin and a member of a subspace and leave all journeys he still has subspace-admin credentials(even, after it has been removed from the community)
image

src/domain/access/role-set/role.set.service.ts
```diff
@@ -756,8 +756,8 @@ export class RoleSetService {
       );
 
       if (!hasAnotherAdminRole) {
-        await this.removeContributorToImplicitRole(
-          parentRoleSet,
+        await this.removeContributorFromImplicitRole(
+          roleSet,
```
techsmyth: Actually this must remain the parent role set, as that is where the "subspaceAdmin" role makes sense. 

So the logic is that if the user is no longer an admin in the subspace, and that was the last admin role, then the user is no longer a "subspace admin" in the parent space

PR #4652, opened 2024-10-31
Schema setup

valentinyanakiev:
What has been done (high level):

- Migration is split into two:
  - First the schema is created
  - Then the initial data is populated
- Schema is re-created if there are no migration in the `migrations_typeorm` table.
- Data migration is executed if there is only one prior migration (the schema migration).
- Re-created the auth policy migration with new timestamp to be executed last (so there is no adverse effect)

- **Data Initialization**
  - **Migration**
    - **platform**
      - library
      - templates manager
      - licensing framework
    - ai server
    - preferenceDefinitions
  - **Bootstrap**
    - Added license plans to the bootstrap

 

Areas affected & to be tested:
  - library
  - platform templates
  - ai server (engines, if one is working they will all be working)
  - license plans
  - preferences


valentinyanakiev:
Converting to draft so we don't merge it for this release @Comoque1

techsmyth (approved):
Worked smoothly for me. Nice setup.

Only comment is that I think we should move the bootstrap related data to be under the bootstrap module. There is no real value in having it down under configuration/templates. So that means moving the bootstrap users + bootstrap license plans both to be under bootstrap module. Makes the maintenance + being able to navigate simpler - plus these files are actually not exposed through configuration under platform api.

Comoque1:
All api tests are pass (except the ones for which we have bugs )

PR #4702, opened 2024-11-15
fix demo space search visibility

valentinyanakiev:
- filter out DEMO spaces when we have `onlyPublicResults` (we are unauthenticated)
- so far the filter was only an intersection of indices, but it needed to go deeper than that, based on the state of the `Space`

## Summary by CodeRabbit

- **New Features**
	- Enhanced search functionality to filter results based on visibility with the new `onlyPublicResults` parameter.
- **Bug Fixes**
	- Improved error handling for space identification failures, providing detailed logs when a space is not found.


valentinyanakiev:
chose different approach

Issue #4650, opened 2024-10-29
Copy documents when creating a Collaboration Template from an existing Collaboration


ccanos:
As an Alkemio user I want the documents of my template to not be shared with an existing space and vice-versa.

** Context **

When saving collaboration as template, the server just copies the contents of the profiles of the callouts.
That includes the urls of the images in the markdowns, any references in the profiles or any references on profiles (this I think cannot happen, but may in the future). 
Those documents are still in the storageBucket of the original collaboration. They need to be duplicated and stored as documents of the new collaboration of the template, and use the storageBucket of the profile of the template.

** Acceptance Criteria **
- When a user creates a Collaboration Template from an existing collaboration all the images, documents, references, are stored in the new StorageBucket
- When a subspace is created from a template, all the documents are also copied to the new StorageBucket. 

** Testing **
- Create a subspace that will be converted to a template.
- Add a callout to it, add an image to the markdown and a reference to the profile
- Create a template from that subspace
   - That template should have a different url for that image and for that reference
- Select that template as the default for subspace creation
  - That new subspace should have different urls too
- Delete the first subspace. The image&ref in the template and in the new subspace should remain available
- Select a different default template for subspace creation, any other
- Delete the subspace template. The image&ref in the new should remain available

Comoque1:
The flow described above is tested OK on DEV env. Now each new created entity from template stores the files separately.
A bug was raised, that the template files are stored in a wrong storage bucket: https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/server/4861

Issue #4649, opened 2024-10-29
BUG: Server doesn't copy whiteboard visuals on Create Template from Collaboration


ccanos:
**Describe the bug**
When creating template from a collaboration, the profile visuals of the whiteboards are not copied to the profiles created of the new whiteboard in the template.
**Also fix when using a subspace template (import or create subspace using template) is not cloning the visuals of the wbs.** So the subspace has collaboration tools without previews, which is not great UX.

**To Reproduce**
1. Create a space or a subspace
2. Add a whiteboard callout on it
3. Edit the whiteboard, so it gets a whiteboard visual
4. Create a template from that collaboration
5. The whiteboard doesn't get visual

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Areas that will be affected as part of the fix**
_Optional, to be added by developer if relevant_

Comoque1:
Tested on DEV env.
The part with the subspace template works OK.
The WB - split into a separate issue: https://app.zenhub.com/workspaces/alkemio-development-5ecb98b262ebd9f4aec4194c/issues/gh/alkem-io/client-web/7505

PR #4763, opened 2024-12-09
Handle Visuals, References and attached images into markdowns on Create entities #4649 and #4650

ccanos:
Issues: #4649  and #4650 
- Copies the visuals urls when creating an entity.
- Also checks the content of the Markdowns and the References when creating other entities, and copies the documents.

Done:
- Changed CreateWhiteboardInput object to be able to receive a profile with the visuals of the whiteboard
- Changed `addVisualOnProfile` to process received visuals from the client and re-upload them to the profile just created.
   - (only already alkemio documents, not external urls will be called)
- removed `visualUri` and `avatarURL` from some dtos that were not used anywhere

Notes / doubts:
- Profile module now depends on Document module, to able to reupload the visuals.
- whiteboard.dto.create does not extend `CreateNameableInput` anymore, it has a profile and a NameID properties like callout.dto.create, to be able to pass those visuals when creating the WB.
   - Specially needed that to be able to get the visuals from the inputCreator service

## Summary by CodeRabbit

- **New Features**
	- Enhanced visual handling in various services, allowing for multiple visuals to be added to profiles.
	- Introduced new input types for visual data in profile creation.
	- Added new methods for document re-uploading and management in profiles.

- **Bug Fixes**
	- Improved error handling across multiple services, ensuring better validation and exception management.

- **Documentation**
	- Updated method documentation to reflect changes in parameters and functionality.

- **Chores**
	- Removed unnecessary dependencies and streamlined module imports.


src/domain/common/profile/profile.service.ts
```diff
@@ -219,14 +225,90 @@ export class ProfileService {
...
 
     profile.visuals.push(visual);
     return profile;
   }
 
+  /***
+   * Checks if a url is living under the storage bucket
+   * of a profile and re-uploads it if not there
+   */
+  public async reuploadVisualToProfile(
```
ccanos: practically copied from `reuploadDocumentsIfNotInBucket` on `whiteboard.service`

src/domain/storage/storage-bucket/storage.bucket.service.ts
```diff
@@ -260,23 +260,29 @@ export class StorageBucketService {
     }
   }
 
-  public async addDocumentToBucketOrFail(
-    storageBucketId: string,
+  public async addDocumentToStorageBucketOrFail(
```
ccanos, line 263: I needed a function like `addDocumentToBucketOrFail` but passing the bucket not just the id, because my bucket is not yet saved to database here

src/domain/common/profile/profile.service.ts
```diff
@@ -48,11 +55,11 @@ export class ProfileService {
 
   // Create an empty profile, that the creating entity then has to
   // add tagets / visuals to.
-  public createProfile(
+  public async createProfile(
```
ccanos, line 56: we had already `await this.profileService.createProfile`  in many places

src/domain/common/profile/profile.service.ts
```diff
@@ -188,43 +213,155 @@ export class ProfileService {
     return await this.profileRepository.save(profile);
   }
 
-  public addVisualOnProfile(
+  public async addVisualsOnProfile(
```
ccanos, line 223: made it a loop to call it just once

src/domain/common/profile/profile.service.ts
```diff
@@ -188,43 +213,155 @@ export class ProfileService {
...
+        case VisualType.BANNER_WIDE:
+          visual = this.visualService.createVisualBannerWide();
+          break;
+
+        default:
+          throw new Error(
+            `Unable to recognise type of visual requested: ${visualTypes}`
+          );
+      }
+      const providedVisual = visualsData?.find(v => v.name === visualType);
```
ccanos, line 256: and added the re-upload functionality

src/domain/common/profile/profile.service.ts
```diff
@@ -66,16 +74,15 @@ export class ProfileService {
     profile.storageBucket = this.storageBucketService.createStorageBucket({
       storageAggregator: storageAggregator,
     });
-
+    profile.description = await this.reuploadDocumentsInMarkdownProfile(
+      profile.description ?? '',
+      profile.storageBucket
+    );
```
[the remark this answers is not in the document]
ccanos: storageBucketService.createStorageBucket is not asynchronous

src/common/enums/visual.type.ts
```diff
@@ -10,3 +10,11 @@ export enum VisualType {
 registerEnumType(VisualType, {
   name: 'VisualType',
 });
+
+export function validateAndConvertVisualTypeName(name: string): VisualType {
+  if (Object.values(VisualType).includes(name as VisualType)) {
+    return name as VisualType;
+  } else {
+    throw new Error(`Invalid VisualType: ${name}`);
```
valentinyanakiev: use a more concrete `Exception`, not generic `Error`

src/domain/common/profile/dto/profile.dto.create.visual.ts
```diff
@@ -0,0 +1,15 @@
...
+@InputType()
+@ObjectType('CreateVisualOnProfileData')
+export class CreateVisualOnProfileInput {
+  @Field(() => VisualType, {
+    nullable: false,
+    description: 'The type of visual.',
+  })
+  name!: VisualType;
+
+  @Field(() => String, { nullable: false })
```
valentinyanakiev: add description

src/services/api/input-creator/input.creator.service.ts
```diff
@@ -283,6 +301,26 @@ export class InputCreatorService {
     };
   }
 
+  private buildCreateVisualsOnProfileInputFromVisuals(
```
valentinyanakiev, line 334: You can change those two methods to 

```
private buildCreateVisualsOnProfileInputFromVisuals(
  visuals?: IVisual[]
): CreateVisualOnProfileInput[] {
  if (!visuals) return [];
  return visuals.map(visual => ({
    name: validateAndConvertVisualTypeName(visual.name),
    uri: visual.uri,
  }));
}
```

src/services/api/input-creator/input.creator.service.ts
```diff
@@ -184,7 +199,7 @@ export class InputCreatorService {
   ): CreateWhiteboardInput | undefined {
     if (!whiteboard) return undefined;
     return {
-      profileData: this.buildCreateProfileInputFromProfile(whiteboard.profile),
+      profile: this.buildCreateProfileInputFromProfile(whiteboard.profile),
```
valentinyanakiev, line 232: I am not sure about this renaming, you are building an input DTO from the profile and name it profile, but pass the DTO? Looks very strange to me.

src/domain/storage/document/document.service.ts
```diff
@@ -188,7 +188,9 @@ export class DocumentService {
 
     const documentID = url.substring(documentsBaseUrlPath.length + 1);
     try {
-      return await this.getDocumentOrFail(documentID);
+      return await this.getDocumentOrFail(documentID, {
+        relations: { storageBucket: true },
```
valentinyanakiev: Do you really need this relation everywhere? Why not have the options as optional parameter on `getDocumentFromURL`?

src/domain/storage/storage-bucket/storage.bucket.service.ts
```diff
@@ -260,23 +260,29 @@ export class StorageBucketService {
...
     );
     return this.documentService.save(document);
   }
 
+  public async addDocumentToStorageBucketByIdOrFail(
+    storageBucketId: string,
+    document: IDocument
+  ): Promise<IDocument> | never {
+    const storageBucket = await this.getStorageBucketOrFail(storageBucketId);
+    return this.addDocumentToStorageBucketOrFail(storageBucket, document);
```
valentinyanakiev, line 285: deliberately missing the await?

src/domain/storage/storage-bucket/storage.bucket.service.ts
valentinyanakiev: You end up saving the document before saving the storage bucket, which is quite dangerous, as you leave the behavior to be implicit, based on entity relationships:

Some explanation by ChatGPT:

In TypeORM, if you have an entity `A` with a relation to entity `B` but you only save `B` without saving `A`, the following happens depending on the type of relation and the cascade settings:

---

### **1. Without Cascading Save**
If `A` is not saved and no cascade options are configured, `B` is saved independently, and the relationship between `A` and `B` is not persisted. Specifically:

- The `B` record is saved to the database.
- The database does not associate `B` with `A` since `A` does not exist in the database.
- The foreign key field on `B` (or the join table, depending on the relation type) remains `NULL` or unaffected.

---

### **2. With Cascading Save**
If the relationship is configured with `cascade: true`, TypeORM attempts to save `A` when saving `B`. However, because `A` is not explicitly saved or persisted in the database yet, one of two things can happen:

- If `A` is a new (unsaved) object, TypeORM tries to insert `A` into the database before saving `B`.
- If `A` is undefined or null, no save occurs for `A`.

---

### **Relation Types and Behavior**
#### **One-to-One**
- If `A` references `B` in a one-to-one relationship and `A` is not saved, the foreign key on `B` referencing `A` will remain `NULL` or invalid.

#### **One-to-Many / Many-to-One**
- In a one-to-many relationship where `A` has many `B`s, saving `B` does not affect `A`. If the foreign key linking `B` to `A` depends on `A` being saved, the relationship will not be persisted.

#### **Many-to-Many**
- If `A` and `B` have a many-to-many relationship, saving `B` does not create an entry in the join table unless both entities are saved.

---

### **Example**
#### **Entity Definitions**
```typescript
@Entity()
class A {
  @PrimaryGeneratedColumn()
  id: number;

  @OneToOne(() => B, (b) => b.a, { cascade: true })
  @JoinColumn()
  b: B;
}

@Entity()
class B {
  @PrimaryGeneratedColumn()
  id: number;

  @ManyToOne(() => A, (a) => a.b)
  a: A;
}
```

#### **Code**
```typescript
const b = new B();
await dataSource.getRepository(B).save(b);
```

#### **Outcome**
- `b` is saved, but no `a` exists in the database. 
- The `aId` foreign key in the `B` table remains `NULL`.

---

### **Best Practices**
1. **Explicit Saves**: Always save the "owning side" of the relationship explicitly to ensure the relationship is persisted.
2. **Cascading**: Use `cascade: true` cautiously, as it can lead to unintended side effects, such as saving incomplete entities.
3. **Validation**: Validate the presence and state of related entities before saving.

src/domain/common/whiteboard/dto/whiteboard.dto.create.ts
```diff
@@ -1,11 +1,24 @@
...
-import { CreateNameableInput } from '@domain/common/entity/nameable-entity/dto/nameable.dto.create';
+import { CreateProfileInput } from '@domain/common/profile/dto';
+import { Type } from 'class-transformer';
+import { NameID } from '@domain/common/scalars';
 
 @InputType()
 @ObjectType('CreateWhiteboardData')
-export class CreateWhiteboardInput extends CreateNameableInput {
+export class CreateWhiteboardInput {
+  @Field(() => CreateProfileInput, { nullable: true })
```
valentinyanakiev, line 11: Not sure why this was needed like that instead of the inheritance?

src/domain/common/profile/profile.service.ts
```diff
@@ -188,43 +220,153 @@ export class ProfileService {
...
+          break;
+        case VisualType.CARD:
+          visual = this.visualService.createVisualCard();
+          break;
+        case VisualType.BANNER_WIDE:
+          visual = this.visualService.createVisualBannerWide();
+          break;
+
+        default:
+          throw new Error(
```
valentinyanakiev: Use meaningful exception

src/domain/common/profile/profile.service.ts
```diff
@@ -188,43 +220,153 @@ export class ProfileService {
...
+  }
+
+  /***
+   * Checks if a url is living under the storage bucket
+   * of a profile and re-uploads it if not there
+   * @param fileUrl The url of the file to check
+   * @param storageBucket The StorageBucket in which the file should be
+   * @param alkemioRequired If true, the file must be inside Alkemio and if a fileUrl passed is outside Alkemio function will return undefined
+   */
+  private async reuploadFileOnStorageBucket(
```
valentinyanakiev: Add those two files in a separate service, split them and generate unit tests over them. Inherently algorithmic, but also a bad practice to have such long methods.

src/domain/common/profile/profile.service.ts
```diff
@@ -188,43 +220,153 @@ export class ProfileService {
...
+        LogContext.PROFILE
+      );
     }
 
-    profile.visuals.push(visual);
-    return profile;
+    const docInContent = await this.documentService.getDocumentFromURL(fileUrl);
+
+    if (!docInContent) {
+      throw new BaseException(
```
valentinyanakiev: there is definitely a NotFound exception or something, if there isn't - add it.

src/domain/common/profile/profile.service.ts
valentinyanakiev: Have you checked whether it is awaited `everywhere`?

valentinyanakiev (changes requested):

src/domain/common/profile/profile.service.ts
```diff
@@ -219,14 +225,90 @@ export class ProfileService {
...
+        providedVisual.uri,
+        profile
+      );
+      if (url) {
+        visual.uri = url;
+      } else {
+        this.logger.warn(
+          `Visual with URL '${providedVisual.uri}' ignored when creating profile ${profile.id}`
+        );
+      }
```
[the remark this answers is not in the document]
ccanos: undefined uri is alright

src/domain/common/profile/profile.service.ts
ccanos: yes, for sure

src/domain/storage/storage-bucket/storage.bucket.service.ts
ccanos: yeah, this is already returning the promise that I need to return, so if anyone wants to await after this return is fine, no?

ccanos:
@coderabbitai full review

src/domain/common/whiteboard/whiteboard.service.ts
```diff
@@ -52,12 +52,18 @@ export class WhiteboardService {
...
+      },
       ProfileType.WHITEBOARD,
       storageAggregator
     );
-    this.profileService.addVisualOnProfile(whiteboard.profile, VisualType.CARD);
+    await this.profileService.addVisualsOnProfile(
+      whiteboard.profile,
+      whiteboardData.profile?.visuals,
+      [VisualType.CARD]
+    );
```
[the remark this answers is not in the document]
ccanos: this is not true, when you create a profile the function `createProfile` always creates a StorageBucket and an empty array of visuals.

src/domain/common/whiteboard/dto/whiteboard.dto.create.ts
```diff
@@ -1,11 +1,24 @@
...
+  @ValidateNested({ each: true })
```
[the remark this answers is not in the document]
ccanos: @valentinyanakiev I have changed this as suggested, but this was copied from somewhere else. Doing a quick search I have seen a few places with `({ each: true })` which are not arrays. Do you want me to create a task to change them?

src/domain/common/whiteboard/dto/whiteboard.dto.create.ts
ccanos: the thing is `CreateNameableInput` is an `@InputType` but not an `@ObjectType`, so GraphQL was not returning the profile information, only the content field:
image
image

I saw that in framing for example it's made like this: `CreateCalloutFramingData` has `profile` instead of inheriting from NameableInput:
image

But for the whiteboard we don't have that profile available so I couldn't push the visuals into that profile when creating the whiteboard of the framing of the callout:
image

valentinyanakiev (changes requested):
Please review my comments and resolve / let's comment the ones outstanding. I don't believe this is error-less implementation and there are additional tests failing (213 on your branch, 45 'expected' to fail on dev).

src/domain/space/space/space.service.ts
```diff
@@ -246,9 +246,21 @@ export class SpaceService {
...
+    await this.profileService.addVisualOnProfile(
+      space.profile,
+      VisualType.BANNER,
+      spaceData.profileData.visuals
+    );
+    await this.profileService.addVisualOnProfile(
+      space.profile,
+      VisualType.CARD,
+      spaceData.profileData.visuals
+    );
```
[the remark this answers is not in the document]
ccanos: addressed

src/domain/common/profile/profile.service.ts
```diff
@@ -40,6 +46,8 @@ export class ProfileService {
     private tagsetTemplateService: TagsetTemplateService,
     private referenceService: ReferenceService,
     private visualService: VisualService,
+    private documentService: DocumentService,
+    private documentAuthorizationService: DocumentAuthorizationService,
```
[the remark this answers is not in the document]
ccanos: addressed

src/common/enums/visual.type.ts
ccanos: addressed

src/domain/storage/document/document.service.ts
ccanos: addressed

src/domain/community/virtual-contributor/virtual.contributor.service.ts
```diff
@@ -119,7 +119,7 @@ export class VirtualContributorService {
     virtualContributor.aiPersona =
       await this.aiPersonaService.createAiPersona(aiPersonaInput);
 
-    virtualContributor.profile = this.profileService.createProfile(
+    virtualContributor.profile = await this.profileService.createProfile(
```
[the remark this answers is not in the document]
valentinyanakiev: it is actually async, so you are wrong

valentinyanakiev:
I can confirm updated test suites pass against this branch

src/domain/storage/storage-bucket/storage.bucket.service.ts
valentinyanakiev: @coderabbitai please raise a task about it on the backlog

valentinyanakiev (approved):

PR #4766, opened 2024-12-10
Async invoke all engines ingest

valentinyanakiev:
- invoke ingest all

### DIFF
--- a/src/services/ai-server/ai-persona-service/ai.persona.service.service.ts
+++ b/src/services/ai-server/ai-persona-service/ai.persona.service.service.ts
@@ -197,6 +197,15 @@ export class AiPersonaServiceService {
     });
   }
 
+  public async invokeIngestAll(): Promise<void> {
+    const aiPersonaServiceIDs = (
+      await this.aiPersonaServiceRepository.find({
+        select: ['id'],
+      })
+    ).map(aiPersonaService => aiPersonaService.id);
+    return this.aiPersonaEngineAdapter.invokeIngestAll(aiPersonaServiceIDs);
+  }
+
   private encryptExternalConfig(
     config: IExternalConfig | undefined
   ): IExternalConfig {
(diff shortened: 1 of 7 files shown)


src/services/ai-server/ai-persona-service/ai.persona.service.service.ts
```diff
@@ -197,6 +197,15 @@ export class AiPersonaServiceService {
     });
   }
 
+  public async invokeIngestAll(): Promise<void> {
```
valeksiev: This method should be in `AiPersonaService` (collab server level) and not in the `AiPersonaServiceService` (ai server level).

The `AiPersonaServeice` has an instance of the `AiServerAdapter` and something like:
```
for (const aiPersona of aiPersonas) {
      const personaService = await this.aiServerAdapter.getPersonaServiceOrFail(
        aiPersona.aiPersonaServiceID
      );
      this.aiServerAdapter.refreshBodyOfKnowledge(personaService.id);
}
```

should do the trick. This is less code but suboptimal - lot's of DB hits. 

If you push the method down to the server it should be something like:

`aiPersonaService.refresllBoKAll() -> aiServerAdapter.refreshBokAll() -> aiServerService.refreshBokAll() -> aiPersonaServiceService.refreshAll()`

where:
```
// aiPersonaServiceService.refreshAll()
public async refreshAll() {
    const aiPersonaServices = await this.aiPersonaServiceRepository.find({
      select: ['id', 'bodyOfKnowledgeID'],
    })
    for (const personaService of aiPersonaServices) {
        this.eventBus.publish(
          new IngestSpace(
            savedAiPersonaService.bodyOfKnowledgeID,
            SpaceIngestionPurpose.KNOWLEDGE,
            savedAiPersonaService.id
          )
      );

    }
}
```

valentinyanakiev: Affirmative, moving it to aiPersonaService

PR #4775, opened 2024-12-11
Dataloader auto resolve

hero101:
Dataloaders now auto-resolves for nullable graphql fields as the default behaviour when resolving the unresolved result for a key.
This is useful when the result is expected to be null, and it's not an exceptional case.

The default behaviour is determined by the Loader decorator on the dataloader`s creation.
If the underlying graphql field, for which the dataloader is created, is nullable - the result can be resolved to null.
Example:
```
@ResolveField(() => ICallout, {
  nullable: true,
  description: 'The Callout that was published.',
})
public callout(
  @Parent() { payload }: InAppNotificationCalloutPublished,
  @Loader(CalloutLoaderCreator) loader: ILoader<ICallout | null>
) {
  return loader.load(payload.calloutID);
}
```
The `callout` is decorated as nullable, so the dataloader will auto-resolve to `null` if the result is not found.
You can override this behaviour by setting the `resolveToNull` flag to `false`. That way the problematic values will always be resolved to errors.

### DIFF
--- a/src/core/dataloader/creators/base/data.loader.creator.base.options.ts
+++ b/src/core/dataloader/creators/base/data.loader.creator.base.options.ts
@@ -29,10 +29,28 @@ export interface DataLoaderCreatorBaseOptions<TParent, TResult> {
    */
   cache?: boolean;
   /***
-   * What to return when resolving the unresolved result for a key.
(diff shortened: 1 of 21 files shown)


hero101:
closing in favor of https://github.com/alkem-io/server/pull/4776

PR #4788, opened 2024-12-14
Authorization: remove usage of anonymousReadAccess

techsmyth:
Triggered by work on READ_ABOUT...where we will need more control that the simple flag that controls READ access.

Add new global credential GLOBAL_ANONYMOUS
Assign this credential to AgentInfos that are representing anonymous usage

Removed flag from authorization policy
Added utility method to assign a credential rule for anonymous usage with a specified privilege to an authorization policy

Updated Space sorting to use privacy mode (public/private) instead of anonymousReadAccess (basically fixing a potential bug). Potentially the new setup is also faster.

Reworked handling of AgentInfo to bring the creation / updating of this data structure to be in one place

Client PR is already merged

Initial results:
image

Todo:
- check for edge cases by having test suites passing; there may be cases where something is over ruled in the child previously which the new setup will not cover
- update unit tests on Space: they were using anonymousReadAccess but now should be using settings.privacy.mode = Public/Private
- see if can easily share the logic for generating an AgentInfo for anonymous user requests, currently duplicated 3x. Problems with module dependency in one case, plus second case where have functional vs service usage.
- check with light testing the impact this has on performance; risk is that it is heavier
- test test test

## Summary by CodeRabbit

- **New Features**
  - Introduced a new enumeration value `GLOBAL_ANONYMOUS` for non-authenticated interactions.
  - Added `AgentInfoService` to enhance agent information management.
  - Implemented methods for creating and managing anonymous agent information.

- **Bug Fixes**
  - Removed redundant `anonymousReadAccess` properties from various authorization contexts.

- **Refactor**
  - Updated logic for handling anonymous read access across multiple services and modules to use method calls instead of direct property assignments.

- **Documentation**
  - Adjusted JSON structures in test data to reflect changes in authorization handling.

### DIFF
--- a/src/common/enums/authorization.credential.ts
+++ b/src/common/enums/authorization.credential.ts
@@ -5,6 +5,7 @@ export enum AuthorizationCredential {
   GLOBAL_ADMIN = 'global-admin', // able to do everything, god mode
   GLOBAL_SUPPORT = 'global-support', // able to manage platform level information, can per space have admin rights
   GLOBAL_LICENSE_MANAGER = 'global-license-manager', // able to manage platform level information, can per space have admin rights
+  GLOBAL_ANONYMOUS = 'global-anonymous', // credential issued to all non-authenticated interactions
   GLOBAL_REGISTERED = 'global-registered', // credential issued to all registered users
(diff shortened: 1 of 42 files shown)


techsmyth:
@coderabbit review

PR #4809, opened 2024-12-27
VC KnowledgeBase; CalloutsSet module from Collaboration

techsmyth:
## CalloutsSet entity
Split CalloutsSet entity out from Collaboration so that it is then re-used in a KnowledgeBase entity (good from a maintainability point of view as the Collaboration entity was huge...)
image

Dealt with module dependency mess, needed a lot of refactoring to be able to get the server working without circular depdendencies. 
- no longer any forwardRefs being used. But there is a more structural design needed near term. 

Migration is added + tested on simple DB

Client is updated and create space + add callout flow is working.

Test Suites are updated re codegen + compiling

## KnowledgeBase on Virtual Contributor
New entity added to the domain model.
Knowledge Base added to all VCs
image

Logic related to usage of CalloutsSet updated to reflect now more than one location.

Able to create a Callout on the Knowledge Base is tested + working

When creating a VC of type knowledge base then the platform template for this is picked up.

### Testing the new VC functionality
From graphql, trigger a create VC mutation with the type set to knowledge base:

image

## To do:
- validate all the different callout creation flows
- check events etc that are triggered come through ok

Later:
- will also need to modify the platform template to only have the callouts that make sense in the new setup. Ideally also get the templates manager at platform level working to make that a reasonable effort. 

https://alkem.io/building-alkemio-org/challenges/virtualcontributors-54/collaboration/vcbodyofknowledge


bobbykolev (approved):
The volume of changes in this review is huge and looking into every file is going to take many hours. 
I've done testing of the main changes - calloutSet-related - and everything works as expected.
I didn't test the e2e Expert integration. In case of issues, we'll follow up.

⚠️ Re: tests - the test project is in the middle of a huge refactoring. Once stabilized, the collaboration test has to be fixed. Suggested by @techsmyth. @Comoque1 FYI

PR #4820, opened 2025-01-06
Update the VC Knowledge base description

bobbykolev:
Allows update of the description of the Knowledge base of VC. 
Client work in https://github.com/alkem-io/client-web/pull/7388

## Summary by CodeRabbit

- **New Features**
  - Added a new optional field for knowledge base data in the virtual contributor update process.
  - Enhanced virtual contributor data retrieval to include knowledge base information.

- **Bug Fixes**
  - Updated error messages to correctly reference "callout" instead of "whiteboard" for better context.

### DIFF
--- a/src/domain/community/virtual-contributor/virtual.contributor.service.ts
+++ b/src/domain/community/virtual-contributor/virtual.contributor.service.ts
@@ -203,7 +203,9 @@ export class VirtualContributorService {
       {
         relations: {
           profile: true,
-          knowledgeBase: true,
+          knowledgeBase: {
+            profile: true,
+          },
         },
       }
     );
--- a/src/domain/community/virtual-contributor/virtual.contributor.service.ts
+++ b/src/domain/community/virtual-contributor/virtual.contributor.service.ts
@@ -201,7 +201,10 @@ export class VirtualContributorService {
     const virtual = await this.getVirtualContributorOrFail(
       virtualContributorData.ID,
       {
-        relations: { profile: true },
+        relations: {
+          profile: true,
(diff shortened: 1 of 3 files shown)


valeksiev (dismissed):

src/domain/community/virtual-contributor/virtual.contributor.service.ts
```diff
@@ -201,7 +201,10 @@ export class VirtualContributorService {
     const virtual = await this.getVirtualContributorOrFail(
       virtualContributorData.ID,
       {
-        relations: { profile: true },
+        relations: {
```
valentinyanakiev, line 204: you are loading         

```
relations: {
          profile: true,
          knowledgeBase: true,
        },
``` 

but I believe it should be

```
        relations: {
          knowledgeBase: {
                 profile: true,
        },
}
```

Also I'd check if the knowledgeBase and its profile are loaded and throw an exception if not, as it's an explicit request to the DB. You can then omit the second check in :
```
if (
      virtualContributorData.knowledgeBaseData?.profile?.description &&
      virtual?.knowledgeBase?.profile
    )
```

It should always be true.

valentinyanakiev (changes requested):

src/domain/community/virtual-contributor/virtual.contributor.service.ts
bobbykolev: As agreed, I'm applying the first suggestion to explicitly add relation to the profile from the knowledgeBase.
We are keeping the check for Knowledge base just in case because of the other types and legacy VCs.

PR #4855, opened 2025-01-20
Avoid circular dependencies on JSON.stringify

ccanos:
- This is related to the last changes about moving documents to the correct StorageBucket when creating an entity.
- It's because now the newly created `StorageBucket` has documents, before it didn't. Those documents point to their parent `StorageBucket` which inside has `documents` which... creates a circular reference.
- The error is thrown from somewhere inside `server/node_modules/graphql-amqp-subscriptions/dist/amqp/publisher.js`
- We are calling it from around the line 278: 
```
  await this.postCreatedSubscription.publish(
    SubscriptionType.CALLOUT_POST_CREATED,
    postCreatedEvent
  );
```
Fix:
- I am deconstructing the post object so what is passed to the subscription doesn't have `StorageBucket`. 
- I have confirmed that the client is not querying the StorageBucket in the subscription.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the data integrity for contributions by ensuring the correct post information is utilized during creation.
	- Improved event publication by updating the post details associated with contributions.

The changes focus on optimizing data handling, ensuring a more reliable experience for users without altering visible functionality.

### DIFF
--- a/src/domain/collaboration/callout-contribution/callout.contribution.service.ts
+++ b/src/domain/collaboration/callout-contribution/callout.contribution.service.ts
@@ -229,7 +229,7 @@ export class CalloutContributionService {
   }
 
   public async getPost(
-    calloutContributionInput: ICalloutContribution,
+    calloutContributionInput: Pick<ICalloutContribution, 'id'>,
     relations?: FindOptionsRelations<ICalloutContribution>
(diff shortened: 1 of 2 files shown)


hero101 (approved):
lgtm

PR #4883, opened 2025-01-27
NameID usage 

techsmyth:
`npm i` required

## Summary by CodeRabbit

- **New Features**
	- Launched enhanced URL resolution that directs users seamlessly to various content types.
	- Introduced new modules for improved innovation pack defaults and forum discussion lookup.
	- Added several GraphQL schema types to enrich API responses.

- **Refactor**
	- Streamlined lookup and retrieval logic by removing redundant identifier validations.
	- Improved error messaging and logging for better troubleshooting.

- **Chore**
	- Consolidated URL path settings into a centralized configuration for greater consistency and easier maintenance.


src/services/api/url-resolver/url.resolver.service.ts
```diff
@@ -0,0 +1,782 @@
...
+import { UrlResolverQueryResultSpace } from './dto/url.resolver.query.space.result';
+import { match } from 'path-to-regexp';
+import { InnovationPackService } from '@library/innovation-pack/innovation.pack.service';
+import { UrlResolverQueryResultCalloutsSet } from './dto/url.resolver.query.callouts.set.result';
+import { InnovationHubService } from '@domain/innovation-hub/innovation.hub.service';
+
+@Injectable()
+export class UrlResolverService {
+  private spacePathMatcher = match(
+    `/:spaceNameID{/${URL_PATHS.CHALLENGES}/:challengeNameID}{/${URL_PATHS.OPPORTUNITIES}/:opportunityNameID}{/*path}`
```
bobbykolev, line 32: I'm concerned that we're coupling the client router too much with the server. I know it's too late and the server is already building the entities URLs. 

So this is more of a conceptual discussion rather than a change request.

Ideally, every client should define its one routes. The infromation that is passed could be in format:
```
{
type: "oportunity"
spaceNameId: "",
challengeNameId: "",
opportunityNameId: ""
}
```
That way if we change a route in the clinet, no changes will be required on the server. It's more robust in my oppinion. 

I thought we're also moving towards naming like spaceNameL0, spaceNameL1, spaceNameL2, instead of challenge & opportunity.

src/services/api/url-resolver/url.resolver.service.ts
```diff
@@ -0,0 +1,782 @@
...
+            );
+          result.discussionId = discussion.id;
+          result.type = UrlType.DISCUSSION;
+        }
+        return result;
+      }
+    }
+
+    // Assumption is that everything else is a Space!
+    await this.populateSpaceResult(result, agentInfo, urlPath);
```
bobbykolev, line 168: We should be very cautious in the client with this assumption.

src/services/api/url-resolver/url.resolver.service.ts
bobbykolev: @coderabbitai can you open a github issue regarding this comment? Please, assign it to me.

ccanos (dismissed):

ccanos (approved):

PR #4956, opened 2025-02-19
Reduce the permissions required to resolve a spaceNameId to a spaceId

ccanos:
There's no task number about this. 
But Evgeni pointed me to a problem in dev:
`"message":"Authorization: unable to grant 'read' privilege: lookup L0 Space by NameID: 59f90753-7f4f-454f-802c-2ceb892da9fe user: on authorization 122c7b32-0fbd-489b-b098-27be80145882 of type 'space'"`
image
On the `/spaces` url we are resolving the space named `welcome-space` to show the Welcome Space on the first position.

- I think anybody with the READ_ABOUT permission should be able to resolve a spaceNameId to id. And not necessarily needs the READ permission.
- Also I think the error message should show the nameId requested


techsmyth (approved):
Agreed, READ_ABOUT is enough

PR #4977, opened 2025-02-25
Space about profile type

techsmyth:
Tidying up a small bit of tech debt, whereby we had multiple types for profiles at the different levels: now all SpaceAbout entities have the same profile type. 

Fixed a typo in the folder name for Oppportunity. 

Note: looks like a lot of commits but that is just due to how the branch merging has gone. 

## Summary by CodeRabbit

- **New Features**
	- Introduced a unified “About” section for spaces, allowing users to view and manage detailed space information in one consolidated view.
	- Enhanced functionality for creating, updating, and deleting space details.

- **Refactor**
	- Streamlined space profile categorization by replacing legacy types with the new “About” type.
	- Improved navigation and URL generation for space information to ensure a more consistent user experience.


src/domain/space/space.about/space.about.resolver.fields.ts
```diff
@@ -15,7 +15,7 @@ import { SpaceAbout } from './space.about.entity';
 export class SpaceAboutResolverFields {
   constructor() {}
 
-  @AuthorizationAgentPrivilege(AuthorizationPrivilege.READ)
```
bobbykolev: I think that this was fixed yesterday by @valentinyanakiev - https://github.com/alkem-io/server/pull/4949/commits/775ee5cca2df9443997ece007f6103208d2fde8f

This validation was failing for some users. 
We either need to add this read_about credential where needed or keep the READ check here. 
@techsmyth

techsmyth: It is ok for now to not have this field protected, as the about field is already protected. So this one should not hold up the PR being merged. Not  least as this whole area will be getting an authorization revamp near term...

bobbykolev: Ok I'll fallback to READ as this was already tested yesterday and you can remove it or revamp it when suitable.

bobbykolev (approved):

PR #4984, opened 2025-02-25
NOTE: TO BE REBASED Fix KB tagset + calloutsSet tagsetTemplate

techsmyth:


Comoque1:
The migration seems to be successful on local env with ACC db.
The User can update the Callouts, delete them, but doesn't have rights to "Contribute" - i.e. if he/she try to add a post or document
The above happens for both migrated and newly created callouts
image

PR #4995, opened 2025-02-28
Innovation Flow in Spaces - Classification tagsets: usage simplification + data migration 

ccanos:
### Problem 1:
On Space Level 0. We were not using innovation_flows at all. 
Regardless of the level, all the spaces have a collaboration, and all the collaborations have an innovation_flow. (at least on accepance. the first migration will fail if this is not true in prod).
Innovation flows have a `innovation_flow > profile > tagsets` tagset with `name='flow-state'` that specifies the current selected state of the Innovation Flow.
- Problem: There are innovation-flows (in acc) that don't have a tagset and that gives an error when you query the currentState. Also gives an error when the state is not one of the 4 valid states (which happens a lot). At space level 0, innovation flows were not used at all so we were not facing these errors in the past.
- The first migration, will go through all level 0 spaces, get their innovation_flow, get their tagsets, set the correct `tags` (current state), and the allowedValues and defaultValue in the tagset_template.
- This `tagset_template` has a `tagsetTemplateSetId` which has to be set to `callouts_set.tagsetTemplateSetId` of the space.
- This problem is bad data, AND more importantly the logic was way to convoluted. 

### Problem 2:
Callouts in Spaces are organized in tabs using a `tagset` with `name='callout-group'`. In subspaces they are organized in those tabs/flow-states with a `tagset` with `name='flow-state'`.
Those two tagsets are linked by the `callout > callout_framing > profile > tagsets`. 
So normally callouts have 3 tagsets: `name = 'default'`, `name = 'callout-group'`, `name = 'flow-state'`.
- On Space L0, I have found many cases (with acc data) of callouts that don't have a tagset with `name = 'flow-state'`. Or that have it incorrect.

## Updated design
To address the above, and to simplify logic that was way too convoluted, the following changes were made.

InnovationFlow:
* added currentState field
* added a direct link to the flowTagsetTemplate
Now if the set of states are changed, the innovation flow directly updates the set of allowed values.

Note: no longer then using the tagset on the innovationFlow.profile - which makes the code *much* simpler to rationalize and should also be a good bit faster.

Callout:
* added a new entity called Classification, which holds the tagsets used for classification. It was simply a bad design (my design) that had us using the tagsets on the callout.framing. Moved the tagsets to this new entity, fixing any bad data in the process.
* removed usage of groups for classifying callouts. So we now only have flow states. Simplifies the logic substantially.

Removed CalloutGroups enum.

Removed also Space.type as hit some weird code in the client that was using the space.type field.

Updated the filtering of callouts to allow specifying the classification to filter on. 

Rational for Classification is that I can see us needing to do more with data classification near term, and also actually makes the code clean.

PENDING:
- For retrieving callouts from the client we use a parameter `groups` at the callouts resolver to filter callouts by group.
- It would be nice to have a filter by state there.
- We have some logic to split callouts in groups that may be quite useful to have also for states.


ccanos:
@coderabbitai review

techsmyth:
@ccanos thanks for the super summary. The current status once this work goes through is that we will I think not be using the groups functionality at all - right? And we have the secondary consideration that flow states is specific to callouts in use within Collaboration: that does not apply to callouts within a knowledge base. A third one is that it feels odd to be using the callout framing profile for the filtering, this feels more like a direct property on the callout. @valentinyanakiev lets discuss this on Tuesday.

techsmyth:
Closing this PR in favor of a new, cleaner one.

PR #5007, opened 2025-03-07
Feature: Flow states, About v2

techsmyth:
## SpaceAbout additional fields

Additional fields exposed on SpaceAbout so that this is the only entity that needs to be queried for information that is visible *outside* the Space.

Goal is that all information that should be accessed on private spaces by non-members should be retrieved via this route. 

This approach addresses a number of issues:
- there is no leakage of information that really should be externally accessed e.g. all settings
- the client just needs to query a spaceAbout, not all the containing entities such as collaboration, community etc. 
- authorization can be much simpler (and likely also faster)

```
query {
  spaces {
    about {
      id
      profile {
        displayName
      }
      provider {
        id
				nameID
      }
      metrics {
        name
        value
      }
      membership {
        myPrivileges
        applicationForm {
          questions {
            question
          }
        }
        myMembershipStatus
      }
    }
  }
}
```

giving 

```
{
  "data": {
    "spaces": [
      {
        "about": {
          "id": "736bd3ef-c224-4fbe-827d-58a860714e7e",
          "profile": {
            "displayName": "Default Space"
          },
          "provider": {
            "id": "cb2ab7d3-3d40-4311-92d1-a791f170976f",
            "nameID": "eco1host"
          },
          "metrics": [
            {
              "name": "members",
              "value": "0"
            }
          ],
          "membership": {
            "myPrivileges": [],
            "applicationForm": {
              "questions": [
                {
                  "question": "What makes you want to join?"
                },
                {
                  "question": "Any particular role or contribution that you have in mind?"
                },
                {
                  "question": "Through which user,organization or medium have you become acquainted with this community?"
                },
                {
                  "question": "Anything fun you want to tell us about yourself?!"
                },
                {
                  "question": "Do you already want to join a Challenge? If so, please provide the name(s)."
                }
              ]
            },
            "myMembershipStatus": "NOT_MEMBER"
          }
        }
      }
    ]
  }
}
```

## Authorization + READ_ABOUT

Authorization on SpaceAbout and KnowledgeBase updated to be much cleaner + simpler:
* if an object contents, such as Space or KnowledgeBase, can only be read *about* then give READ_ABOUT on the object & then give READ on the object which has information about the contents
* if an object  contents can be read, then grant the READ on the base entity and cascade

Above applied to both the Space entity and KnowledgeBase.

------

Simplified the domain model and logic so that Callouts in a CalloutsSet now only work with a single classification parameter: flowState.

InnovationFlow:
* added currentState field
* added a direct link to the flowTagsetTemplate
Now if the set of states are changed, the innovation flow directly updates the set of allowed values.

Note: no longer then using the tagset on the innovationFlow.profile - which makes the code *much* simpler to rationalize and should also be a good bit faster.

Callout:
* added a new entity called Classification, which holds the tagsets used for classification. It was simply a bad design (my design) that had us using the tagsets on the callout.framing. Moved the tagsets to this new entity, fixing any bad data in the process.
* removed usage of groups for classifying callouts. So we now only have flow states. Simplifies the logic substantially.

Removed CalloutGroups enum.

Removed also Space.type as hit some weird code in the client that was using the space.type field.

Updated the filtering of callouts to allow specifying the classification to filter on. 

Rational for Classification is that I can see us needing to do more with data classification near term, and also actually makes the code clean.

### Problem 1:
On Space Level 0. We were not using innovation_flows at all. 
Regardless of the level, all the spaces have a collaboration, and all the collaborations have an innovation_flow. (at least on accepance. the first migration will fail if this is not true in prod).
Innovation flows have a `innovation_flow > profile > tagsets` tagset with `name='flow-state'` that specifies the current selected state of the Innovation Flow.
- Problem: There are innovation-flows (in acc) that don't have a tagset and that gives an error when you query the currentState. Also gives an error when the state is not one of the 4 valid states (which happens a lot). At space level 0, innovation flows were not used at all so we were not facing these errors in the past.
- The first migration, will go through all level 0 spaces, get their innovation_flow, get their tagsets, set the correct `tags` (current state), and the allowedValues and defaultValue in the tagset_template.
- This `tagset_template` has a `tagsetTemplateSetId` which has to be set to `callouts_set.tagsetTemplateSetId` of the space.
- This problem is bad data, AND more importantly the logic was way to convoluted. 

### Problem 2:
Callouts in Spaces are organized in tabs using a `tagset` with `name='callout-group'`. In subspaces they are organized in those tabs/flow-states with a `tagset` with `name='flow-state'`.
Those two tagsets are linked by the `callout > callout_framing > profile > tagsets`. 
So normally callouts have 3 tagsets: `name = 'default'`, `name = 'callout-group'`, `name = 'flow-state'`.
- On Space L0, I have found many cases (with acc data) of callouts that don't have a tagset with `name = 'flow-state'`. Or that have it incorrect.

## Summary by CodeRabbit

- **New Features**
	- Introduced a classification-driven organization for callouts to provide more structured tagging and filtering.
	- Enhanced innovation flow tracking with clear current state information and flow state templates for improved progress visibility.

- **Refactor**
	- Streamlined authorization and access control across communities, collaborations, and spaces.
	- Simplified module dependencies and internal workflows for a more efficient system.

- **Chores**
	- Updated the database schema via migrations to support the new classification and innovation flow features.


src/domain/innovation-hub/innovation.hub.entity.ts
```diff
@@ -19,13 +15,6 @@ export class InnovationHub extends NameableEntity implements IInnovationHub {
   })
   account!: Account;
 
-  @Column('varchar', {
-    length: NAMEID_MAX_LENGTH_SCHEMA,
-    nullable: false,
-    unique: true,
-  })
-  nameID!: string;
```
ccanos, line 27: removed nameID on IH?

src/domain/space/space/space.interface.ts
```diff
@@ -38,10 +38,10 @@ export class ISpace extends IAuthorizable {
   })
   level!: SpaceLevel;
 
-  @Field(() => SpaceType, {
-    nullable: false,
-    description: 'The Type of the Space e.g. space/challenge/opportunity.',
-  })
+  // @Field(() => SpaceType, {
```
ccanos: removed from the client, but I see it's still difficult to fully remove it, should we leave it like this?

src/domain/collaboration/callout/callout.resolver.fields.ts
```diff
@@ -77,6 +78,19 @@ export class CalloutResolverFields {
...
+  @AuthorizationAgentPrivilege(AuthorizationPrivilege.READ)
+  @UseGuards(GraphqlGuard)
+  @ResolveField('classification', () => IClassification, {
+    nullable: true,
+    description: 'The comments for this Callout.',
+  })
+  async classification(
+    @Parent() callout: Callout
+  ): Promise<IClassification | undefined> {
+    // TODO: must be a loader, will be used a lot
```
ccanos, line 90: I've seen a loader created in this review, should this function use it?

Comoque1:
Beta tester doesn't have right to create a VC on own space or using own VC BoK
image

src/domain/innovation-hub/innovation.hub.entity.ts
techsmyth: No, inherits it from the NameableEntity parent - so this was double.

src/domain/space/space/space.interface.ts
techsmyth: Separate issue raised to look at usage of space.type inside server + services.

techsmyth:
@Comoque1 tried to reproduce the issue you report above re beta tester not being able to create - and for me it was working fine - but that is after some fixes that Carlos and myself did this afternoon. Can you please retest with latest?

Comoque1:
> @Comoque1 tried to reproduce the issue you report above re beta tester not being able to create - and for me it was working fine - but that is after some fixes that Carlos and myself did this afternoon. Can you please retest with latest?

@techsmyth the issue was resolved with some of the above commits.

src/domain/collaboration/callouts-set/callouts.set.service.ts
```diff
@@ -432,11 +399,14 @@ export class CalloutsSetService {
       ...keyBy(allCallouts, 'id'),
     };
 
-    const minimumSortOrder = Math.min(
+    const sortOrders = [
       ...sortOrderData.calloutIDs
         .map(calloutId => calloutsByID[calloutId]?.sortOrder)
-        .filter(sortOrder => sortOrder)
-    );
+        .filter(sortOrder => sortOrder !== undefined),
```
valentinyanakiev, line 405: this type of filtering is narrower than `.filter(sortOrder => sortOrder)` --> filtering out all falsy values. The current filter removes only `undefined` values.

then you add     `const minimumSortOrder =
      sortOrders.length > 0 ? Math.min(...sortOrders) : 0;` to partially offset that behavior. 

Overall, I believe this is more inefficient code doing the same thing.

src/domain/community/virtual-contributor/virtual.contributor.resolver.fields.ts
```diff
@@ -101,21 +105,30 @@ export class VirtualContributorResolverFields {
...
-      virtualContributor
+    const knowledgeBase =
+      await this.virtualContributorService.getKnowledgeBaseOrFail(
+        virtualContributor
+      );
+
+    this.authorizationService.grantAccessOrFail(
+      agentInfo,
+      knowledgeBase.authorization,
+      AuthorizationPrivilege.READ_ABOUT,
```
valentinyanakiev, line 125: Wouldn't that mean, in the long run, that we need to have VirtualContributor --> About --> KnowledgeBase, with READ_ABOUT on VC, Then READ everywhere below?

valentinyanakiev (commented):
I have gone through the changes and have the following observations:
- overall, good approach, definitely going in the right direction
- using classification is, in my opinion, excellent, but whether continuing to use tags below calssification in our domain model is something we may want to reconsider (e.g. Category)
- some data loaders need to be reviewed / added
- About on VC remains and open question
- small comment re sorting

valentinyanakiev:
> I have gone through the changes and have the following observations:
> 
> * overall, good approach, definitely going in the right direction
> * using classification is, in my opinion, excellent, but whether continuing to use tags below calssification in our domain model is something we may want to reconsider (e.g. Category)
> * some data loaders need to be reviewed / added
> * About on VC remains and open question
> * small comment re sorting

From ChatGPT:
Tags are indeed often used in software engineering as a flexible, lightweight mechanism for classification. They’re popular in systems like content management, issue tracking, and social media platforms because they allow users to quickly assign, filter, and search for items without imposing rigid structure. However, when it comes to domain model design—especially in systems where business rules and integrity are paramount—there are several alternatives that might be more appropriate:

### Alternatives in Domain Model Design

- **Enumerated Types (Enums):**  
  When classifications are known and limited, enums enforce consistency and restrict values to a predefined set. This approach reduces errors and makes domain constraints explicit.

- **Hierarchical Taxonomies:**  
  Instead of using free-form tags, you can design a structured classification system (often a tree or graph) that represents the relationships between categories. This provides a clearer organizational hierarchy and better supports business logic that relies on these relationships.

- **Explicit Domain Entities:**  
  In domain-driven design (DDD), classifications can be modeled as first-class citizens—separate entities or value objects. For example, rather than assigning a “tag” to a product, you might have a dedicated “Category” or “Status” entity that encapsulates all necessary business rules.

- **Polymorphic Object Hierarchies:**  
  Using inheritance and polymorphism, you can create specialized subclasses for each classification. This method makes the classification an intrinsic part of the object’s identity and behavior, aligning well with object-oriented design principles.

- **Attributes/Annotations:**  
  Some systems leverage metadata through annotations or attributes, which can be validated against domain rules. This approach often provides a middle ground by offering flexibility while still retaining some level of structure.

### Considerations

- **Flexibility vs. Structure:**  
  Tags offer high flexibility but can lead to inconsistencies or ambiguities if not managed carefully. In contrast, alternatives like enums or explicit entities enforce consistency and align better with a rigorous domain model.

- **Business Rules and Integrity:**  
  For core business domains, where classification affects the application’s behavior and integrity, a well-defined structure (e.g., explicit entities or hierarchies) is usually more appropriate than free-form tagging.

- **Evolution of Requirements:**  
  Tags are excellent for exploratory phases or applications with evolving classification needs. However, as requirements solidify, transitioning to a more structured model can improve maintainability and robustness.

In summary, while tags are widely used for their simplicity and adaptability, domain model design often benefits from more structured alternatives when classification plays a critical role in the system’s business logic.

My concern is that we may need stricter business rules and integrity (e.g., the issues with Ideeen, Definieren etc breaking the application because of tags are a nice example of why we may want to reconsider in the future).

valentinyanakiev (approved):

PR #5011, opened 2025-03-11
fix timeline check

valentinyanakiev:


Comoque1:
Tested locally.
The issue with the timeline is resolved.
Now hitting the same error on callout level
image

valentinyanakiev:
the issue is in the client, this should not be needed

PR #5012, opened 2025-03-11
Cascade membership removal 

valentinyanakiev:
- if a user is a MEMBER of Space / Subspace, all related Memberships / Admin / Lead credentials are removed
- This PR does NOT look into deletion of Spaces - that is altogether entirely different problem

### DIFF
--- a/src/domain/access/role-set/role.set.service.ts
+++ b/src/domain/access/role-set/role.set.service.ts
@@ -1041,8 +1041,7 @@ export class RoleSetService {
     roleSet: IRoleSet,
     roleType: RoleName,
     userID: string,
-    validatePolicyLimits = true,
-    fullyRevokeMembership: boolean = false
+    validatePolicyLimits = true
   ): Promise<IUser> {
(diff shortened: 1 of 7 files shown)


src/domain/access/role-set/role.set.service.ts
```diff
@@ -1375,11 +1377,71 @@ export class RoleSetService {
...
+
+  private async getAllSubspaceIds(spaceId: string): Promise<string[]> {
+    const spaceHierarchy =
+      await this.spaceLookupService.getFullSpaceHierarchy(spaceId);
+    const subspaces = spaceHierarchy?.subspaces || [];
+    return this.flattenSubspaces(subspaces);
+  }
+
+  private flattenSubspaces(
+    subspaces: { id: string; subspaces?: any[] }[]
```
ccanos: instead of this `any` we can put another `{ id: string }` here no?

ccanos (approved):

ccanos (approved):

PR #5034, opened 2025-03-19
Fix migration

ccanos:
- Previously the migration was wrongly deleting some tagsets
- In acceptance we have some bad data, but with prod data, these three queries, should return 0 results after the migration:
```
select * from tagset where tagset.name = 'callout-group';
select * from tagset_template where tagset_template.name = 'callout-group';
select * from tagset where profileId is null and classificationId is null and id not in (select tagsetId FROM document);
```

There is also another problem not addressed by this migration, in the VC's knowledgeBases there are some callouts that have classification tagsets, and some others that don't. It doesn't make sense (for now) that they have any tagset in the classification, but this may be addressed in the future with another migration.
```
SELECT callouts_set.*, tagset.* FROM callout 
	JOIN callouts_set ON callouts_set.id = callout.calloutsSetId
    LEFT JOIN tagset ON tagset.classificationId = callout.classificationId
WHERE callouts_set.type = 'knowledge-base';
```

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the tag management update process to enhance data consistency.
  - Replaced removal operations with controlled updates for improved reliability.
  - Introduced a new reference property to better support accurate tagging updates.

### DIFF
--- a/src/migrations/1741340581699-innovation-flow-tagsets.ts
+++ b/src/migrations/1741340581699-innovation-flow-tagsets.ts
@@ -85,15 +85,6 @@ export class InnovationFlowTagsets1741340581699 implements MigrationInterface {
 
     // Now have the data in the new setup, convert the L0 callouts to use the new flow states
     await this.setFlowStateOnLevelZeroCallouts(queryRunner);
-
-    // Finally, delete all tagset templates and tagsets for the groups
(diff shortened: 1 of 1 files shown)


valentinyanakiev (approved):
works for me

PR #5047, opened 2025-03-23
Conversion service refresh: simplified + safer approach

techsmyth:
also cleaned up namings within the file.

Previously, we created a new space and moved over key entities. That made sense when there were three different actual entities for the three levels. 

Now with there just being spaces, it is much safer, faster + simpler to just update the key fields on the existing entity. 

Primary risk area is around rolesets, so please carefully review the resulting contents of the role set entity from the converted space. Especially the parent relationship, credentials etc. 

Testing done:
* created new L1 space, converted to L0
* created new L2 space in another L1 space, and converted it fine to L1.

NEW:
* ability to move a L1 space to be a child (and hence L2) of another L1 space in the same L0 space

Open point: how do deal with URL caching? Just wait for it to expire? Or do we explicitly wipe it?

## Summary by CodeRabbit

- **New Features**
  - Enhanced role management with new options to remove current roles and disassociate parent roles.
  - Introduced upgraded space conversion functionalities that now support transforming spaces between hierarchical levels with improved validations.
  - Streamlined license creation and template management for spaces.
  - Added new input types for space conversion processes.

- **Refactor**
  - Consolidated and renamed conversion processes to improve clarity and organization while updating dependency structures.


src/domain/access/role-set/role.set.service.ts
```diff
@@ -1174,21 +1174,26 @@ export class RoleSetService {
...
-        throw new RoleSetPolicyRoleLimitsException(
-          `Max limit of users reached for role '${roleType}': ${userPolicy.maximum}, cannot assign new user.`,
-          LogContext.COMMUNITY
-        );
+    switch (action) {
+      case RoleSetUpdateType.ASSIGN: {
+        if (userMembersCount === 0) {
+          break;
+        }
+        if (userMembersCount === userPolicy.maximum) {
```
valentinyanakiev, line 1182: what if this is higher?

src/domain/access/role-set/role.set.service.ts
```diff
@@ -1174,21 +1174,26 @@ export class RoleSetService {
...
-    }
-
-    if (action === RoleSetUpdateType.REMOVE) {
-      if (userMembersCount === userPolicy.minimum) {
-        throw new RoleSetPolicyRoleLimitsException(
-          `Min limit of users reached for role '${roleType}': ${userPolicy.minimum}, cannot remove user from role on RoleSet: ${roleSet.id}, type: ${roleSet.type}`,
-          LogContext.COMMUNITY
-        );
+      case RoleSetUpdateType.REMOVE: {
+        if (userMembersCount === userPolicy.minimum) {
```
valentinyanakiev, line 1191: again, I can see a scenario where this is below (e.g. the policy has changed).

src/services/api/conversion/dto/convert.dto.space.l1.to.space.l2.input.ts
```diff
@@ -0,0 +1,19 @@
+import { Field, InputType } from '@nestjs/graphql';
+import { UUID } from '@domain/common/scalars';
+
+@InputType()
+export class ConvertSpaceL1ToSpaceL2Input {
+  @Field(() => UUID, {
+    nullable: false,
+    description:
+      'The Space L1 to be moved to be a child of another Space L. Both the L1 Space and the parent Space must be in the same L0 Space. ',
```
valentinyanakiev, line 9:
```suggestion
      'The Space L1 to be moved to be a child of another Space L1. Both the L1 Space and the parent Space must be in the same L0 Space. ',
```

valentinyanakiev (changes requested):
A couple of minor comments.

hero101:
@techsmyth Are the entitlements considered here?

techsmyth:
@hero101 yes take aim to that into account. When promoting L1 to L0 then the license is created + added. For converting L1 to L2, or L2 to L1 then nothing should be needed. A license reset is needed on the space after the changes.

PR #5055, opened 2025-04-02
fix empty reference names

valentinyanakiev:
- references for calloutFraming profiles with empty name get name 'default'
- we can change the name if needed - it is in a constant in the migration

## Summary by CodeRabbit

- **Bug Fixes**
  - Enforced a minimum of 3 characters for reference names during creation.
  - Automatically updated any previously empty reference names to a default value for consistency.

### DIFF
--- a/src/domain/common/reference/dto/reference.dto.create.ts
+++ b/src/domain/common/reference/dto/reference.dto.create.ts
@@ -4,12 +4,13 @@ import {
   MID_TEXT_LENGTH,
   SMALL_TEXT_LENGTH,
 } from '@src/common/constants';
-import { IsOptional, MaxLength } from 'class-validator';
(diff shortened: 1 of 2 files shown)


techsmyth (dismissed):
Makes sense, and yes there should have been a check on the validation here. 

Only question is why not fix all references, not just those on CalloutFraming?

Note: I tested with local setup, not with corrupted data.

valentinyanakiev:
> Makes sense, and yes there should have been a check on the validation here.
> 
> Only question is why not fix all references, not just those on CalloutFraming?
> 
> Note: I tested with local setup, not with corrupted data.

I tried to narrow down the scope as there was no validation for '' (there was min size on the client yup validation), just for NULL on the server, but you may be right and I found one such entry in prod so I am going to update all.

PR #5060, opened 2025-04-06
PaginatedVCs 

techsmyth:
Looking to make it possible to get also available VCs on a roleSet for inviting.

Added the functions, and also the migration.

To test, simply go to a L1 space and retrieve available VCs where there is one in the parent L0 space:

```
query AvailableVirtualContributorsInSpaceL0($spaceId: UUID!) {
  lookup {
    space(ID: $spaceId) {
      id
      community {
        id
        roleSet {
          id
          availableVirtualContributorsForEntryRole {
            virtualContributors {
              id
              profile {
                id
                displayName
                description
                url
              }
            }
          }
        }
      }
    }
  }
}
```

## Summary by CodeRabbit

- **New Features**
  - Added support for paginated lists of virtual contributors, including a new paginated GraphQL type and resolver field for querying available virtual contributors for entry roles.
  - Introduced a new unique, auto-incrementing identifier (rowId) for virtual contributors to enhance pagination and data handling.

- **Bug Fixes**
  - Removed unnecessary parameter from the virtual contributor profile resolver to streamline data fetching.

- **Chores**
  - Database migration to add the new rowId column with a unique index for virtual contributors.


techsmyth:
@hero101 need your help here...

src/migrations/1743831360133-paginatedVCs.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class PaginatedVCs1743831360133 implements MigrationInterface {
+  name = 'PaginatedVCs1743831360133';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE \`virtual_contributor\` ADD \`rowId\` int NOT NULL AUTO_INCREMENT`
```
hero101: interesting....  the error suggests there is already an `auto_increment` column and a second of the kind needs to have an index applied, which I don't believe is true?
Combining both statements will let SQL know that you are planning on applying an index to it. The same goes for the `down` migration.
I also suggest adding `FIRST` to the column so it fits the same table structure as the rest of the paginated tables.
```suggestion
      `ALTER TABLE \`virtual_contributor\` ADD \`rowId\` int NOT NULL AUTO_INCREMENT,
      ADD UNIQUE INDEX `IDX_a643bc875218dd4abbf86bbf7f` (`rowId`) FIRST`
```

techsmyth:
@hero101 this PR is further, but there is an issue with the paginated results not having authorization loaded. It is actually also an issue for the User paginated results; but there we step past it by doing the authorization check on Profile inside the field resolver + not use a guard. Think this needs a generic fix for paginated results.

techsmyth:
@hero101 I updated the migrations order, and also had to fix a bug in a previous template re assuming a template already exists (not true on new db).

techsmyth:
Also put the migrations fix in separately as its own PR

hero101 (approved):

PR #5112, opened 2025-04-28
Single activity type for creating subspaces

techsmyth:
Previously the code had different activities for creation of L1 and L2, called challenge + opportunity created.

Now just a single subspace created activity type. 

This is about removing concept of challenge + opportunity further from the code base, especially in this case regarding the client. 

Remove type field on community result, should not be used anywhere. Client was querying it though.

===

## Summary by CodeRabbit

- **New Features**
  - Introduced a new activity event type: "Subspace Created".

- **Bug Fixes**
  - Updated activity logs and database records to replace "Challenge Created" and "Opportunity Created" events with the new "Subspace Created" event.

- **Refactor**
  - Simplified activity log interfaces and removed outdated event types and related data structures.
  - Enhanced role set event handling with explicit space level validation.
  - Removed the "type" field from role-related data structures for clarity.
  - Removed several deprecated GraphQL fields related to space types and parent identifiers.

- **Chores**
  - Removed obsolete code and data transfer objects associated with deprecated event types.
  - Corrected naming typos in community application form exports.


src/domain/access/role-set/role.set.service.events.ts
```diff
@@ -74,7 +76,9 @@ export class RoleSetEventsService {
           }
         );
         break;
-      default: // Challenge, Opportunity, VIRTUAL_CONTRIBUTOR, BLANK_SLATE...
+      }
```
bobbykolev, line 79: The comment here was saying 'VIRTUAL_CONTRIBUTOR, BLANK_SLATE...', together with Challenge & Opportunity in the default clause. @techsmyth can you double-check we're safe to throw an error in the default from now on?

bobbykolev:
We're still using `export enum SpaceType`. Should we keep them?

bobbykolev (approved):
Tests were fixed with the following PR: https://github.com/alkem-io/test-suites/pull/432

valentinyanakiev (approved):

PR #5136, opened 2025-05-15
Fix application admin Email link to point to settings/community instead of community

bobbykolev:
⚠️ Pointed to `release/19`
Regression around the SpaceAbout and Journey removal.


techsmyth (approved):
This looks fine.

The only concern is that now all space URLs wll start at community for notifications, which I think is ok - but worth keeping in mind that we may want different space admin url landings later.

PR #5149, opened 2025-05-21
Support admins to also act as members when admin flag is set

techsmyth:
also return the support admins as valid credential when the flag for allowing global support to behave as admins for the Member role

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated role logic so that both ADMIN and MEMBER roles are now treated consistently when platform support as admin is enabled in privacy settings. This ensures correct assignment of support credentials for these roles.


techsmyth:
Note: this PR becomes redundant with the bigger PlatformRolesAccess changes - so take it in or not depending on when you can look at the other PR.

PR #5157, opened 2025-05-22
Add BlueSky and remove twitter from users' references

bobbykolev:
As part of the task, we need to:
- Remove not used Twitter references.
- Add BluSky as the default social option instead.

## Summary by CodeRabbit

- **New Features**
	- Added BlueSky as a default social media reference for user and organization profiles.
	- Organization profiles now include default social references for BlueSky, GitHub, and LinkedIn upon creation.

- **Bug Fixes**
	- Removed empty or null Twitter references from user profiles.

- **Chores**
	- Updated naming and references to ensure consistency across user and contributor profile defaults.

### DIFF
--- a/src/migrations/1747925999514-AddBlueSkyRefToUserProfiles.ts
+++ b/src/migrations/1747925999514-AddBlueSkyRefToUserProfiles.ts
@@ -1,9 +1,10 @@
 import { MigrationInterface, QueryRunner } from 'typeorm';
 import { randomUUID } from 'crypto';
 
-export class AddBlueSkyRefToUserProfiles1747925999514 implements MigrationInterface {
-
-    public async up(queryRunner: QueryRunner): Promise<void> {
+export class AddBlueSkyRefToUserProfiles1747925999514
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
     // Define the new reference
     const bskyReference = {
       name: 'bsky',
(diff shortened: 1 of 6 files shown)


src/migrations/1747925999514-AddBlueSkyRefToUserProfiles.ts
```diff
@@ -0,0 +1,70 @@
...
+    // Remove all "twitter" references where the URI is empty
+    await queryRunner.query(`
+      DELETE FROM reference
+      WHERE name = 'twitter' AND (uri IS NULL OR uri = '')
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    // Remove the "bsky" reference from all profiles
+    await queryRunner.query(`
```
techsmyth: We cannot delete users data, they are free to add whatever references they want. It is part of our UX design which socials get features. @bobbykolev @SimoneZaza

techsmyth (changes requested):
See note re not touching existing references.

src/migrations/1747925999514-AddBlueSkyRefToUserProfiles.ts
bobbykolev: I'll remove the down migration. We're not deleting user data. The only thing that we're removing is the explicit Twitter featured social field, but only if it's not filled in. 
For example, your Twitter link Neil, will continue to be presented with an icon on the profile (no change). But it will be extracted as regular reference in your settings.

valentinyanakiev (approved):

PR #5159, opened 2025-05-23
Update org profile to support social references

bobbykolev:


src/domain/community/organization/organization.service.ts
```diff
@@ -616,4 +622,21 @@ export class OrganizationService {
       reservedNameIDs
     );
   }
+
+  private getDefaultContributorProfileReferences(): CreateReferenceInput[] {
+    const references: CreateReferenceInput[] = [];
+    const referenceTemplates = userDefaults.references; // Reuse userDefaults.references
```
ccanos: can we call them contributorDefaults? just a suggestion, it makes no sense for virtualContributors, but sounded weird to import userDefaults here

bobbykolev: it was like that but it makes sense, yes

ccanos (approved):

PR #5162, opened 2025-05-28
Subspace tutorials fixes

ccanos:
- Migration to rewrite flowState and sortOrder on the callouts in the tutorials callouts according to the bootstrap hardcoded values.
- There was a different problem creating subspaces L2:
  - Set the default subspace template in the space settings
  - Create a L2 subspace inside another subspace
  - The platform subspace template was used. 
  - Only L0 spaces have a templatesManager. The subspaces have templatesManagerId = null. It was not getting properly the templatesManager, so it was not using the space template. I've made a while loop climbing up to parent space that may be implemented somewhere else in a better way.
  -  (This may need to be fixed in a different way, maybe we can create a migration to make all subspaces to point to the root templates manager) 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved subspace creation by ensuring the correct templates manager is inherited from the nearest parent space with a templates manager.
  - Enhanced collaboration input building to include classification tagsets for callouts.

- **Chores**
  - Updated tutorial callouts in the database to align their flow state and sort order with the latest configuration.

### DIFF
--- a/src/migrations/1748424718837-tutorialsFlowState.ts
+++ b/src/migrations/1748424718837-tutorialsFlowState.ts
@@ -1,0 +1,108 @@
+import { TagsetReservedName } from '@common/enums/tagset.reserved.name';
+import { bootstrapSpaceTutorialsCallouts } from '@core/bootstrap/platform-template-definitions/space-tutorials/bootstrap.space.tutorials.callouts';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class TutorialsFlowState1748424718837 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // Get the calloutsSetId from the collaboration associated with the platform space tutorials template
+    const [{ calloutsSetId }]: { calloutsSetId: string }[] =
+      await queryRunner.query(
+        `SELECT calloutsSetId FROM collaboration WHERE id = (
+          SELECT collaborationId FROM template WHERE id = (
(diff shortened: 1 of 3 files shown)


src/migrations/1748424718837-tutorialsFlowState.ts
```diff
@@ -0,0 +1,108 @@
+import { TagsetReservedName } from '@common/enums/tagset.reserved.name';
+import { bootstrapSpaceTutorialsCallouts } from '@core/bootstrap/platform-template-definitions/space-tutorials/bootstrap.space.tutorials.callouts';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class TutorialsFlowState1748424718837 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // Get the calloutsSetId from the collaboration associated with the platform space tutorials template
+    const [{ calloutsSetId }]: { calloutsSetId: string }[] =
+      await queryRunner.query(
+        `SELECT calloutsSetId FROM collaboration WHERE id = (
```
valentinyanakiev, line 10: there is some redundancy here, it could be

```sql
            SELECT 
    c.calloutsSetId
FROM
    template_default AS td
        JOIN
    template AS t ON td.templateId = t.id
        JOIN
    collaboration AS c ON t.collaborationId = c.id
WHERE
    td.type = 'platform-space-tutorials';
```

valentinyanakiev (approved):

PR #5171, opened 2025-06-02
Enhances name ID generation for uniqueness

valentinyanakiev:
Improves the algorithm for generating unique name IDs by:

- Adding a 4-digit numeric suffix to ensure uniqueness.
- Limiting the base string length to accommodate the suffix.
- Replacing spaces with hyphens for better formatting.
- Providing a fallback for empty input to ensure a valid ID is always generated.

This ensures that generated IDs are more unique and conform to length constraints.


techsmyth:
@valentinyanakiev  it was an deliberate decision to **remove** the four digits from the generated nameID, simply not friendly for the users and very visible. To be discussed with @SimoneZaza

PR #5187, opened 2025-06-11
Space creation - use about information from the template

bobbykolev:
- [x] Apply the available 'About' information from the template on space creation.
- Additional fields to be obtained on Space Template creation (visuals, guidelines, tags, etc) To be extracted in another sotry.
 

## Summary by CodeRabbit

- **New Features**
  - Improved space creation by automatically merging template "About" information with user-provided details, ensuring key fields are preserved or enhanced based on available data.  
  - Enhanced template reloading to fetch more detailed "About" profile information within space content.

### DIFF
--- a/src/domain/space/space/space.service.ts
+++ b/src/domain/space/space/space.service.ts
@@ -1258,17 +1258,21 @@ export class SpaceService {
     templateSpaceContent: ITemplateContentSpace,
     spaceData: CreateSpaceInput
   ): CreateSpaceAboutInput {
+    const templateAbout = templateSpaceContent.about;
+    if (!templateAbout || !templateAbout.profile) {
+      return spaceData.about;
+    }
+
     return {
-      why: templateSpaceContent.about.why,
-      who: templateSpaceContent.about.who,
+      why: spaceData.about.why || templateAbout.why,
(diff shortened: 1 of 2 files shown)


src/domain/space/space/space.service.ts
```diff
@@ -1245,4 +1253,28 @@ export class SpaceService {
...
+  private mergeTemplateSpaceAbout(
+    templateSpaceContent: ITemplateContentSpace,
+    spaceData: CreateSpaceInput
+  ): CreateSpaceAboutInput {
+    const templateAbout = templateSpaceContent.about;
+    if (!templateAbout || !templateAbout.profile) {
+      return spaceData.about;
+    }
+
+    return {
```
hero101, line 1266: `||` does not allow empty strings. I believe `??` is more appropriate.

hero101 (changes requested):

src/domain/space/space/space.service.ts
bobbykolev: It was left that way intentionally, as if a form is left empty, then it will not pick up the template value.
I think it makes more sense on creation. Then, when you edit `??` makes it possible to keep it empty.

hero101 (approved):

PR #5205, opened 2025-06-21
Allow mentioning Vcs from any level higher in the hierarchy

techsmyth:
Note: encountered a separate bug related to parentCredentials on role definitions on L2. Included in this branch as needed it to test but the PR for that fix should be merged first. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new field to display all virtual contributors associated with a role, including those from parent role sets.

- **Bug Fixes**
  - Corrected how parent credentials are referenced when updating role set relationships.

- **Chores**
  - Added a migration to update and fix parent credential assignments for certain spaces at level 2.


bobbykolev (dismissed):
Looks and works good. Tested locally. 
The migration makes the logic work for L2 spaces, otherwise the VCs were not available.

From the migration audit, I got this warning. I guess we shouldn't hit it but FYI:

> If a matching parent role isn’t found, the migration throws an error and stops. This could leave your database in a partially-migrated state.

I'm not merging as the migration is assigned for review to @valentinyanakiev in another PR

valentinyanakiev (approved):

PR #5243, opened 2025-07-17
Recursive logic for deletion of space tmpl

bobbykolev:
Extending the deletion to include all* relations. @ccanos , please double check that I'm not missing some relation and we're not leaving orphan data.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deletion process for template content spaces to ensure all related subspaces and associated data are properly removed, preventing potential errors and data inconsistencies.

### DIFF
--- a/src/domain/template/template-content-space/template.content.space.service.ts
+++ b/src/domain/template/template-content-space/template.content.space.service.ts
@@ -122,7 +122,11 @@ export class TemplateContentSpaceService {
       }
     );
 
-    if (!templateContentSpace.authorization) {
+    if (
+      !templateContentSpace.authorization ||
+      !templateContentSpace.about ||
+      !templateContentSpace.collaboration
+    ) {
       throw new RelationshipNotFoundException(
         `Unable to load authorization policy to delete TemplateContentSpace: ${templateContentSpace.id} `,
         LogContext.SPACES
--- a/src/domain/template/template-content-space/template.content.space.service.ts
+++ b/src/domain/template/template-content-space/template.content.space.service.ts
@@ -128,7 +128,7 @@ export class TemplateContentSpaceService {
       !templateContentSpace.collaboration
     ) {
       throw new RelationshipNotFoundException(
-        `Unable to load authorization policy to delete TemplateContentSpace: ${templateContentSpace.id} `,
+        `Unable to load entities to delete TemplateContentSpace: ${templateContentSpace.id} `,
         LogContext.SPACES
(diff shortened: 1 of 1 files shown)


src/domain/template/template-content-space/template.content.space.service.ts
```diff
@@ -89,45 +89,82 @@ export class TemplateContentSpaceService {
...
         },
       }
     );
 
-    if (
-      !templateContentSpace.collaboration ||
-      !templateContentSpace.about ||
-      !templateContentSpace.authorization
-    ) {
+    if (!templateContentSpace.authorization) {
```
valentinyanakiev: Why were these checks removed? If the entities are not loaded it is indeed an exception (in the file below you try to delete them).

valentinyanakiev (changes requested):

src/domain/template/template-content-space/template.content.space.service.ts
bobbykolev: If in the future we start copying about and collaboration conditionally, this logic will break. We have checks below, and we try to delete only if the entities are present. Otherwise, we delete what was copied/part of the template.

Let me know if you think we should force that.

valentinyanakiev: Why did you keep the authorization then? What's the rationale there?

ccanos (dismissed):

ccanos:
restored the checks and tested it locally

src/domain/template/template-content-space/template.content.space.service.ts
```diff
@@ -89,45 +89,86 @@ export class TemplateContentSpaceService {
...
     if (
-      !templateContentSpace.collaboration ||
+      !templateContentSpace.authorization ||
       !templateContentSpace.about ||
-      !templateContentSpace.authorization
+      !templateContentSpace.collaboration
     ) {
       throw new RelationshipNotFoundException(
-        `Unable to load entities to delete TemplateContentSpace: ${templateContentSpace.id} `,
+        `Unable to load authorization policy to delete TemplateContentSpace: ${templateContentSpace.id} `,
```
valentinyanakiev:
```suggestion
        `Unable to load entities to delete TemplateContentSpace: ${templateContentSpace.id} `,
```

valentinyanakiev (changes requested):

valentinyanakiev (approved):

PR #5246, opened 2025-07-18
Assign user roles when creating space with subspaces

ccanos:
## Summary by CodeRabbit

* **New Features**
  * Improved role assignment by supporting nested role sets within spaces and their subspaces when creating a space on an account. This ensures user roles are assigned more comprehensively across related spaces.
  * Enhanced space creation to include root spaces along with their subspaces automatically.


src/domain/space/account/account.service.ts
```diff
@@ -153,6 +168,25 @@ export class AccountService {
     });
   }
 
+  private findNestedRoleSets = (
+    space: ISpace,
+    maxDepth: number = 2
+  ): IRoleSet[] => {
+    if (maxDepth < 0) {
```
valentinyanakiev: Rename maxDepth to spaceLevel and invert the logic

valentinyanakiev (changes requested):
Please rename createSpaceL0 as it is misleading now `let space = await this.spaceService.createSpaceL0(spaceData, agentInfo);`

valentinyanakiev (approved):

PR #5262, opened 2025-07-26
Invite to multiple additional roles

techsmyth:
the mutation to invite a user to a role set now takes an array of extra roles, not just potentially one

domain model updated for invitation + platform invitation to store an array of roles that are being invited for.

## Summary by CodeRabbit

* **New Features**
  * Invitations and platform invitations now support assigning multiple extra roles instead of just one.

* **Bug Fixes**
  * Updated all related forms, APIs, and data structures to handle multiple extra roles consistently.

* **Database Migration**
  * Existing invitations and platform invitations are automatically migrated to the new multi-role format.


bobbykolev (approved):
Overall, the changes are looking good. I've pushed a small fix for the migration as the platform's extraRoles were not copied. 
The extraRole information is not used in the notifications, so no need to check that.

PR #5277, opened 2025-07-29
Fix tagset templates on every innovation flow state update

ccanos:
- Moved the `updateInnovationFlowState` mutation to the innovation-flow.service because it does more work on the innovation-flow than on the actual state.
- The mutation still calls the same `update` function in the innovationFlowState service
- The `updateInnovationFlowState` now updates the flowState, if the name of the flowState has been changed then updates the flowState tagset-template and updates the tagsets associated with it moving all the classifications to the new state
- A few tweaks about states not in the correct sort order
- Added the `valueRenamed` param to updateTagsetsSelectedValue to handle cases where a flowState has been renamed so the old value is not valid anymore, but we don't want to move the callouts to the first state, we want to keep them in the renamed state. - This is actually fixing an old issue on the client that on state rename we had to move callouts one by one to the "new" state.
- Removed a few `isTemplate` parameters that were not used anywhere and will not be used because we don't have innovation-flow templates anymore

## Summary by CodeRabbit

* **New Features**
  * Added the ability to update individual innovation flow states directly, including support for renaming state values and synchronizing tagsets accordingly.

* **Bug Fixes**
  * Improved default value selection for innovation flow state tagsets, ensuring it matches the current state or defaults appropriately.

* **Refactor**
  * Unified and simplified sorting of innovation flow states by sort order for consistent display and processing.
  * Streamlined update logic for innovation flow states and tagsets, removing obsolete parameters and code paths.

* **Chores**
  * Removed deprecated mutation resolver for innovation flow states.


src/domain/collaboration/innovation-flow/innovation.flow.service.ts
```diff
@@ -64,15 +64,13 @@ export class InnovationFlowService {
...
 
-    if (!isTemplate) {
-      if (!flowTagsetTemplate) {
-        throw new ValidationException(
-          `Require flowTagsetTemplate on non-template InnovationFlow: ${innovationFlowData}`,
-          LogContext.INNOVATION_FLOW
-        );
-      }
-      innovationFlow.flowStatesTagsetTemplate = flowTagsetTemplate;
+    if (!flowTagsetTemplate) {
```
hero101: if missing it is an exception flow, then it must be required in the args.

hero101 (changes requested):
Except that single thing, I would suggest @valentinyanakiev to look at it because he raised some concerns about the tagsets.

src/domain/collaboration/innovation-flow/innovation.flow.service.ts
ccanos: right, addressed thanks

valentinyanakiev (approved):

PR #5285, opened 2025-07-30
Notification settings; recipients api

techsmyth:
extended authorization to have privilege for determining if a user can receive notifications on an entity

UserSettings now a full entity, with a set of json fields.

replaced preferences with extension to the UserSettings entity.

Migration is added.

New query to allow getting the email + inapp recipients for events.

InApp modules tidied up: proper module for the entity, separated out fully from reading module.

Beta testers now get a privilege a platform level to use InApp Notifications - the notification receivers logic checks that privilege. The previous code to look up beta tester roles is removed. 

InApp components moved to the right location in the code base: cluster, api, platform.

InApp Receiver is hosted by the server ==> it MUST be responsible for the payloads that it receives, the previous logic of having that coming from the notifications-lib is simply not the right pattern architecturally. The server now manages the enum types and also the  payload definitions, so they are also removed from notifications-lib package. 

InApp data is now stored for all notification event types.
InAppReader has dtos + receivers for all notification event types.

NotificationAdapter has split out the Space notification related work.

Filtering now possible on the set of event types to retrieve. 

InApp notifications now contain a payload with additional notification specific data. There are then resolvers for all the different types of payloads. Note: the set of payloads is less than the set of events.

Todo:
- test all the different event types + flows
- round out the resolvers + data loaders for returning the details per in app notification payload type. 

## Summary by CodeRabbit

- **New Features**
  - Persistent per-user settings with granular notification controls (space, organization, platform, user) and full in‑app notification delivery.

- **Refactor**
  - Notifications reorganized: separate space/platform adapters, centralized recipient resolution, in‑app delivery pipeline, and user settings replacing legacy preference sets; settings access now loader-protected.

- **Migrations**
  - DB migrations to create user_settings, migrate legacy preferences, and normalize in‑app notification data.

- **Authorization**
  - New notification-related privileges and USER_SETTINGS policy type.

- **Chores**
  - Updated notifications library dependency.


techsmyth:
@coderabbitai review

bobbykolev (approved):

PR #5304, opened 2025-08-06
Better visibility into the Wingback license process

hero101:
todo: unit tests

- improved the flow around assigning license entitlements, including Wingback
- `ACCOUNT_LICENSE_PLUS` is now granted upon creating a Wingback account


valentinyanakiev:
updated hotfix branch with this code

PR #5357, opened 2025-08-28
Remove demo spaces from search

valeksiev:
adds levelZeroSpace relation to Space entity and corresponding index
fixes *Count functions for search ingest
filters all search ingested contents based on the levelZeroSpace visibility

## Summary by CodeRabbit

- New Features
  - Improved linking between callouts, framings, whiteboards, posts, contributions, collaborations, and spaces.
  - Search now indexes callouts, posts, and whiteboards directly, respecting root space visibility for more accurate results.
- Refactor
  - Search ingestion uses stricter visibility filters for cleaner, relevant indexing.
  - Space visibility updates now apply in a single bulk operation for faster propagation.
  - Minor cleanup in whiteboard error handling.
- Chores
  - Added database migration to establish a self-referential Space foreign key.


src/domain/space/space/space.entity.ts
```diff
@@ -102,6 +102,10 @@ export class Space extends AuthorizableEntity implements ISpace {
   @Column('char', { length: UUID_LENGTH, nullable: true })
   levelZeroSpaceID!: string;
 
+  @ManyToOne(() => Space)
+  @JoinColumn({ name: 'levelZeroSpaceID' })
+  levelZeroSpace?: Space;
```
hero101: This may create some sort of circular dependency, since a root Space level zero Space is itself.
I would not dismiss it lightly...

hero101 (changes requested):
The Space to L0 Space may create a circular dependency - @techsmyth can you give your opinion here.

src/domain/space/space/space.entity.ts
techsmyth: Indeed this one may well cause issues - I explicitly avoided it when doing the original implementation. Please call after the daily to discuss.

valeksiev: @hero101 can you, please take another look?

hero101 (approved):

PR #5359, opened 2025-08-28
Filter out unsupported inAppNotifications

bobbykolev:
Note the target branch - release/30

**The Issue**
We have the filter for supported in-apps passed by the client in the query.
However, on subscription update, the in-apps are not filtered.

**The fix**
Filter out non-supported in-app events on the server.

**The Why**
We have one client. Let's filter all supported types in one place and avoid sending redundant events. 
I'd suggest removing the filter from the client (next release).


techsmyth (approved):
Clean

PR #5416, opened 2025-09-15
Improves pagination for notifications

valentinyanakiev:
Enhances pagination logic for in-app notifications to ensure correct ordering and stability, especially when using cursor-based pagination.

Uses `triggeredAt` and `id` as a composite key for cursor-based pagination, which ensures stable pagination while maintaining the semantic order of notifications.

Preserves existing order by clauses for relay style pagination and only adds rowId if it doesn't exist.

### DIFF
--- a/src/core/pagination/relay.style.pagination.fn.ts
+++ b/src/core/pagination/relay.style.pagination.fn.ts
@@ -84,10 +84,22 @@ export const getRelayStylePaginationResults = async <
 
   const { first, after, last, before } = paginationArgs;
 
-  query.orderBy({
-    [`${query.alias}.${SORTING_COLUMN}`]: sort,
-    ...query.expressionMap.orderBys,
-  });
+  // Preserve existing orderBy clauses and only add rowId if not already present
+  const existingOrderBys = query.expressionMap.orderBys || {};
+  const hasExistingOrder = Object.keys(existingOrderBys).length > 0;
+
+  if (hasExistingOrder) {
+    // Add rowId as secondary sort for pagination efficiency, preserving existing primary sort
+    const rowIdColumn = `${query.alias}.${SORTING_COLUMN}`;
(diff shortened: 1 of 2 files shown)


bobbykolev:
Not sure if we need that complexity. CreatedAt should be very close to the row added to the table. 
Leaving Svetlio to decide whether this change is good to go, considering it affects the core paginating fn.

src/core/pagination/relay.style.pagination.fn.ts
```diff
@@ -84,10 +84,22 @@ export const getRelayStylePaginationResults = async <
 
   const { first, after, last, before } = paginationArgs;
 
-  query.orderBy({
-    [`${query.alias}.${SORTING_COLUMN}`]: sort,
-    ...query.expressionMap.orderBys,
-  });
+  // Preserve existing orderBy clauses and only add rowId if not already present
```
hero101, line 87: `rowId` is always present.
What do we win with this new code?

src/platform/in-app-notification/in.app.notification.service.ts
```diff
@@ -107,9 +107,85 @@ export class InAppNotificationService {
...
+    if (paginationArgs.after || paginationArgs.before) {
+      // If using cursor pagination, add secondary ordering by id for stability
+      queryBuilder.addOrderBy('notification.id', 'DESC');
+      return await this.getPaginatedNotificationsWithTriggeredAtCursor(
+        queryBuilder,
+        paginationArgs
+      );
+    } else {
+      // For simple pagination (first page), we can use the standard approach
+      queryBuilder.addOrderBy('notification.rowId', 'DESC');
```
hero101: the sort order is already defined by `getRelayStylePaginationResults`

hero101 (changes requested):
I have tested it and it works as expected.
I am unsure about of the changes in the core logic.

PR #5420, opened 2025-09-15
Zero downtime ingestion

hero101:
This PR addresses some issues related to the search data ingestion.
Specifically addresses the downtime issues related to partial results or errors while the process is running.

It uses variable index names and aliases to manage which index the usage should be routed to.

The implementation is as follows
```
Create new indices with new names
Ingest into new indices
Do the aliases exist?
 - no - fresh new setup -> assign alias to new indices
 - yes - we have old indices -> remove old aliases; assign alias to new indices
Did we have aliases?
 - no -> do nothing
 - yes -> delete old index associated with the old alias
```

## Summary by CodeRabbit

- New Features
  - Reindexing starts asynchronously and returns a task ID immediately; tasks track progress, errors, and can be marked errored.
  - Full "ingest from scratch" flow: suffixed deployments, alias rotation, and targeted old-index cleanup.

- Refactor
  - Search ingestion modernized to non-blocking, suffix-based deployments with improved per-step status updates and alias management.
  - Reduced batch ingestion delay for faster, consistent pacing.

- Tests
  - Added integration tests covering error, success, and old-index deletion paths for the scratch-ingest workflow.

- Chores
  - Removed a deprecated Elasticsearch policy setting from configuration; TLS settings unchanged.


src/services/api/search/ingest/search.ingest.service.ts
```diff
@@ -122,7 +125,149 @@ export class SearchIngestService {
...
+      }
+
+      this.logger.verbose?.(
+        `Assigning alias '${alias}' to point to index '${index}'`,
+        LogContext.SEARCH_INGEST
+      );
+
+      await this.elasticClient?.indices.updateAliases({ actions });
+    }
+  }
```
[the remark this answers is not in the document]
hero101: I have opted out because it seemed to say "too many actions" but now it seems ok - comment accepted

src/services/api/search/ingest/search.ingest.service.ts
```diff
@@ -346,7 +489,7 @@ export class SearchIngestService {
 
       start += batchSize;
       // delay between batches
-      await setTimeout(1000, null);
+      // await setTimeout(1000, null);
```
hero101:
```suggestion
      await setTimeout(500, null);
```

src/services/api/search/ingest/search.ingest.service.ts
```diff
@@ -122,7 +126,147 @@ export class SearchIngestService {
...
+   * create new indices with new names
+   * ingest into new indices
+   * does the aliases exist?
+   * - no - fresh new setup -> assign alias to new indices
+   * - yes - we have old indices -> remove old aliases; assign alias to new indices
+   * did we have aliases?
+   *  - no -> do nothing
+   *  - yes -> delete index
+   */
+  public async ingestFromScratch(task: Task) {
```
valentinyanakiev, line 138: Can this method be split? I am on 32.5" screen and 4k res and not being able to fit a method in one screen is a smell...

valentinyanakiev (changes requested):
- some (AI-generated) unit tests may make sense in case you decide to change this logic in the future
- as a rule of thumb, shorter methods (100 LOC is long for a single method).

valentinyanakiev (approved):

PR #5423, opened 2025-09-16
Fixes VC Visibility Auth

valentinyanakiev:
- VC visibility change didn't reset the auth
- VS visibility change had 2 positions - HIDDEN / NOT HIDDEN
   - split that into 3
   - now there is an ACCOUNT one that has credentials just for the ACCOUNT_ADMIN

## Summary by CodeRabbit

- New Features
  - After updating a virtual contributor, access policies are automatically recalculated and saved to reflect any visibility changes.
  - Introduces clearer visibility behavior with three modes:
    - Public: globally accessible.
    - Account: restricted to the related account.
    - Hidden: no global access granted.
  - Ensures the latest state is returned after updates for consistent authorization and search visibility.


techsmyth:
@valentinyanakiev understood re account split out, but afraid that I do not think the current way will work. This logic means that only account *admins* will be able to see it, not all users in spaces in the account...to be talked through.

bobbykolev (approved):

PR #5425, opened 2025-09-17
Ability to disable user geo location

hero101:
## Summary by CodeRabbit

- **New Features**
  - Added a configuration toggle to enable/disable the Geo integration (default: off; controllable via environment).
  - Exposed an enabled flag in the public configuration so clients can detect Geo availability.

- **Bug Fixes**
  - When Geo is disabled, requests now return a clear service-unavailable error instead of proceeding.
  - Improved error propagation for Geo requests so failures reach callers reliably.

- **Tests**
  - Updated Geo tests to reflect the new enabled configuration flag.

### DIFF
--- a/src/services/external/geo-location/geo.location.service.ts
+++ b/src/services/external/geo-location/geo.location.service.ts
@@ -106,7 +106,7 @@ export class GeoLocationService {
   }
 
   public getCacheMetadata() {
-    if (this.enabled) {
+    if (!this.enabled) {
       throw new GeoServiceNotAvailableException(
         'Geo location service is disabled',
         LogContext.GEO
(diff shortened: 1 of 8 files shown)


src/services/external/geo-location/geo.location.service.ts
```diff
@@ -38,9 +39,16 @@ export class GeoLocationService {
     this.allowedCallsToService = config.allowed_calls_to_service;
     this.allowedCallsToServiceWindow = config.allowed_calls_to_service_window;
     this.cacheEntryTtl = config.cache_entry_ttl;
+    this.disabled = !config.enabled;
```
valeksiev: I wouldn't invert the flag here. `if(!this.enabled)` looks alright to me in the return early statements below.

valeksiev (dismissed):

src/services/external/geo-location/geo.location.service.ts
```diff
@@ -98,6 +106,13 @@ export class GeoLocationService {
   }
 
   public getCacheMetadata() {
+    if (this.enabled) {
```
valeksiev:
```suggestion
    if (!this.enabled) {
```

valeksiev (approved):

PR #5432, opened 2025-09-18
Proper cache invalidation around applications and invitations

hero101:
Membership status cache invalidated on:
- new invitation
- ~~archived invitation~~
- accepted invitation
- new application
- rejected application
- archived application

Suggestions for later:
- the client right now depends on an open application to determine the membership status and to display the membership button properly. **It would be best if it depends on the membership status only**
- After an application is sent, the client to refetch the membership status
- After an application is rejected there is not indication to the user that it was, except the button lets them apply again - if attempted an error is throw. **I would suggest introducing a new membership status for rejected applications.**

## Summary by CodeRabbit

- New Features
  - None.
- Bug Fixes
  - Membership status and open application/invitation indicators update promptly after creating or deleting applications and invitations, reducing stale states.
- Performance & Reliability
  - Improved cache invalidation for membership and application states for faster, more consistent UI updates.
  - Enhanced logging when participant data is missing without interrupting user actions.
  - Deletion flows now proceed without waiting for background authorization, reducing perceived delay (may alter asynchronous error timing).
- Tests/Docs/Chores
  - None.

### DIFF
--- a/src/domain/access/invitation/invitation.resolver.mutations.ts
+++ b/src/domain/access/invitation/invitation.resolver.mutations.ts
@@ -26,12 +26,12 @@ export class InvitationResolverMutations {
     const invitation = await this.invitationService.getInvitationOrFail(
       deleteData.ID
     );
-    await this.authorizationService.grantAccessOrFail(
+    this.authorizationService.grantAccessOrFail(
       agentInfo,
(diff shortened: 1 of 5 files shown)


bobbykolev (approved):
Looking good on my end.

PR #5445, opened 2025-09-29
Callouts contributions new visualization

ccanos:
## Summary by CodeRabbit

* **New Features**
  * Added MEMO contribution type, memo field on create input, and canonical list of contribution types.
  * Contributions can be filtered by IDs and types via a new filter input; optional limit/shuffle.
  * New contributionsCount field returning per-type counts.
  * Lookup API can resolve a single contribution by ID.

* **Refactor**
  * Centralized contribution-type validation; getContributions accepts types parameter.

* **Chores**
  * Migrations to add/type column, clean data, and backfill banner visuals.
* **Style**
  * Deletion now uses a structured input object.


ccanos:
@coderabbitai full review

ccanos:
@coderabbitai full review

ccanos:
@coderabbitai full review

src/domain/collaboration/callout/dto/callout.contributions.count.dto.ts
```diff
@@ -0,0 +1,29 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+
+@ObjectType()
+export class CalloutContributionsCountOutput {
+  @Field({
+    nullable: false,
+    description: 'The number of contributions of type Post in this callout',
+  })
+  post!: number;
```
bobbykolev, line 9: Why are all of these mandatory? There is only one count based on the type, isn't it?

src/migrations/1758116200183-CleanUpContributions.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+    await queryRunner.query(
+      `DELETE FROM \`callout_contribution\` WHERE
+        \`whiteboardId\` IS NULL AND
+        \`postId\` IS NULL AND
+        \`linkId\` IS NULL AND
+        \`memoId\` IS NULL;`
+    );
+
+    const problems = [
```
bobbykolev, line 29: Do we need this check? You already updated all non-null entries with types.

src/domain/collaboration/callout/dto/callout.contributions.count.dto.ts
ccanos: It's just the count is never nullable, the client can query it or not but if queried it is never going to be `null`.
Also, now that I think about it, we may want to implement the query better, because in `getContributionsCount` in `callout.service.ts`, the mysql query is always counting all contributions types no matter which ones has the client queried, but I think it's simpler and fast enough implemented this way. 
Does this answer your question?

src/migrations/1758116200183-CleanUpContributions.ts
ccanos: yes, I preferred to have a check there, it's a sanity check of the database,
"post contributions that also have whiteboardId", "whiteboard contributions that also have postId", weird things that shouldn't be there, I haven't found any issue in acc db, but I'd like @valentinyanakiev  to run this migration on prod data

bobbykolev (approved):

PR #5455, opened 2025-10-03
Delete default CG

bobbykolev:
As part of the SpaceTemplates, we've set some default titles and descriptions to the CG templates. 
This results in unwanted behaviour in multiple places.

The fix is to update all default text to empty strings.

The migration covers all CG profiles, both for spaces and templates.
The description has a couple of variations because the MD string has a new line at the end (sometimes).

## Summary by CodeRabbit

* **Bug Fixes**
  * Cleans up profiles where community guidelines fields were auto-filled with default copy, preventing duplicated or unintended text from appearing.
* **Chores**
  * Runs a one-time data migration to normalize community guidelines fields by clearing out default placeholder content.

### DIFF
--- a/src/migrations/1759497165260-removeDefaultCommunityGuidelinesCopy.ts
+++ b/src/migrations/1759497165260-removeDefaultCommunityGuidelinesCopy.ts
@@ -1,0 +1,38 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+const TARGET_DISPLAY_NAME = 'Professional Networking Community Name';
+const TARGET_DESCRIPTION = 'Please fill out the community guidelines here';
+// markdown field adds '\n' at the end of the strings
+const TARGET_DESCRIPTION_WITH_NEWLINE =
+  'Please fill out the community guidelines here\n';
+const TARGET_DESCRIPTION_WITH_CRLF =
+  'Please fill out the community guidelines here\r\n';
+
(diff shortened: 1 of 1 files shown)


src/migrations/1759497165260-removeDefaultCommunityGuidelinesCopy.ts
```diff
@@ -0,0 +1,38 @@
...
+        TARGET_DESCRIPTION,
+        TARGET_DESCRIPTION_WITH_NEWLINE,
+        TARGET_DESCRIPTION_WITH_CRLF,
+        TARGET_DISPLAY_NAME,
+        TARGET_DESCRIPTION,
+        TARGET_DESCRIPTION_WITH_NEWLINE,
+        TARGET_DESCRIPTION_WITH_CRLF,
+      ]
+    );
+  }
```
hero101, line 35: is the `CASE` needed when you have the `WHERE` ?

hero101 (changes requested):

src/migrations/1759497165260-removeDefaultCommunityGuidelinesCopy.ts
bobbykolev: From what I remember, it was for the case where we're updating only one of the fields. So we don't want to override the description if changed by the user even if the title is matching. I guess it could be improved but I tested this one and it works with most combinations.

hero101 (approved):

PR #5482, opened 2025-10-11
Direct messages2

techsmyth:
Uses cleaned up api for working with matrix adapter

Updated api around messaging to users
- rename messageUser with sendMessageToUserDirect. Likely this can be merged in with sendMessageToUsers - needs discussion
- remove field communityRooms on a user, not needed

Adds configuration to enable direct messages

Note: requires the directMessages branch on the matrix-adapter + client-web

To test: 
* set the flag to state that sending a message to a user also sends a direct message
```COMMUNICATIONS_DIRECT_MESSAGE_ROOMS_ENABLED=true ```
* send a message to another user using the client web
* using graphql query the direct message rooms for the user using

```
query me {
  me {
    user {
      id
      profile {
        displayName
      }
      directRooms {
        id
        messages {
          message
        }
      }
    }
  }
}
```

ALso useful is to connect in to the synapse admin service using the following details:
username: @matrixadmin44=alkem.io:alkemio.matrix.host
password: change_me_now
homeserver url: http://localhost:8008

## Summary by CodeRabbit

* **New Features**
  * Direct messaging: private one-to-one messages supported when enabled; single-recipient sends route as direct messages.

* **Configuration**
  * New communications setting to enable/disable direct message rooms.

* **API**
  * Messaging mutation updated to accept multiple recipients and to use direct-message flow for single recipients; direct-room retrieval is now exposed.

* **Chores**
  * Package and adapter image versions bumped.


valentinyanakiev (changes requested):
overall, to test this, the client needs to work (codegen fails at the moment and direct messaging fails because of type changes), the matrix-adapter comments should be resolved so we merge it (otherwise the whole messaging will break)

valentinyanakiev:
> overall, to test this, the client needs to work (codegen fails at the moment and direct messaging fails because of type changes), the matrix-adapter comments should be resolved so we merge it (otherwise the whole messaging will break)

As a positive, sending / receiving old messaging is not broken for me, but there is a regression with direct messaging

valentinyanakiev:
BREAKING-APPROVED

valentinyanakiev (dismissed):
BREAKING-APPROVED

PR #5510, opened 2025-10-22
Avatar visual on innovation pack profile

bobbykolev:
Relates to https://github.com/alkem-io/client-web/issues/7286

- [x] create avatar visual on innovation pack creation;
- [x] migration to add an avatar visual to the profiles of all existing innovation packs; ⚠️

## Summary by CodeRabbit

* **New Features**
  * Innovation pack profiles now include avatar visuals alongside other visual elements.
  * Existing innovation pack profiles are automatically updated to add avatar visuals where missing, ensuring profiles consistently display an avatar.


src/migrations/1761125765749-InnovationPacksAvatar.ts
```diff
@@ -0,0 +1,88 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { randomUUID } from 'crypto';
+import { DEFAULT_VISUAL_CONSTRAINTS } from '@domain/common/visual/visual.constraints';
```
techsmyth: There must not be any dependency on code from outside the migrations area, this brings a dependency that we need to avoid. So please duplicate what you need.

techsmyth (dismissed):
In general looks fine, main comment is that the migration is using code from outside the migrations area which must not be done.

src/migrations/1761125765749-InnovationPacksAvatar.ts
hero101: I agree

src/migrations/1761125765749-InnovationPacksAvatar.ts
```diff
@@ -0,0 +1,88 @@
...
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    // Remove avatar visuals added by this migration
+    // could potentially delete avatar visuals added by users later!
+    await queryRunner.query(`
+      DELETE \`visual\` FROM \`visual\`
+      JOIN \`profile\` ON \`visual\`.\`profileId\` = \`profile\`.\`id\`
+      JOIN \`innovation_pack\` ON \`innovation_pack\`.\`profileId\` = \`profile\`.\`id\`
+      WHERE \`visual\`.\`name\` = 'avatar'
+    `);
+  }
```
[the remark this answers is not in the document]
hero101: @coderabbit Deleting the files would be annoying for the end users, in case of a revert and patch, they need to reupload - the changes of that happening are slim though.

hero101 (changes requested):

bobbykolev:
Done, thanks @hero101 & @techsmyth . I also commented the down part just in case, but it's there if someone needs to revert locally.

Note that we have a similar migration that has the imports. I'm not sure how critical that is:
https://github.com/alkem-io/server/blob/develop/src/migrations/1759313997577-addBannerToWhiteboardContributions.ts

hero101:
> Done, thanks @hero101 & @techsmyth . I also commented the down part just in case, but it's there if someone needs to revert locally.
> 
> Note that we have a similar migration that has the imports. I'm not sure how critical that is: https://github.com/alkem-io/server/blob/develop/src/migrations/1759313997577-addBannerToWhiteboardContributions.ts

In general migrations have to reflect the current point in time when they were written and to overwrite the value in later migrations. That is the only way to guarantee compatibility.

hero101 (approved):

PR #5535, opened 2025-10-28
Fix PROD error when accessing space storage settings (w tmp)

bobbykolev:
Fix the error on storage settings page related to CG and community without space parent (when you have assets for a space template which has ContentSpace instead).

Please, double check the fix. As part of an Error squashing session I decided to return empty string. 
Ideally, we should probably support previewing the assets under the parent contentSpace, but this shouldn't be part of a quick fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced system resilience when template and community guidelines resources are unavailable, preventing errors and ensuring graceful handling of missing associations.


hero101 (commented):
I am unsure how this new behaviour of always returning a string is fixing the issue.

It does not produce an error any more, but it looks like it will conceal more issues.
Maybe I am missing context.

hero101 (changes requested):
I am unsure how this new behaviour of always returning a string is fixing the issue.

It does not produce an error any more, but it looks like it will conceal more issues.
Maybe I am missing context.

bobbykolev:
It's a quick solution of the problem for the Error squashing session. You can see the details in the description. 
If you think it's better to discard this solution and keep the existing logic until we have the time to properly resolve this one, feel free to close the PR. @hero101

hero101 (approved):

PR #5569, opened 2025-11-05
lookup of space template url to resolve the hierarchy of template content spaces

techsmyth:
Logic had a clear flaw: namely only looking up template with single level of templateContentSpace. 

Not been able to test as was struggling to reproduce locally, but if reproducible with e.g. acc db then should be quick to verify that this fixes the issue. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template resolution to correctly traverse nested content space hierarchies and locate templates from root spaces.
  * Updated error messaging to clarify when a space template cannot be resolved; behavior returns null when no template is found.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


bobbykolev (approved):
The change makes sense. I couldn't verify that it resolves the issue, as it's no longer reproducible on/with ACC DB.

PR #5573, opened 2025-11-06
Space: Whiteboards new privileges for creators/admins

polibb:
This pull request adds comprehensive documentation for the "Guest Contributions Privilege Management" feature, including a specification quality checklist and a detailed data model. The main focus is on 
- extending the authorization system to support a new `PUBLIC_SHARE` privilege, 
  - which is conditionally granted to space admins and whiteboard owners based on space settings. 
- No database schema changes are introduced;
- performance through the existing cascade infrastructure.

**Data Model and Authorization Design:**

* `AuthorizationPrivilege` enum is extended with a new `PUBLIC_SHARE` value (TypeScript only, no DB migration needed).
* Described the authorization cascade pattern
* atomicity and consistency during authorization resets.

*Data Model and Authorization Design:*
- extension of the `AuthorizationPrivilege` enum to include `PUBLIC_SHARE`, enabling user control over guest access to whiteboards. This privilege is dynamically granted to space admins and whiteboard owners when `allowGuestContributions` is enabled in space settings.
- On admin roles change - also include `PUBLIC_SHARE` privilege per whiteboard in space if setting is True


hero101:
Converting to draft after our discussion with @polibb

src/domain/common/whiteboard/whiteboard.service.authorization.ts
```diff
@@ -156,6 +155,29 @@ export class WhiteboardAuthorizationService {
       }
     }
 
+    if (spaceSettings?.collaboration?.allowGuestContributions) {
```
techsmyth: The PUBLIC_SHARE privilege should be assigned to all admins, also those from parent spaces. There is a method on RoleSet that gives this. Plus there should also be assigned to any platform roles that have admin access - see the usage of PlatformRolesAccess.

polibb: Updated, admins of parents can also open whiteboards to the public; gave platform admins the PUBLIC_SHARE privilege as well (change of scope); and added new privilege to global support team if the setting "allow global support to act as admins" is also enabled (forgotten use case).

For the forgotten use case, refactoring task https://github.com/alkem-io/server/issues/5580 so it doesn't happen again. 

Thank you for pinpointing the places to look into!

src/domain/access/role-set/role.set.module.ts
```diff
@@ -49,6 +50,7 @@ import { RoleSetResolverMutationsMembership } from './role.set.resolver.mutation
     UserModule,
     OrganizationLookupModule,
     SpaceLookupModule,
+    forwardRef(() => SpaceModule),
```
hero101: i don't believe you need this any more

polibb: I tried without it, but that's when the whole server stopped working until I got the just-in fix from `develop`.
Now I tested without it again, and it's alright indeed. Removed

src/domain/collaboration/callout/callout.service.authorization.ts
```diff
@@ -43,7 +44,8 @@ export class CalloutAuthorizationService {
     calloutID: string,
     parentAuthorization: IAuthorizationPolicy | undefined,
     platformRolesAccess: IPlatformRolesAccess,
-    roleSet?: IRoleSet
+    roleSet?: IRoleSet,
+    spaceSettings?: ISpaceSettings
```
hero101: why are they optional?

src/domain/common/whiteboard/whiteboard.service.authorization.ts
```diff
@@ -110,6 +134,71 @@ export class WhiteboardAuthorizationService {
...
+            context: LogContext.COLLABORATION,
+          }
+        );
+      } else {
+        this.logger.verbose?.(
+          `No admin credential holders resolved for whiteboard ${whiteboard.id}; skipping PUBLIC_SHARE admin rule`,
+          LogContext.COLLABORATION
+        );
+      }
+    } else if (spaceSettings?.collaboration !== undefined) {
```
hero101: i am unsure this check makes sense

src/domain/common/whiteboard/whiteboard.service.authorization.ts
```diff
@@ -110,6 +134,71 @@ export class WhiteboardAuthorizationService {
           CREDENTIAL_RULE_WHITEBOARD_CREATED_BY
         );
       newRules.push(manageWhiteboardCreatedByPolicy);
+
+      // T007: Add PUBLIC_SHARE credential rule for whiteboard owner when guest contributions enabled
+      if (spaceSettings?.collaboration?.allowGuestContributions) {
```
hero101: extract to a function specific for appending via settings.
You can exit early if the setttings aren't defined. I suppose the type suggest either all the fields of the settings are provided or it's `undefined` - that would make some of the checks down obsolete.

src/migrations/1760357366405-NotificationEntityTracking.ts
hero101, line 1: revert this

hero101 (changes requested):
Does it work now with global support?

src/migrations/1760357366405-NotificationEntityTracking.ts
polibb: Idk, my IDE shows me the same thing on `develop` and subsequently `feature-whiteboards-public`

Image

Image

hero101 (approved):

PR #5698, opened 2025-12-04
File service integration to use remote authz evaluation

hero101:
The changes shift the authz evaluation from the local (server) implementation to the remote implementation.
It is still not the ideal situation but is a step in the right direction.


techsmyth (commented):
I would have assumed that the call to the remote authz service took place from within the file service? Now it is calling the server which in turn calls the authz - or am I missing something?

hero101:
> I would have assumed that the call to the remote authz service took place from within the file service? Now it is calling the server which in turn calls the authz - or am I missing something?

@techsmyth - Yes, it's intended.
Since neither of our services have access to the database, or are able to build the agent / auth information, this service (as any in the future) will have to go through the following flow:
- service A receives a request that needs to be authorized. The user presents a cookie/auth header/api token to authenticate.
- Hop 1: A calls the collaboration server (CS) to check WHO is this user, build the `AgentInfo`
- Hop 2: CS sends back to A the agent, and any other information about the original request
- Hop 3: A calls the remote authz with authorization info (agent, policy, required privilege)
- Hop 4: remote authz sends back `allowed`

That is a lot of hops, and to two different services, only to serve 1 request...

---

With the suggest approach in this PR, the total hops are 3, hitting again 3 services, but in one straight forward flow.
It is not ideal and should not remain like that, but is a step in the right direction.

As a future step, I can see the agent information logic and the authz to be in the same domain / service and each service to have individual access to the database, following the 12 factor app best practices.

techsmyth:
Thanks for clarifying @hero101 - this needs a discussion - once the work from @antst has merged in we can talk about it.

PR #5734, opened 2026-01-09
Converts varchar(36) to UUID

valentinyanakiev:
Has potential to break things, as there is business logic relying on "" in some places, and that will need adjusting if we migrate, hence the split.


techsmyth (commented):
I think spaceID can be removed fully on communication, think it is legacy.

Also re resourceID being UUID - at the moment yet, but it is not a strict requirement - the credential could be used potentially to point to another resource. So if it helps to make UUID then ok for now, but with a comment that it may be any resource ID later...

antst:
> I think spaceID can be removed fully on communication, think it is legacy.
> 
> Also re resourceID being UUID - at the moment yet, but it is not a strict requirement - the credential could be used potentially to point to another resource. So if it helps to make UUID then ok for now, but with a comment that it may be any resource ID later...

We have empty resource only for global admin (or there is more?). But global admin is an admin of platform. So, this is what has to be in resourceId for GA, id of platform. But, again, we have to remove all this "*-admin" roles, and have just "admin" role. And scope is  defined by resource on which credential is. This also should reduce a bit amount of code in the system and make handling a bit more generic.

techsmyth:
@antst all global roles have empty resource ID

antst:
> @antst all global roles have empty resource ID

But all of them has to have platformId as resource :) This is their scope - platform. But as this will come as part of authorization works later, I suggest that for now we just make it nullable, and use null  for global roles. Should be easy to adopt in the code.

techsmyth:
Agreed - we did not have platform entity when adopted the convention of empty resourceID for global roles...

valentinyanakiev:
closing this - will potentially re-do after drizzle after other redesigns are finalized

PR #5757, opened 2026-01-15
Enables subspace ordering within a space

valentinyanakiev:

This pull request introduces a `sortOrder` field to the `Space` entity, allowing subspaces to be ordered within their parent space. It also adds an API endpoint to update the sort order of multiple subspaces simultaneously. This feature enables users to define the order in which subspaces are displayed, enhancing the user experience when navigating complex space hierarchies.

### Additional context

This change introduces a new `UpdateSubspacesSortOrderInput` and the corresponding resolver and service methods.

## Summary by CodeRabbit

* **New Features**
  * Users can reorder subspaces within a parent space; subspaces are displayed by explicit sort order, then name.
  * Spaces now expose a sort-order attribute; new subspaces receive an initial sort position on creation.

* **Chores**
  * Database migration added to persist and initialize sort order for existing spaces.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/domain/space/space/space.service.ts
+++ b/src/domain/space/space/space.service.ts
@@ -974,7 +974,12 @@ export class SpaceService {
     );
 
     // Sort the subspaces based on sortOrder, with displayName as tiebreaker
-    const sortedSubspaces = limitAndShuffled.sort((a, b) => {
+    // Skip sorting when shuffle is requested to preserve randomization
+    if (args?.shuffle) {
+      return limitAndShuffled;
+    }
+
+    return limitAndShuffled.sort((a, b) => {
       if (a.sortOrder !== b.sortOrder) {
         return a.sortOrder - b.sortOrder;
       }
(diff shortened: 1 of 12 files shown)


src/domain/space/space/space.service.ts
```diff
@@ -970,16 +973,71 @@ export class SpaceService {
...
+      subspacesInOrder.push(subspace);
+      const newSortOrder = minimumSortOrder + index;
+      if (subspace.sortOrder !== newSortOrder) {
+        subspace.sortOrder = newSortOrder;
+        modifiedSubspaces.push(subspace);
+      }
+      index++;
+    }
+
+    await Promise.all(
```
hero101: `Repository.save` supports saving multiple at once.

hero101 (changes requested):
Code look ok.
Be aware that the approach is suboptimal, where you can have subspace with the same sort order - the good thing is the tie will be broken by the displayName.

PR #5815, opened 2026-01-30
Lowercase email and track email duplicates

bobbykolev:
With this PR we aim to:

- normalize user creation (lowercase email);
- normalize the user email on user profile update;
- normalize users' emails in queries;
- migration to log any duplicated user entries (same email different case);


hero101 (changes requested):
Code looks OK.
I am blocking this because I do not see the purpose of having a migration, that does not migrate, to look into the data and console log - this can be done by us either through code or sql query.

hero101 (approved):

PR #5834, opened 2026-02-11
Image compression support + heic conversion

bobbykolev:
## Implementation

- [x] **Conversion** - Based on the specs we now have support of heic/heif images.
- [x] **Compression** - All non-svg, non-png, non-gif images are _resized_ to max 4096px (longest side) and converted to jpeg 82% quality.

_Here are some examples:_
Heic to jpeg: from ~2mb to <1mb and visible on all OS/Browser
Screenshot 2026-02-12 at 18 35 42

jpg to jpeg: from >10mb to <2mb (+ resized to 4096px)
Screenshot 2026-02-12 at 16 55 07

___
## SPEC review (completed)
Please, review the specification & plan.

- First commit: the initial plan was to use the most popular library - **sharp** - that can help with resizing and optimizations in the future.However, that one doesn't support heic (our main issue).
- Second commit: So the plan switched to using lib only for the heic to jpeg conversion as a fix for the main problem.
- Third commit: After a sync with @hero101 and @ccanos , we decided to go all at once, combining **heic support** (conversion) together with **compression** and **resizing**.

With these changes will potentially cover the following tickets:
https://github.com/alkem-io/client-web/issues/9264
https://github.com/alkem-io/client-web/issues/9292
https://github.com/alkem-io/client-web/issues/9283

Let me know if you have any questions or concerns with the approach.

#### The heic issue (other platforms):

> - Slack, Telegram, Signal, Twitter/X, Instagram, Reddit, Gmail — all handle HEIC and have supported it for years, most via server-side auto-conversion
> - Discord — notably the worst offender. They still don't properly support HEIC; images get uploaded as unknown files and don't display. Users have been begging for this for years. [Discord](https://support.discord.com/hc/en-us/community/posts/22721424654231-Support-HEIC-images-HEVC-videos) Don't be Discord.
> - Confluence/Atlassian — still has issues with HEIC, especially from Live Photos

## Summary by CodeRabbit

* **New Features**
  * HEIC/HEIF uploads are now accepted, auto-converted to JPEG and then optimized (resize/compress) before storage.

* **Tests**
  * Added comprehensive unit tests covering conversion and compression flows and edge cases.

* **Documentation**
  * New spec, plan, quickstart, research and task docs detailing the conversion/compression feature and rollout.

* **Chores**
  * Updated dependencies and added migrations to recognize HEIC/HEIF MIME types.


specs/001-heic-jpeg-conversion/spec.md
```diff
@@ -0,0 +1,144 @@
...
+- What happens if storage space is insufficient to store both the temporary HEIC and converted JPEG?
+- How does the system behave when conversion takes longer than expected (e.g., timeout scenarios)?
+- What happens if a HEIC file uses a newer codec version not supported by the conversion library?
+- Multi-frame HEIC containers (Live Photos, bursts): only the primary still image is extracted; additional frames are silently discarded.
+- What happens when compression cannot bring an image below 3MB without unacceptable quality loss?
+  - The system resizes images with longest side >4096px, then compresses at JPEG quality 80–85; if the result is still >3MB, the best-effort compressed version is stored.
+- What about SVG files — are they compressed?
+  - No. SVG is a vector format; compression/resizing does not apply. SVG files pass through unchanged.
+- What about PNG transparency?
+  - PNG files that require compression are converted to JPEG, which does not support transparency. The alpha channel is composited against a white background.
```
hero101: I have to disagree. That would be annoying to the users that have chosen it specifically for a situation.

hero101 (changes requested):
- The given 25 MB limit is way above the maximum allowed on a storage bucket (15728640 Bytes ~ 15.73MB).
- The 3 MB threshold is too high. I would suggest to compress everything. It's not a file / image storage service - it's an image gallery.
- Compression can be done dynamically based on the size or quality, but the encoding must be non negotiable. That would guarantee the same encoding and quality across all images.
- We can always give the option "upload original image", which might the popular option when since we are not enforcing any limits.
- In general if we consider that image size is not an issue, it will always be an issue for users. Big images load slower, eat up more bandwidth, and are keeping the file service busier (platform issues).

specs/001-heic-jpeg-conversion/spec.md
bobbykolev: Yes, also agree, but what about if the PNG doesn't meet the size conditions? It's lossless so we can't optimise it.
@hero101

bobbykolev: noted, spec and plan updated

hero101 (dismissed):
I am satisfied with this spec.

Just a note: You may get away with even lower quality.

Quality Setting | Visual Impact | Best Use Case
-- | -- | --
95–100 | Pristine | Professional printing or archival.
80–90 | Virtually Indistinguishable | High-end websites, photography portfolios.
60–75 | Good (minor artifacts) | Social media, general web browsing.
Below 50 | Noticeable blur/blocking | Only when file size is an emergency.

src/domain/common/visual/image.compression.service.ts
```diff
@@ -0,0 +1,143 @@
...
+        mimeType: 'image/jpeg',
+        fileName: outputFileName,
+        compressed: true,
+        originalSize,
+        finalSize: outputBuffer.length,
+      };
+    } catch (error: unknown) {
+      // Graceful fallback: log the error and return the uncompressed image
+      // rather than blocking the upload entirely
+      this.logger.warn?.(
```
hero101: i would argue this is an error

src/domain/common/visual/image.compression.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      const longestSide = Math.max(metadata.width ?? 0, metadata.height ?? 0);
+
+      let resizeApplied = false;
+      if (longestSide > MAX_DIMENSION) {
+        pipeline = pipeline.resize({
+          width: MAX_DIMENSION,
+          height: MAX_DIMENSION,
+          fit: 'inside',
+          withoutEnlargement: true,
+        });
```
hero101, line 83: would that mean it will resize in a square? width and height are the same

src/migrations/1770909862763-AddHeicHeifMimeTypes.ts
hero101, line 1: - Guard condition: you guard on NOT LIKE '%image/heic%', but you append both image/heic and image/heif. If a row already had image/heif but not image/heic, this will append image/heif again (duplicate). If that matters, you may want an additional AND "allowedMimeTypes" NOT LIKE '%image/heif%' or a more robust normalization.
- LIKE '%image/%': this is a broad check and will match anywhere. It’s fine if the goal is “any image types at all”, but it can also match unintended substrings in messy data.
- Down migration: it removes exact substrings ,image/heif and ,image/heic. That matches what up() appends (with leading commas), which is good—but it won’t remove variants like image/heic at the beginning of the string (no leading comma) if they ever exist.

hero101 (changes requested):
some questions and suggestions

src/domain/common/visual/image.compression.service.ts
bobbykolev: it doesn't but I'll double-check

bobbykolev: confirmed

hero101 (approved):

PR #5844, opened 2026-02-13
when moving the callout ensure that the classification allowed values are updated and that the selected value is properly set

techsmyth:
The need to move callouts has come up multiple times in the past month, and each time Neil had to do it - which was time consuming as needed to be executed by global admin role and there were some issues found. This PR addresses the issues found, and expands who can do the action.

Key changes:
* ensuring that the callout is visible on the new calloutsSet by always being put in the default state of the flow
* urls work now directly after the move (cleared out urls for any profiles related to the callout)
* the createdBy is not changed to the current user (feedback from TIP change)

In addition, the ability to transfer callouts is expanded to also allow global support managers to do this. 

## Schema Contract Checklist (Feature 002)

If this PR affects the GraphQL schema, complete ALL items below:

- [ ] Ran `npm run schema:print` (and optionally `npm run schema:sort`) to regenerate `schema.graphql`.
- [ ] Retrieved base snapshot (e.g. `git show origin/develop:schema.graphql > tmp/prev.schema.graphql`).
- [ ] Ran `npm run schema:diff` to produce `change-report.json` and `deprecations.json`.
- [ ] Reviewed classifications; only expected changes present.
- [ ] (Optional) Ran `npm run schema:validate` and artifacts passed validation.
- [ ] Committed ONLY `schema.graphql` (not the JSON artifacts).

### Change Report Summary

Paste the key counts from `change-report.json` (example format below) — remove any zero lines if desired:

```
Additive: X
Deprecated: Y
Deprecation Grace: Z
Breaking: B (override applied? yes/no)
Premature Removals: P
Invalid Deprecations: I
Info: N
```

### Deprecations Added / Updated

List new or updated deprecations with schedules (`REMOVE_AFTER=YYYY-MM-DD | reason`). Indicate any in grace period.

### Breaking Changes (If Any)

If BREAKING changes are intentional:

- Rationale:
- Risk assessment / mitigation:
- CODEOWNER approval with phrase `BREAKING-APPROVED` requested? (link)

### Other Notes

---

Reference docs: `specs/002-schema-contract-diffing/quickstart.md` for full workflow and troubleshooting.

## Summary by CodeRabbit

* **New Features**
  * Callout transfer now applies classification tagset templates so classifications stay in sync.
  * URL caches are revoked during transfers to ensure updated links are served.

* **Bug Fixes**
  * Fixed tagset input handling for more reliable classification updates.

* **Security / Permissions**
  * Transfer-related permissions expanded to include an additional support role.

* **Chores**
  * Internal integrations improved to support classification updates and URL cache revocation during transfers.


.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
     "source.organizeImports.biome": "explicit"
   },
   "[typescript]": {
-    "editor.defaultFormatter": "biomejs.biome"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
bobbykolev: @valentinyanakiev, what default formatter should we use on the server?

techsmyth: am pretty sure it should be biome, this was a mistake on my part...

hero101 (changes requested):
- I have made some changes to how the data is loaded
- When transferring from somewhere to Space L0 it's fine
- When transferring from somewhere to L1, the callout is transferred but does not show up in either of the tabs. You can only find it in the Index dialog.

hero101 (approved):

PR #5847, opened 2026-02-16
Allow avif and gif file formats for visuals

bobbykolev:
- extending the allowed file formats with avif and gif;
- making sure we're not blocking the callout creation if some visuals fails to upload;

### DIFF
--- a/src/domain/collaboration/callout-framing/callout.framing.service.ts
+++ b/src/domain/collaboration/callout-framing/callout.framing.service.ts
@@ -25,9 +25,10 @@ import { IWhiteboard } from '@domain/common/whiteboard/whiteboard.interface';
 import { WhiteboardService } from '@domain/common/whiteboard/whiteboard.service';
 import { ProfileDocumentsService } from '@domain/profile-documents/profile.documents.service';
 import { IStorageAggregator } from '@domain/storage/storage-aggregator/storage.aggregator.interface';
-import { Injectable } from '@nestjs/common';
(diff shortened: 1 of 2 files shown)


src/domain/collaboration/callout-framing/callout.framing.service.ts
```diff
@@ -220,20 +223,31 @@ export class CalloutFramingService {
...
+          if (newUrl) {
+            const visual =
+              await this.mediaGalleryService.addVisualToMediaGallery(
+                calloutFraming.mediaGallery.id,
+                visualData.name,
+                visualData.sortOrder
+              );
+            visual.uri = newUrl;
+            calloutFraming.mediaGallery.visuals.push(visual);
+            await this.mediaGalleryService.saveVisual(visual);
```
hero101, line 241: try to find a way to reduce the db requests - it's now saving each visual separately.
I do not see the framing being saved? I suspect it's saved from the outside... That being said, do we need to save the media gallery?

hero101 (changes requested):

src/domain/collaboration/callout-framing/callout.framing.service.ts
bobbykolev: I'm not sure, this is part of the initial logic, it's just moved in try catch. Let's leave that out of scope for this release, please. @ccanos can confirm tomorrow.

hero101 (approved):

PR #5853, opened 2026-02-17
File upload error code

bobbykolev:


ccanos (approved):

polibb (approved):
okay, so always getting the message that upload failed, not other types of errors - doesn't affect the error codes registry we added, as it's mapped-to-user-messages in the base exception only

PR #5884, opened 2026-03-02
038 pwa push notifications

### LINKED ISSUE #1797: Progressive Web App (PWA) Push Notifications
## Summary
Implement push notification support for the Alkemio platform as a Progressive Web App (PWA).

## Goals
- Enable real-time push notifications to users via the PWA
- Support notification delivery even when the app is not actively open
- Integrate with the existing Alkemio notification system

## Scope
- Service Worker registration and lifecycle management
- Push API integration (subscription, delivery, display)
- Backend support for sending push messages (web-push protocol)
- User opt-in/out preferences for push notifications
- Notification payload design and display customization

valentinyanakiev:
## Summary

- **Push notification infrastructure**: Server-side W3C Web Push (VAPID) delivery pipeline — subscription management, throttling, per-event push content, and automated cleanup
- **New `PushSubscription` entity** with GraphQL mutations (`subscribeToPushNotifications`, `unsubscribeFromPushNotifications`) and queries (`vapidPublicKey`, `myPushSubscriptions`)
- **Extended notification settings**: Added `push` channel to `IUserSettingsNotificationChannels` (mirrors `inApp` default), with JSONB migration for existing users
- **All 30 notification events** send push via the existing adapter pattern (Space, User, Platform, Organization, VC adapters)
- **Throttling**: Redis-based 10/min/user rate limit with silent drop
- **Subscription lifecycle**: Max 10 per user (oldest replaced), daily cron cleanup of stale/expired, cascade delete on user removal
- **Bug fix**: Removed competing `connectMicroservice` Transport.RMQ consumer on `alkemio-push-notifications` queue that was silently dropping ~50% of push messages — push delivery uses `@golevelup/nestjs-rabbitmq` `@RabbitSubscribe` exclusively
- **Spec updates**: Aligned spec/plan/data-model documents with implementation (retry strategy, RabbitMQ consumer approach, logging fields)

## Key Files

| Area | Files |
|------|-------|
| Domain | `src/domain/push-subscription/` (entity, service, resolvers, DTOs) |
| Adapter | `src/services/adapters/notification-push-adapter/` (adapter, delivery, throttle) |
| Notification integrations | `notification.{space,user,platform,organization,virtual.contributor}.adapter.ts` |
| Settings | `user.settings.service.ts`, `user.service.ts` (defaults), migration |
| Config | `alkemio.yml`, `alkemio.config.ts` |
| Fix | `src/main.ts` (removed competing consumer) |

## Test plan

- [x] Unit tests for `PushSubscriptionService` (cap enforcement, expiry, cleanup)
- [x] Unit tests for `PushDeliveryService` (delivery, 410 handling, error handling)
- [x] Unit tests for `PushThrottleService` (rate limiting)
- [x] Unit tests for `NotificationPushAdapter` (filtering, subscription lookup)
- [x] Manual: subscribe via GraphQL, trigger event, verify push delivered
- [x] Manual: verify push no longer silently dropped after competing consumer fix

Closes https://github.com/alkem-io/alkemio/issues/1797


rtharakan:
Hi @valentinyanakiev - sorry for the delay in reviewing this - I just studied the spec, and it looks solid. Three points for consideration, which I could not explicitly notice within the spec. 

- The notifications consider several platform actions. Will this also work seamlessly with messages? I think so, since the backend infrastructure is common with actions like comments, but confirming. 
- One of the challenges with today's in-app notifications is that it doesn't disappear when you open the notification; user actually has to "mark as read". This is a no-go for mobile notification. Once the user opens the notification, it has to go - the behaviour should not resemble the in-app situation. 
- In terms of delivery, does it specific the mode; I believe there are three levels here - badge, notification center, banners, lock screen. I think most of these won't apply for PWA, so good to be clear. 

That's it for now, but looks a top specification for an MVP.

valentinyanakiev:
Related to https://github.com/alkem-io/alkemio/issues/1797

Client-web PR: https://github.com/alkem-io/client-web/pull/9433

.build/traefik/http.yml
```diff
@@ -272,6 +272,7 @@ http:
         - 'web'
       priority: 150
 
+
```
bobbykolev: not sure we need change in this file

.env.docker
```diff
@@ -133,3 +133,9 @@ LOCAL_STORAGE_PATH=/storage
 CHROMA_SERVER_AUTHN_CREDENTIALS='root:$2y$05$Zj4eHEt9RHn96JlRPPq82Olxjr4CbQTN7Wgcskfu4hHyuMtHXjpRe'
 CHROMA_SERVER_AUTHN_PROVIDER=chromadb.auth.basic_authn.BasicAuthenticationServerProvider
 GEOAPIFY_API_KEY=your-geoapify-api-key
+
+# Push Notifications
+PUSH_NOTIFICATIONS_ENABLED=true
+VAPID_PUBLIC_KEY=BKqdDhb3qBD7lNiqe-sxXQQqmxdMNfBjLPcj9kypfXfxv1LT_c4JS5gukL0Np3lhdd_jqqlZNR714gr2SGT-V44
```
bobbykolev: I guess this key is safe for the public file/access?

.env.docker
```diff
@@ -133,3 +133,9 @@ LOCAL_STORAGE_PATH=/storage
...
+VAPID_PRIVATE_KEY=KGcZQvup8fMXYxkK1HZTfYJRqvnAOZ6x_F85dqiYcq8
+VAPID_SUBJECT=mailto:notifications@alkem.io
```
bobbykolev: Not sure if the right email?

CLAUDE.md
```diff
@@ -187,7 +187,8 @@ Key rules:
 - TypeScript 5.3, Node.js 22 LTS (Volta pins 22.21.1) + NestJS 10, TypeORM 0.3, Apollo Server 4, GraphQL 16 (034-unit-tests)
 - N/A (no data model changes — test-only feature) (034-unit-tests)
 - PostgreSQL 17.5
-- PostgreSQL 17.5 (conversation, conversation_membership, room tables) (040-group-conversations)
+- TypeScript 5.3, Node.js 22 LTS (Volta pins 22.21.1) + NestJS 10, TypeORM 0.3, Apollo Server 4, GraphQL 16, `web-push` (new), `@golevelup/nestjs-rabbitmq` 6.1.0 (existing), `cache-manager-redis-store` 2.0.0 (existing) (038-pwa-push-notifications)
```
bobbykolev, line 190: Are we updating Claude.md with all of the feature branches?

bobbykolev (dismissed):
More of a questionnaire to understand rather then PR feedback. 
Not tested, but logically look nice! 
The goal of this implementation is not only to have the web notifications but to support future PUSH_NOTIFICATIONS for native apps, isn't it?

hero101 (approved):

hero101 (approved):
Functionally tested. Works most of the time. It doesnt seem reliable a 100% of the time.
The automated testing of this will be interesting...

hero101 (changes requested):
- The default env variables need adjusting. It's 'false' by default. Also the keys do not have any defaults.

.env.docker
valentinyanakiev: this is dead weight, doesn't do anything - removed it.

.env.docker
valentinyanakiev: this is not actually used - removed it (.env.docker variables for the current context)

CLAUDE.md
valentinyanakiev: I am not usre either, but it keeps updating it. Not sure it adds value, but certainly it doesn't break anything.

.build/traefik/http.yml
valentinyanakiev: probably accidental keystroke or auto-format, removing it.

hero101 (approved):

PR #5901, opened 2026-03-06
fix: prevent anonymous users from getting write mode on whiteboards in public spaces (#9146)

valeksiev:
Anonymous users viewing whiteboards via the normal space route were getting write access due to the guest-access credential rule matching GLOBAL_GUEST regardless of access path. Gate the UPDATE_CONTENT check in info() on the presence of a guestName — only users arriving via the public whiteboard URL (with a guest name) should receive write mode.

Ref: alkem-io/client-web#9146

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access control for whiteboard guests: truly anonymous users (no id and no provided guest name) can view but cannot edit whiteboard content; named guests and recognized guest IDs retain appropriate write access.

* **Tests**
  * Added tests covering anonymous vs. guest write-access scenarios to ensure consistent behavior.


bobbykolev (changes requested):
@valeksiev this is fixing the issue, but it causes bigger regression. Now I'm unable to edit a public whiteboard with a guest session.

bobbykolev (approved):

PR #5908, opened 2026-03-09
fix: actor cache invalidation on deletion and Kratos metadata cleanup

antst:
## Summary
- **Cache invalidation gap**: `deleteActorById` was not clearing `@actor:id`, `@actorContext`, or `@actorType` Redis caches, leaving stale actor data accessible for up to 1 hour (cache TTL) after deletion
- **Kratos metadata cleanup**: Always clear `alkemio_actor_id` from Kratos identity `metadata_public` before attempting identity deletion, preventing stale actor references when users re-register (fixes re-registration errors for MS/OIDC users — client-web#9392)
- **Error visibility**: Escalate Kratos identity deletion failures from `warn` to `error` log level
- **Lint cleanup**: Fix pre-existing biome lint errors (formatting, unused imports, import ordering)

## Context
When a Global Admin deletes a user, if the Kratos identity deletion fails silently, the surviving identity retains a stale `alkemio_actor_id` pointing to the deleted actor. On re-registration with the same OIDC provider, this stale reference can cause authentication failures. By clearing the metadata before deletion, the identity is safe regardless of whether the full deletion succeeds.

## Test plan
- [x] `pnpm lint` — 0 errors, 0 warnings
- [x] `pnpm test` — 2810 tests passed, 0 failures
- [ ] Delete a user via GA, verify Kratos identity metadata is cleared
- [ ] Re-register deleted MS user, verify no `actorID` errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Actor deletions now clear all related caches (including actor type) to prevent stale data.
  * User deletion now always clears actor metadata when an authentication ID exists; identity deletion remains optional and logs clearer errors on failures.
* **New Features**
  * Added support to proactively clear actor metadata from external identity records before identity removal.


valentinyanakiev (commented):
## Code Review: Actor Cache Invalidation & Kratos Metadata Cleanup

### Overview

This PR addresses two real bugs:
1. **Stale actor caches** after deletion — `deleteActorById` was missing invalidation for `@actorType` cache
2. **Stale Kratos metadata** — `alkemio_actor_id` in `metadata_public` survives failed identity deletion, causing re-registration errors for OIDC users

Plus ~160 lines of Biome lint/formatting fixes.

### What's Good

- **Clear separation of concerns**: `invalidateActorCache` (credential changes) vs `invalidateAllActorCaches` (deletion) is well-named and intentional
- **Defensive metadata cleanup**: Clearing `alkemio_actor_id` before attempting identity deletion is the right order of operations — if deletion fails, the identity is still safe for re-registration
- **Error escalation**: `warn` → `error` for Kratos identity deletion failures is appropriate — these are actionable operational issues
- **RFC 6902 `add` op**: Good choice over `remove` to avoid failures when the path doesn't exist

### Issues & Suggestions

**1. `clearIdentityActorMetadata` nukes all `metadata_public` (Medium)**

`kratos.service.ts:350-357` — The patch sets `metadata_public` to `{}`, wiping **all** metadata, not just `alkemio_actor_id`. If other keys exist in `metadata_public` (now or in the future), they'll be silently lost.

Consider targeting only the specific key:
```typescript
jsonPatch: [
  {
    op: 'remove',
    path: '/metadata_public/alkemio_actor_id',
  },
]
```

Or if you want the `add`-is-safe semantics, use `add` on the specific path:
```typescript
{ op: 'add', path: '/metadata_public/alkemio_actor_id', value: null }
```

The JSDoc says "clears alkemio_actor_id" but the implementation clears everything — this mismatch could cause subtle bugs.

**2. Metadata clearing happens even when `deleteIdentity` is false (Low-Medium)**

`user.service.ts:468` — The metadata is now cleared whenever `authenticationID` exists, regardless of `deleteData.deleteIdentity`. If a GA deletes a user but explicitly chooses NOT to delete the Kratos identity (maybe to preserve it for audit), the actor metadata is still wiped. Is this intentional? The PR description suggests yes, but worth confirming — a comment clarifying the intent would help.

**3. Cache invalidation should be fire-and-forget or fail-safe (Low)**

`actor.service.ts:124` — `invalidateAllActorCaches` is `await`ed after the DB delete. If any cache operation throws (Redis connectivity issue), the actor is already deleted from the DB but the method throws, potentially breaking the caller's transaction flow. Consider wrapping in a try-catch or making it fire-and-forget.

**4. Lint changes inflate the diff (Nit)**

~160 of 240 additions are formatting-only changes. These are fine but inflate the review surface. Consider splitting lint fixes into a separate commit for cleaner history.

### Test Coverage

- No new unit tests for `clearIdentityActorMetadata` or `invalidateAllActorCaches`
- The existing test suite passes (2810 tests), but the new `KratosService` method and the new cache invalidation path have zero test coverage
- The manual test plan items (delete user via GA, re-register MS user) are appropriate for the Kratos integration but should be supplemented with unit tests for the service methods

### Security

No concerns — the changes operate on internal service calls with existing authorization boundaries.

### Verdict

The core logic is sound and addresses real bugs. The main concern is **item #1** — `clearIdentityActorMetadata` wiping all `metadata_public` rather than just `alkemio_actor_id`. This should be addressed before merge. Items #2-3 are worth a quick discussion.

valentinyanakiev (changes requested):
See my comment :)

antst:
> See my comment :)

It is alright, whole goal is to wipe it all :)

valentinyanakiev (approved):

PR #5924, opened 2026-03-18
Back to lookUpByName. 

bobbykolev:
Instead of resolving the client URLs:
- blocking rendering;
- doubling auth checks;
- coupling client routes with server parsing;

open the lookupByName so every client can gather the ids it needs to continue making required requests based on the UI/client criteria (not hardcoded on the server).

Client PR: https://github.com/alkem-io/client-web/pull/9421


antst (approved):
LGTM

PR #5929, opened 2026-03-19
Fix guidance privilege

bobbykolev:
In order to access the well-known vcs as regular registered user, we had to change the privilege.


antst (approved):
LGTM

PR #5934, opened 2026-03-23
Add upload stream timeout, duration logging, and increase proxy timeouts for file uploads

bobbykolev:
- Added a 60-second timeout to streamToBuffer() to clean up hung connections instead of waiting indefinitely
  - Added verbose logging for upload duration (stream buffering time + total processing time) to aid debugging bulk upload issues
  - Increased Oathkeeper proxy timeouts from Go defaults (~8s) to 120s — the ~8s default was the root cause of 499 errors on slow connections
  - Increased Traefik forwarding timeouts to Oathkeeper to 120s to match

  Root cause

  The 499 "Request disconnected during file upload stream parsing" errors were caused by Oathkeeper's default Go HTTP timeout (~8 seconds). On slow
  networks, file uploads that took longer than ~8s were terminated by the proxy before the server could finish reading the stream.

## Summary by CodeRabbit

* **Infrastructure**
  * Configured timeout settings (120 seconds) for proxy and HTTP routing layers to improve service resilience.

* **Monitoring & Diagnostics**
  * Enhanced logging capabilities for file operations and uploads with performance metrics.
  * Implemented timeout handling for stream processing operations.

### DIFF
--- a/src/common/utils/file.util.ts
+++ b/src/common/utils/file.util.ts
@@ -1,19 +1,41 @@
 import { Readable } from 'stream';
 
-export async function streamToBuffer(stream: Readable): Promise<Buffer> {
+const DEFAULT_STREAM_TIMEOUT_MS = 60_000;
+
+export async function streamToBuffer(
+  stream: Readable,
+  timeoutMs: number = DEFAULT_STREAM_TIMEOUT_MS
+): Promise<Buffer> {
   return new Promise((resolve, reject) => {
     const data: any[] = [];
+    let settled = false;
+
+    const timer = setTimeout(() => {
+      if (!settled) {
+        settled = true;
+        stream.destroy();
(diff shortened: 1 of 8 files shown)


src/common/utils/file.util.ts
```diff
@@ -1,19 +1,41 @@
 import { Readable } from 'stream';
 
-export async function streamToBuffer(stream: Readable): Promise<Buffer> {
+const DEFAULT_STREAM_TIMEOUT_MS = 60_000;
```
valentinyanakiev: Put this in a variable in the config. If you do, the whole changeset will be down to configuration (pretty much) - otherwise changing the timeout will require rebuilding of the server.

valentinyanakiev (changes requested):

valentinyanakiev (approved):

PR #5954, opened 2026-04-01
Preserves push subscription disabled state

valentinyanakiev:

This pull request addresses a bug where push notification subscription preferences were not reliably preserved, particularly across browser refreshes or re-subscriptions. Previously, unsubscribing from push notifications would permanently delete the subscription, leading to a loss of user settings and requiring a full re-subscription process to reactivate.

This change transforms the `unsubscribeFromPushNotifications` mutation to disable a subscription rather than deleting it. This ensures that a user's preference to stop receiving notifications is retained without losing the subscription record itself. It introduces a new `enablePushSubscription` mutation, allowing users to easily re-activate a previously disabled subscription. Additionally, the `subscribe` logic now intelligently preserves the `DISABLED` status of an existing subscription when a user attempts to resubscribe with the same endpoint, preventing unintended re-activation. The `myPushSubscriptions` query has also been updated to return both active and disabled subscriptions, offering a comprehensive overview of the user's notification preferences.

### Additional context

This work is part of the broader effort related to PWA settings and enhancing user control over push notifications.

### Governance

- [x] Documentation is added
- [x] Test cases are added or updated

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -665,6 +665,7 @@ enum ProfileType {
 """Status of a push notification subscription."""
 enum PushSubscriptionStatus {
   ACTIVE
+  DISABLED
   EXPIRED
 }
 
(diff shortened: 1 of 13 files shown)


bobbykolev (changes requested):
The fix does not resolve the reported issue. Currently:
1. if we disable the setting, on every page reload - it persists as enabled (require client fix as the client resubscribes)
2. I'm unable to remove subscriptions from the list below the setting - clicking remove nothing happens

bobbykolev:
@coderabbitai review, focus only on critical/high issues based on the changed files.

bobbykolev (approved):

PR #5967, opened 2026-04-08
Entitlement to control access to OfficeDocuments

techsmyth:
Only adding the licensing setup for the entitlement, not yet using in logic on either server or client. 

Spec + plan + tasks, not yet implemented. 

## Summary by CodeRabbit

* **New Features**
  * Added an Office Documents licensing entitlement to control office-document capabilities at Space and Collaboration levels (SPACE_FLAG_OFFICE_DOCUMENTS / SPACE_FEATURE_OFFICE_DOCUMENTS).

* **Updates**
  * GraphQL schema extended with new credential and entitlement enum values.
  * Licensing data seeded and migration added to provision the new plan and credential rule.

* **Documentation**
  * New spec, quickstart, checklist, and tasks describing behavior, migration, and validation.


specs/083-collab-entitlement/spec.md
rtharakan, line 1: - Is the entitlement only for the "Documents" feature, or does the associated plumbing of "File Service/Storage" require its own entitlement? How does this entitlement map to the storage entitlements? Shouldn't this be decoupled since other "services" could also use this Storage (ex, whiteboards, media gallery, etc)
-  Is this entitlement a combined approach for all kinds of "Documents" or should it be broken down at a feature level? Essentially, is this a "Collabora Service" entitlement, or do we actually need more specific flags to turn on/off features? 
- What is the expected plan on revocation? If a space had a plan and entitlement, created a document and stored it, will revoking the entitlement at the space level, remove the feature completely, or just the "edit" ability? Can they still view? On revocation, what happens to the files? Maybe not an entitlement question.

techsmyth: Answering the questions:
* Yes this entitlement is for the Office Documents feature only. It does not impact the file storage / service at all. The storage entitlement is orthogonal. 
* Yes it is for all of the different Office Document formats supported. No we do not need more granularity.
* Re revocation, this entitlement only impacts the creation of posts with office documents. If the entitlement is revoked, then for now I think we just leave any existing posts working as they were.

Essentially this entitlement allows the server and the client to control if the space is entitled to allow the creation of posts with office documents as the content.

rtharakan:
Hi @valentinyanakiev - the entitlement spec looks good from the business perspective. Neil has answered my queries. Any other blockers for implementation? If not, please go ahead (or would you prefer I speak to @hero101)?

valentinyanakiev:
## QA Test Plan, Execution Log & Deployment Notes

This comment documents what was verified locally, the risk surface, and the checks to run pre- and post-deploy. It complements the three-phase quickstart in `specs/083-collab-entitlement/quickstart.md`.

### 1. Test impact — what this PR changes

**Code surface (runtime behaviour):**
- Two new enum members: `LicenseEntitlementType.SPACE_FLAG_OFFICE_DOCUMENTS`, `LicensingCredentialBasedCredentialType.SPACE_FEATURE_OFFICE_DOCUMENTS`.
- Default entitlement entry added to three license-init arrays (Collaboration, Space, TemplateContentSpace).
- One `case` added to each of the two `extendLicensePolicy` switches (Collaboration copies from parent; Space checks the credential on the L0 agent and sets `limit = 1, enabled = true`).

**Data surface:**
- New migration `1776692645254-AddOfficeDocumentsEntitlement.ts`:
  - Inserts one row into `license_plan` (`name = 'SPACE_FEATURE_OFFICE_DOCUMENTS'`, `sortOrder = 100`).
  - Appends one element to `license_policy.credentialRules` jsonb (single existing row).
  - Idempotent on re-run, symmetric `down()`.
- Bootstrap JSON (`license-plans.json`) kept in sync so fresh installs converge.

**API surface:**
- GraphQL schema: additive enum values only — no removals, no required-field changes. Diff classified `ADDITIVE` for all three new enum members.

**Explicitly out of scope (no consumer yet):**
- No code reads `SPACE_FLAG_OFFICE_DOCUMENTS` to gate behaviour — this PR is pure plumbing. The consumer (whatever calls `isEntitlementEnabled(...)` in the office-docs render path) is a follow-up.

### 2. Tests executed locally

| Suite / command | Result | Notes |
|---|---|---|
| `pnpm lint` (`tsc --noEmit` + `biome check`) | ✅ 0 errors | One pre-existing warning (unused `EntityManager` in `space.service.ts`), unrelated |
| `pnpm test:ci:no:coverage` | ✅ 6283 passed, 7 skipped, 0 failed | Full Vitest suite, 19.1s wall |
| `pnpm run migration:run` | ✅ applied | Verified row inserted + credential rule appended |
| `pnpm run migration:revert` | ✅ reverted | Verified row gone + credential rule removed (rules count 9 → 8) |
| `pnpm run migration:run` (second) | ✅ idempotent | State parity with first run |
| `pnpm run schema:print && pnpm run schema:sort` | ✅ regenerated | `schema.graphql` + `schema-lite.graphql` updated in-place |
| `pnpm run schema:diff` | ✅ additive | My three additions marked `ADDITIVE`; pre-existing `BREAKING` entries are the unrelated Conversation rework on this branch |
| `pnpm run schema:validate` | ✅ valid | Change report + deprecations schema-valid |

**Unit test additions** (three targeted assertions for the new entitlement):
- `space.service.license.spec.ts`:
  - Pre-apply: `SPACE_FLAG_OFFICE_DOCUMENTS` reports `limit = 0`.
  - After `extendLicensePolicy` with L0 agent holding the credential: `enabled = true, limit = 1`.
  - Recursive subspace with only an intermediate (non-L0) agent: remains `enabled = false, limit = 0` (confirms FR-003a, L0-only determinant).
- `collaboration.service.license.spec.ts`: new entitlement added to parameterised fixtures; `findAndCopyParentEntitlement` call count +1.
- `template.content.space.service.spec.ts`: `createLicenseTemplateContentSpace` asserts `SPACE_FLAG_OFFICE_DOCUMENTS` is present with `enabled: true, limit: 0` in the template default set.

### 3. End-to-end manual verification (three-phase quickstart, SC-008)

Executed against a local dev stack restored from the **prod** backup set (2233 users, 10 license plans pre-migration) with the new migration then applied on top. A freshly-created test Space was used because existing pre-migration Spaces do not carry the new default entitlement entry until a license-reset flow runs (see §6 below).

Test artefact: Space `Office Docs E2E Test`, id `0a97c848-2978-4955-b067-ebd8de353d6d`.

| Phase | Action | Space license (`SPACE_FLAG_OFFICE_DOCUMENTS`) | Collab license (`SPACE_FLAG_OFFICE_DOCUMENTS`) | Pass |
|---|---|---|---|---|
| 1 | Create Space, no plan assigned | `enabled: false, limit: 0` | `enabled: false, limit: 0` | ✅ |
| 2 | `assignLicensePlanToSpace(SPACE_FEATURE_OFFICE_DOCUMENTS)` | `enabled: true, limit: 1` | `enabled: true, limit: 1` | ✅ |
| 3 | `revokeLicensePlanFromSpace(...)` | `enabled: false, limit: 0` | `enabled: false, limit: 0` | ✅ |

Credential → Space → Collaboration cascade (FR-003 / FR-003a) works exactly as specified.

**Query used** (for re-execution during deploy validation):

```graphql
query SpaceEntitlements($spaceId: UUID!) {
  lookup {
    space(ID: $spaceId) {
      id
      license { entitlements { type enabled limit } }
      collaboration {
        id
        license { entitlements { type enabled limit } }
      }
    }
  }
}
```

**Mutations used:**
- `assignLicensePlanToSpace(planData: { licensePlanID, spaceID })`
- `revokeLicensePlanFromSpace(planData: { licensePlanID, spaceID })`

### 4. Direct DB checks (runbook)

After the migration runs in any environment, the following should hold:

```sql
-- Should return exactly one row with sortOrder = 100
SELECT name, "licenseCredential", "sortOrder", type, enabled
FROM license_plan
WHERE name = 'SPACE_FEATURE_OFFICE_DOCUMENTS';

-- Should return exactly one element matching the new credentialType
SELECT jsonb_path_query("credentialRules",
  '$[*] ? (@.credentialType == "space-feature-office-documents")')
FROM license_policy;

-- The rules array should have grown by exactly one vs. pre-deploy snapshot
SELECT jsonb_array_length("credentialRules") FROM license_policy;
```

### 5. Risk areas

**Low risk — but worth watching:**

1. **jsonb `||` concatenation path.** The migration appends via `"credentialRules" || $1::jsonb`. If an environment's `credentialRules` column were ever `NULL` (it isn't in prod/dev/acc today — all seeded from the baseline migration), the concatenation would yield `NULL`. Idempotency guard (`alreadyPresent` check) runs before the update and also reads the column, so a NULL would surface as "skip" rather than "corrupt". Still: please eyeball `SELECT COUNT(*) FROM license_policy WHERE "credentialRules" IS NULL` pre-deploy — expected = 0.

2. **Existing Spaces don't gain the new entitlement entry automatically.** A pre-migration Space's `license.entitlements` array was written to the DB at Space-creation time and does not include the new type. The `reset()` function resets existing entries' *values*; it does not insert missing rows. Consequence:
   - On production, every existing Space's license entitlements query will **omit** `SPACE_FLAG_OFFICE_DOCUMENTS` until its license is re-initialised.
   - This is consistent with how `SPACE_FLAG_MEMO_MULTI_USER` rolled out — it's the established pattern, not a new regression — but it **does diverge from the literal reading of spec FR-008** ("existing Spaces… must report the entitlement with `limit = 0`").
   - **Mitigation if product wants FR-008 strict**: a one-time data migration that iterates every `license` row of type `SPACE` / `COLLABORATION` / `TEMPLATE_CONTENT_SPACE` and inserts the missing entitlement entry with defaults (`limit: 0`, `enabled: false`/`true` per init precedent). Not included in this PR — flag for follow-up if Product wants it.
   - **Explicit check post-deploy**: on a few known existing Spaces, verify the absence of the new entry is *silent* (query returns without error) and does **not** break any downstream BFF / web-client path that might assume all known enum members are present.


### 6. Follow-ups / known gaps (not blocking this PR)

- **Consumer wiring** — no code currently reads `SPACE_FLAG_OFFICE_DOCUMENTS` to gate the office-docs feature. Flag separately.
- **FR-008 strict mode** — see §5 item 2. Decide with Product whether an entitlement-backfill migration is required.

### 7. Pre-merge checklist

- [x] Unit tests pass
- [x] Migration up → down → up verified idempotent
- [x] Schema diff is additive for this feature's changes
- [x] Bootstrap JSON in sync with migration (`sortOrder = 100`, credential name matches)
- [x] No edits to `src/migrations/1764590884533-seed.ts` (historical, untouched)
- [x] No edits to `src/domain/common/memo/memo.service.ts` (out-of-scope)
- [x] No edits under `src/platform-admin/` or License-Issuer admin UI paths
- [x] End-to-end cascade verified on a freshly-created Space (three-phase, prod-data DB)

### 8. Deploy runbook

**Pre-deploy**
1. Confirm `schema-baseline.yml` workflow green on develop before release branch cut.
2. Snapshot `license_plan` row count and `license_policy.credentialRules` array length in target env.

**Deploy**
3. Apply migrations — the new one runs as `AddOfficeDocumentsEntitlement1776692645254`, two SQL statements (INSERT + UPDATE), O(1) cost.
4. Post-deploy verify via the SQL block in §4.

**Smoke**
5. Create one throwaway Space → confirm entitlement present with `limit: 0`.
6. `assignLicensePlanToSpace(SPACE_FEATURE_OFFICE_DOCUMENTS)` → confirm `limit: 1` cascades to Collaboration.
7. `revokeLicensePlanFromSpace(...)` → confirm return to `limit: 0`.
8. Delete the throwaway Space.

**Rollback**
- `pnpm run migration:revert` cleanly removes the row and the credential-rule element. Any Spaces created during the window between deploy and rollback will retain the new entitlement entry in their `license.entitlements` array (harmless — the type just won't appear in any credential rule anymore, so the licensing engine will leave it disabled). No data cleanup required.

rtharakan:
Thanks, @valentinyanakiev. On the risks, #1. seems to be technical (and low-risk). I don't think I can confidently inform the mitigation. 

#2 seems to be correct, and it makes sense that existing spaces don't gain the entitlement automatically. Looks good for PR review.

valentinyanakiev:
> Thanks, @valentinyanakiev. On the risks, #1. seems to be technical (and low-risk). I don't think I can confidently inform the mitigation.
> 
> #2 seems to be correct, and it makes sense that existing spaces don't gain the entitlement automatically. Looks good for PR review.

Well, ok - merging then, seems like all is a managed risk. The rest will be verified by @Comoque1

PR #5977, opened 2026-04-17
fix: resolve UQ_actor_nameID_user collision blocking OIDC registration

hero101:
## Problem

OIDC registration (LinkedIn / Microsoft) sometimes failed with `500 /webhooks/kratos/post-login — giving up after 3 attempts`. Root cause: `UserService.createUser` generated a nameID from a reservation list that only read the `User` entity, but the unique index `UQ_actor_nameID_user` lives on `actor.nameID WHERE type='user'`. An orphan `actor` row (type=user with no matching user row) was invisible to the generator, so every retry picked the same colliding slug and the CTI insert failed deterministically — 3 kratos retries, all 500, flow dies.

## Fix

- `naming.service.ts` — `getReservedNameIDsInUsers` now queries `Actor` with `type='user'`, aligning reservation reads with the actual unique index. Orphans and any concurrently-inserting rows are visible.
- `user.service.ts` — `createUser` wraps the transaction in a bounded retry (max 3) that regenerates the nameID on `UQ_actor_nameID_user` violation. Only applies when the server generated the nameID; explicit caller-supplied nameIDs still surface the collision.
- Tests added for both paths: orphan inclusion in reservation query, retry on generated nameID, no retry on explicit nameID, no retry on unrelated unique violations.

## Test plan
- [x] `pnpm test src/services/infrastructure/naming/naming.service.spec.ts src/domain/community/user/user.service.spec.ts` — 79/79 pass
- [x] `tsc --noEmit` — clean
- [ ] Smoke: register via LinkedIn on dev after rollout; verify `alkemio-server` log `Identity resolve: created user …` and no `UQ_actor_nameID_user` errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved user creation reliability by implementing automatic retry logic for nameID collisions. When a system-generated nameID conflicts with existing records, the service regenerates a new nameID and retries the creation process. Explicitly provided nameIDs remain unaffected and will fail on collision as expected.

### DIFF
--- a/src/domain/community/user/user.service.spec.ts
+++ b/src/domain/community/user/user.service.spec.ts
@@ -15,10 +15,12 @@ import { ConfigService } from '@nestjs/config';
 import { Test, TestingModule } from '@nestjs/testing';
 import { getRepositoryToken } from '@nestjs/typeorm';
 import { KratosService } from '@services/infrastructure/kratos/kratos.service';
+import { NamingService } from '@services/infrastructure/naming/naming.service';
 import { MockCacheManager } from '@test/mocks/cache-manager.mock';
 import { MockWinstonProvider } from '@test/mocks/winston.provider.mock';
 import { defaultMockerFactory } from '@test/utils/default.mocker.factory';
 import { repositoryProviderMockFactory } from '@test/utils/repository.provider.mock.factory';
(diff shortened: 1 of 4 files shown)


antst (approved):
LGTM

PR #6000, opened 2026-04-24
Handle file-service-go dedup reuse: release pre-created auth/tagset

antst:
## Summary

file-service-go v0.0.10 adds row-level content dedup: `POST /internal/file` returns the existing `file` row when content + storage bucket match (`reused: true` in the body). Caller-supplied `authorizationId` / `tagsetId` are ignored on reuse — the existing row keeps its own. Without this change, the server would leak the rows it pre-creates before each Go call.

This PR wires the server adapter into the new contract.

### Fixes

- **test-suites regression "upload same file twice"** — two uploads of the same file into the same bucket now return the same document URI (sequential dedup in Go). Test passes without any change on our side beyond bumping the image.
- **Pre-created metadata leak** — when Go returns \`reused: true\`, the server now releases the just-saved \`authorization_policy\` and \`tagset\` rows.
- **Reused-document corruption risk (latent)** — the existing catch branch in \`uploadFileAsDocumentFromBuffer\` used to \`fileServiceAdapter.deleteDocument(result.id)\` on any post-Go failure. If \`result.reused\` was true and reload failed, that would delete another caller's live document. Now the Go-side delete only fires when \`!result.reused\`.

## Changes

- \`CreateDocumentResult\` DTO: add optional \`reused?: boolean\`. Missing is treated as \`false\` (forward-compat with pre-v0.0.10 Go versions).
- \`StorageBucketService.uploadFileAsDocumentFromBuffer\`:
  - After reload, if \`result.reused\` is true, \`tryRollback\`-delete the saved auth + tagset.
  - Catch branch: skip Go-side \`deleteDocument\` when \`result.reused\` is true.
- \`quickstart-services.yml\`: bump \`alkemio/file-service-go\` v0.0.9 → v0.0.10.

## Paired work

- **file-service-go PR #11**: https://github.com/alkem-io/file-service-go/pull/11 — introduces the dedup + \`reused\` signal + HTTP 201 always. Merges and publishes as v0.0.10 before this server PR.

## Not in this PR (intentional)

Unique index on \`file(externalID, storageBucketId)\` is **not** added. Pre-migration prod duplicates can't be merged safely without rewriting URL references across user content (markdown, whiteboard JSON, reference URIs, etc.). The sequential dedup path handles the common case; concurrent races remain theoretically possible but rare. We'll revisit once duplicates have drained naturally.

## Test plan

- [x] 3 new unit tests in \`storage.bucket.service.spec.ts\` covering reused=true (cleanup), reused=false (no cleanup), and reused=true + post-upload failure (catch must not delete the reused doc).
- [x] Full test suite: 6373 passed.
- [ ] Manual: on dev cluster with file-service-go v0.0.10 running, upload the same file twice and confirm identical URIs + no leaked auth/tagset rows.

## Summary by CodeRabbit

* **New Features**
  * Upload results now indicate when an uploaded file was reused (deduplicated).

* **Bug Fixes**
  * Prevents accidental deletion of shared/deduplicated files during rollbacks.
  * Ensures temporary authorization and tag metadata are cleaned up when uploads are reused.

* **Tests**
  * Added unit tests covering deduplication outcomes and cleanup behaviors.

* **Chores**
  * Updated file service image to v0.0.10.

### DIFF
--- a/quickstart-services.yml
+++ b/quickstart-services.yml
@@ -499,7 +499,7 @@ services:
       - 'host.docker.internal:host-gateway'
     container_name: alkemio_dev_file_service
     hostname: file-service
-    image: alkemio/file-service-go:v0.0.9
+    image: alkemio/file-service-go:v0.0.10
     depends_on:
       postgres:
         condition: service_healthy
--- a/src/domain/storage/storage-bucket/storage.bucket.service.spec.ts
+++ b/src/domain/storage/storage-bucket/storage.bucket.service.spec.ts
@@ -533,6 +533,42 @@ describe('StorageBucketService', () => {
       expect(fileServiceAdapter.deleteDocument).not.toHaveBeenCalled();
     });
 
+    it('treats omitted reused as false (backward-compat with pre-v0.0.10 Go)', async () => {
+      // Older file-service-go versions don't emit the `reused` field. Missing
+      // must behave the same as `reused: false` — don't release pre-created
+      // auth/tagset. Pins the forward-compat contract documented on the DTO.
+      const bucket = mockStorageBucket({ id: 'bucket-legacy' });
+      const buffer = Buffer.alloc(10);
+      const freshDoc = mockDocument({ id: 'doc-legacy' });
+
+      (storageBucketRepository.findOneOrFail as Mock).mockResolvedValue(bucket);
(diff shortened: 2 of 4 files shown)


antst:
@coderabbit review

hero101 (approved):

PR #6017, opened 2026-04-29
Fix createTemplateFromContentSpace crash: split contributionDefaults phase-1/phase-2

antst:
## Summary

`createTemplateFromContentSpace` crashes with:

> Storage bucket must be persisted before document re-upload: caller must save the parent entity first

## Root cause

`CalloutContributionDefaultsService.createCalloutContributionDefaults` was running file-service-go work (`reuploadDocumentsInMarkdownToStorageBucket`) inline on `callout.framing.profile.storageBucket` — which is in-memory at that point and has no DB id yet.

Stack from the user's report:

```
ProfileDocumentsService.reuploadFileOnStorageBucket  (storageBucket.id undefined)
  ← CalloutContributionDefaultsService.createCalloutContributionDefaults
    ← CalloutService.createCallout
      ← CalloutsSetService.addCallouts
        ← CollaborationService.createCollaboration
          ← TemplateContentSpaceService.createTemplateContentSpace
            ← TemplateService.createTemplate
              ← TemplatesSetService.createTemplateFromContentSpace
```

Latent since #5969; surfaced now because PR #6014's fail-fast guard in `ProfileDocumentsService.reuploadFileOnStorageBucket` turns the previously-silent misuse into a hard error.

## Fix

Same root pattern as PR #6014: split phase-1 (in-memory build) from phase-2 (file-service work post-save).

- `createCalloutContributionDefaults` is now synchronous, in-memory only; drops the `storageBucket` parameter. Stores `postDescription` raw.
- New `materializeCalloutContributionDefaultsContent(defaults, storageBucket, rollback)` does the markdown re-home and saves the entity directly when content changes (the parent callout isn't re-saved by the caller post-materialize, so we save here). Same rollback convention as the OrRollback helper.
- `CalloutService.createCallout` calls phase-1 only.
- `CalloutService.materializeCalloutContent` calls phase-2 right after the framing materialize, using the now-persisted `callout.framing.profile.storageBucket`.

## Audit

Searched every call site of `profileDocumentsService.reupload*` and `fileServiceAdapter.*` from create flows. The only remaining phase-1-on-unsaved-bucket site was `callout-contribution-defaults`. Other reupload call sites are either:

- Update paths (`whiteboard.service.ts:297`, `memo.service.ts:211`) — entity already persisted
- `callout-framing.service.ts:308` (mediaGallery) — bucket is saved eagerly inside `createMediaGallery` before reupload runs
- `profile.service.ts:400` (`addVisualsOnProfile`) — only invoked from phase-2 `materializeProfileContentAndVisuals`

So this is the only remaining occurrence of the same pattern.

## Test plan

- [x] 6393 unit tests pass (1 new for the materialize method's rollback path)
- [x] lint and typecheck clean
- [ ] Verify `createTemplateFromContentSpace` flow end-to-end on a fresh DB

## Related

Fix-up for #6014. Companion to #6016 (KB bootstrap fix, already merged).

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal architecture by separating content creation and materialization into distinct phases, improving initialization efficiency and system reliability. Contribution defaults content is now processed in a deferred phase to ensure proper bucket assignment.

### DIFF
--- a/src/domain/collaboration/callout-contribution-defaults/callout.contribution.defaults.service.spec.ts
+++ b/src/domain/collaboration/callout-contribution-defaults/callout.contribution.defaults.service.spec.ts
@@ -158,6 +158,40 @@ describe('CalloutContributionDefaultsService', () => {
       ).rejects.toThrow('reupload failed');
       expect(rollback).toHaveBeenCalled();
     });
+
+    it('logs and preserves the original error when rollback itself fails', async () => {
+      const defaults = {
+        id: 'defaults-id',
+        postDescription: 'markdown with ![image](http://old-url)',
+      } as CalloutContributionDefaults;
+      const originalFailure = new Error('reupload failed');
+      const rollbackFailure = new Error('rollback also failed');
+      const failingRollback = vi.fn().mockRejectedValue(rollbackFailure);
+
+      vi.mocked(
+        profileDocumentsService.reuploadDocumentsInMarkdownToStorageBucket
+      ).mockRejectedValue(originalFailure);
+      const errorLog = vi.spyOn(
+        service['logger'] as unknown as { error: (...args: unknown[]) => void },
+        'error'
+      );
+
+      // Original reupload error must propagate; the rollback failure is
+      // logged at ERROR but never replaces the materialization error.
+      await expect(
+        service.materializeCalloutContributionDefaultsContent(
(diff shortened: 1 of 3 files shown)


antst:
@coderabbit review

PR #6028, opened 2026-04-29
fix: stop openai_assistant crash-loop + document AUTH_ADMIN reqs

valentinyanakiev:
## Summary

Two unrelated startup-time fixes for the unified VC stack, both surfaced while testing the post-#6027 setup.

### 1. `openai_assistant` was crash-looping

```
ValidationError: 1 validation error for BaseConfig
  Value error, LLM_API_KEY is required for provider 'mistral'.
  Set LLM_API_KEY or provide LLM_BASE_URL for local models.
```

The plugin talks directly to OpenAI's Assistants API and never uses the main LLM, but `BaseConfig` still validates that `MISTRAL_API_KEY` (or `LLM_BASE_URL`) is set on every plugin. The container's env list in `overlays/ai-stack.yml` omitted both → validation fails before the plugin starts → restart loop (~30 errors / 15 min observed).

Fix: pass `MISTRAL_API_KEY` and `MISTRAL_BASE_URL` through to the openai_assistant container. The plugin won't read them but the validator is satisfied.

**Verified live**: container now logs `Engine ready — consuming from virtual-contributor-engine-openai-assistant` instead of `ValidationError`.

### 2. `AUTH_ADMIN_*` requirement was undocumented

The placeholders in `.env.docker` and `.env.docker.scaleway.example` (`admin@alkem.io` / `password`) trigger a 400 on the Kratos `/self-service/login` flow when no matching identity exists in the target environment, which silently breaks `ingest-space`:

```
Space ingestion failed: Client error '400 Bad Request' for url
  'http://localhost:3000/ory/kratos/public/self-service/login?flow=...'
```

Fix: add a comment block at the `AUTH_ADMIN_*` lines in both env files explaining the requirement so this is discoverable instead of mysterious-400.

## Test plan

- [x] `pnpm test` — 6375 tests pass (pre-commit ran the full vitest suite)
- [x] **Live verified** — `openai_assistant` container starts cleanly after recreate
- [ ] Confirm in dev/test environment that ingest-space works after AUTH_ADMIN_* are set to a valid Kratos identity

## Summary by CodeRabbit

* **Chores**
  * Added environment variables for authentication configuration.
  * Updated AI service configuration with necessary API credentials and settings.

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -165,6 +165,10 @@ VECTOR_DB_CREDENTIALS=username:password
 VECTOR_DB_CREDENTIALS=root:toor
 
 
+# Used by ingest-space to authenticate against Kratos and call the private
+# GraphQL API to fetch space contents. Must be a valid platform-admin Kratos
+# identity in the target environment — otherwise space ingestion 400s on the
+# Kratos /self-service/login flow.
 AUTH_ADMIN_EMAIL=admin@alkem.io
 AUTH_ADMIN_PASSWORD=password
 EMAIL_FROM=notifications@mailer.alkem.io
--- a/.env.docker
+++ b/.env.docker
@@ -166,7 +166,7 @@ VECTOR_DB_CREDENTIALS=root:toor
 
 
 # Used by ingest-space to authenticate against Kratos and call the private
-# GraphQL API to fetch space contents. Must be a valid platform-admin Kratos
+# GraphQL API to fetch space contents. Must be a valid spaces-reader Kratos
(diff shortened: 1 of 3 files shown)


.env.docker
```diff
@@ -165,6 +165,10 @@ VECTOR_DB_CREDENTIALS=username:password
 VECTOR_DB_CREDENTIALS=root:toor
 
 
+# Used by ingest-space to authenticate against Kratos and call the private
```
techsmyth, line 168: Should be SPACES_READER role, not platform admin: 

Image

PR #6030, opened 2026-04-29
fix: template content space settings now editable

hero101:
## Summary
- `updateTemplateContentSpace` mutation silently ignored the `settings` field — only `about` was being saved. Wire `TemplateContentSpaceService.update` through `SpaceSettingsService.updateSettings` so settings actually persist.
- Made all fields on `UpdateSpaceSettingsCollaborationInput` optional and changed the merge in `SpaceSettingsService.updateSettings` to apply only the fields the caller explicitly set. Sending a subset (e.g. just `allowGuestContributions: true`) no longer wipes out the other flags.
- Added unit tests covering the partial-merge contract (subset, explicit `false`, explicit `undefined`) and the template-content-space settings application path.

## Test plan
- [x] `pnpm exec tsc --noEmit` → exit 0
- [x] `pnpm exec biome check src/domain/space/space.settings src/domain/template/template-content-space` → exit 0
- [x] `pnpm exec vitest run src/domain/space/space.settings/space.settings.service.spec.ts src/domain/template/template-content-space/template.content.space.service.spec.ts` → 38/38 pass
- [x] Pre-commit full suite → 6382 pass
- [x] Red-green check on new tests: reverting either fix turns the new tests red; restoring brings them back green

## Schema
Non-breaking input change: 6 fields on `UpdateSpaceSettingsCollaborationInput` flip from `Boolean!` → `Boolean`. Existing clients sending all fields keep working.

## Summary by CodeRabbit

* **New Features**
  * Collaboration settings now accept partial updates: individual options are optional so you can update specific choices without sending all values.
  * Omitting a setting preserves its current value; explicitly setting a value (including false) overwrites prior configuration.
  * Space settings can be updated together with other space properties in a single operation.

* **Tests**
  * Expanded tests to verify partial-update/merge behaviors and combined settings updates.

### DIFF
--- a/src/domain/space/space.settings/space.settings.service.ts
+++ b/src/domain/space/space.settings/space.settings.service.ts
@@ -48,7 +48,10 @@ export class SpaceSettingsService {
         ([, value]) => value !== undefined
       );
       const definedFields = Object.fromEntries(definedEntries);
-      Object.assign(settings.collaboration, definedFields);
+      settings.collaboration = {
+        ...(settings.collaboration ?? {}),
+        ...definedFields,
+      };
     }
     if (updateData.sortMode) {
       settings.sortMode = updateData.sortMode;
--- a/src/domain/space/space.settings/space.settings.service.ts
+++ b/src/domain/space/space.settings/space.settings.service.ts
@@ -45,7 +45,7 @@ export class SpaceSettingsService {
       const collaborationUpdates = updateData.collaboration;
       const allEntries = Object.entries(collaborationUpdates);
       const definedEntries = allEntries.filter(
-        ([, value]) => value !== undefined
+        ([, value]) => value !== undefined && value !== null
       );
       const definedFields = Object.fromEntries(definedEntries);
       settings.collaboration = {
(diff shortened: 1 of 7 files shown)


src/domain/space/space.settings/space.settings.service.ts
```diff
@@ -41,7 +41,17 @@ export class SpaceSettingsService {
       settings.membership = updateData.membership;
     }
     if (updateData.collaboration) {
-      settings.collaboration = updateData.collaboration;
+      // Apply only the fields the caller explicitly set, preserving the rest.
+      const collaborationUpdates = updateData.collaboration;
+      const allEntries = Object.entries(collaborationUpdates);
+      const definedEntries = allEntries.filter(
+        ([, value]) => value !== undefined
```
ccanos: It should never happen, but it could be theoretically possible to pass `null` now to those boolean values that would pass this validation and end up in the db.
replacing this with `([, value]) => value !== undefined && value !== null` or check type === 'boolean' would prevent that.

ccanos (changes requested):
Very good, tested and works well, I was going to approve it but claude mentioned something that I think is worth changing, just one line and ready to go.

PR #6052, opened 2026-05-07
Backfills Office Documents entitlement

valentinyanakiev:

This pull request introduces a database migration to backfill the "Office Documents" entitlement for existing licenses. It solves a bug where licenses created before the introduction of this entitlement (originally shipped with PR #5967) did not automatically receive it.

The migration ensures that all existing `space`, `collaboration`, and `template_content_space` licenses have the `space-flag-office-documents` entitlement. The `enabled` state for the entitlement is set according to the default policy at the time of its introduction: `true` for `space` and `template_content_space` licenses, and `false` for `collaboration` licenses.

### Additional context

The migration uses self-contained constants to prevent future changes to global enums from affecting the historical accuracy of this migration. The `down` method safely removes all instances of this specific entitlement type.


antst (approved):
LGTM

PR #6053, opened 2026-05-07
Read image dimensions from file-service-go response (v0.0.17)

antst:
## Summary

Companion to [alkem-io/file-service-go#19](https://github.com/alkem-io/file-service-go/pull/19) (image canonicalization + post-rotation pixel dimensions, shipped as v0.0.17). Closes the regression where phone-photo avatars/banners failed dimensional validation with confusing `1082 not in 190-410` errors.

## The bug

Pre file-service migration, the server's `visual.service.ts:uploadImageOnVisual` used a sharp pipeline that auto-oriented images before measuring dimensions. Once image processing moved to file-service-go we kept the validation but switched to `image-size` on the raw upload buffer — which **does not apply EXIF rotation**. A portrait phone shot stored as `1082×127` with `Orientation=6` got measured as 1082-wide instead of 127-wide and was rejected as too wide for an AVATAR (190–410).

## The fix

file-service-go is now the single source of truth for canonical pixel dimensions. It already decoded the bytes as part of compression/auto-rotate; v0.0.17 just surfaces the resulting dims on every metadata-returning response (`POST /internal/file`, `POST /internal/file/copy`, `PATCH /internal/file/{id}`, plus the new content-replace endpoint). The server reads them off the response and validates against per-visual constraints. No local decode.

## Changes

- **Migration `1778198400000-AddContentMetadataToFile`**: `ALTER TABLE file ADD COLUMN content_metadata JSONB NOT NULL DEFAULT '{}'::jsonb`. JSONB chosen over top-level columns to forward-fit future content-type metadata (video duration, PDF page count, etc.) without further DDL. file-service-go writes this column at create time and lazy-backfills empty rows on first read.
- **`quickstart-services.yml`**: bump `alkemio/file-service-go` to v0.0.17.
- **file-service-adapter DTOs**: `CreateDocumentResult` (covers POST `/internal/file` AND POST `/internal/file/copy`) and `UpdateDocumentResult` gain optional `imageWidth?` / `imageHeight?`.
- **`IDocument` interface**: gains transient (non-TypeORM-mapped) optional `imageWidth?` / `imageHeight?`. Populated by `uploadFileAsDocumentFromBuffer` from the file-service-go response on upload-result paths; undefined for entities loaded via the standard repo path. Not `@Field`-decorated — server-internal, not exposed in GraphQL.
- **`storage.bucket.service.persistDocumentWithPreparedAuth`**: attaches `result.imageWidth` / `imageHeight` to the returned `IDocument`.
- **`visual.service.uploadImageOnVisual`**: drops the local `getImageDimensions(buffer)` call. Upload happens FIRST, validation reads dims from the returned `IDocument`. On dim-mismatch: delete the just-uploaded doc + throw, mirroring the rollback pattern used elsewhere for post-upload validation failures.
- **New `validateImageDimensionsOnVisual` helper** centralizes the read-and-validate flow.
- **Drop dead code**: `image.util.ts` (`getImageDimensions` wrapper) + spec, `index.ts` re-export, and the `image-size` package dep. Only caller was `visual.service.ts` before this PR.

## Deploy ordering

The migration must run **before** v0.0.17 deploys, otherwise sqlc queries on file-service-go fail. Migration runner runs ahead of service restart in our pipeline — handled automatically.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm test:ci:no:coverage` — all 6431 tests green
- [ ] Local manual verification: upload an EXIF-rotated phone photo (Orientation=6, e.g. 1082×127 raw) to an AVATAR visual; previously rejected, now accepted as canonicalized 127×1082.
- [ ] Deployed regression check against staging once both this PR and file-service-go v0.0.17 are running.

## Summary by CodeRabbit

* **Chores**
  * Upgraded file-service container image to v0.0.17 for improved image handling
  * Added a database migration to store file content metadata
  * Removed a local image-size dependency and shifted image dimension processing to the file-service

* **New Features**
  * Image uploads now surface width/height metadata post-upload and are validated against visual constraints; failed validations trigger best-effort cleanup and error reporting


antst:
@coderabbit review

antst:
@coderabbit declining the `imageDimensions?: { width, height }` grouping suggestion on `src/domain/storage/document/document.interface.ts:52-53`.

**Reasons:**

1. **file-service-go's contract guarantees atomicity at the source.** Both dims come from a single `vips` decode in `Process` — there's no code path on the Go side that produces width without height. The "half-populated" scenario the suggestion guards against would require a file-service-go regression that also bypassed their test suite.
2. **Wire format is flat.** `CreateDocumentResult` from file-service-go has `imageWidth` / `imageHeight` as flat fields (per the OpenAPI contract). Wrapping into a nested object on the server side adds a translation layer at the boundary that doesn't add safety the contract isn't already providing.
3. **Consumer site already enforces atomicity at runtime.** `validateImageDimensionsOnVisual` checks `if (imageWidth === undefined || imageHeight === undefined) throw` before either is used — so a hypothetical partial state would surface as a clear error, not a misleading width-only validation.
4. **Idiomatic consistency.** `IDocument` uses lots of independently optional fields (`createdBy?`, several relations, etc.). Adding a nested-object field for one property would be a one-off pattern in the interface.

The two flat optional fields are the natural shape for this "decorate-result-with-runtime-fields" pattern. Keeping them.

PR #6145, opened 2026-06-11
WIP: spec for actor collection callout

techsmyth:
## Summary
Exploring the requirements around having a callout for displaying a set of actors, with different display options. 

## Schema Contract Checklist (Feature 002)

If this PR affects the GraphQL schema, complete ALL items below:

- [ ] Ran `npm run schema:print` (and optionally `npm run schema:sort`) to regenerate `schema.graphql`.
- [ ] Retrieved base snapshot (e.g. `git show origin/develop:schema.graphql > tmp/prev.schema.graphql`).
- [ ] Ran `npm run schema:diff` to produce `change-report.json` and `deprecations.json`.
- [ ] Reviewed classifications; only expected changes present.
- [ ] (Optional) Ran `npm run schema:validate` and artifacts passed validation.
- [ ] Committed ONLY `schema.graphql` (not the JSON artifacts).

### Change Report Summary

Paste the key counts from `change-report.json` (example format below) — remove any zero lines if desired:

```
Additive: X
Deprecated: Y
Deprecation Grace: Z
Breaking: B (override applied? yes/no)
Premature Removals: P
Invalid Deprecations: I
Info: N
```

### Deprecations Added / Updated

List new or updated deprecations with schedules (`REMOVE_AFTER=YYYY-MM-DD | reason`). Indicate any in grace period.

### Breaking Changes (If Any)

If BREAKING changes are intentional:

- Rationale:
- Risk assessment / mitigation:
- CODEOWNER approval with phrase `BREAKING-APPROVED` requested? (link)

### Other Notes

---

Reference docs: `specs/002-schema-contract-diffing/quickstart.md` for full workflow and troubleshooting.


SimoneZaza (commented):
Very nice to get this thing going! A couple of remarks:
- User story 3 (allowing the admin to determine whether there are filter options or not) doesn't sound very important to me. Instead, I think it's important for the admin to be able to choose which contributors will be shown: all / users / organisations / virtual contributors. Once we have the custom roles, it would be nice to expand this set with roles as well, might be good to take that in from the start, as it will influence the UI choices. 
- I understand that we want to keep it simple for now, but it might be good to already think of the way we want admins to create these callouts. I think this should just be added to the standard post create flow, but the options when creating a post are becoming quite heavy. So we might want to expand the Space setting "Allow members to create posts" with more options than a simple yes or no. But that's of course for a different story.
- Just to be sure we're all on the same page: I understand that making it a collection of Actors works nice in the back, but for users these are two completely different things, and should be treated as different options as well.

techsmyth:
@SimoneZaza thanks for the feedback, gave the spec a good clean up + also took in some additional insights - please give the latest another scan! Can chat through on Monday...

techsmyth:
@SimoneZaza thanks for discussion on this one earlier, made the key changes discussed - think next step is a wider review...

bobbykolev:
I'm closing this PR as big part of the spec is already implemented and released. 
In case of missing refinements we can just create follow up stories.

PR #6196, opened 2026-06-25
feat(009): index + search Collabora office document content

hero101:
Part of alkem-io/client-web#9881

Part of **`workspace#009-office-doc-search`** — Office Document Content Search (rollout step 2 of 3; **LEAD** slice).

## What
Indexes and searches the text content of Collabora office documents (docx/xlsx/pptx) in global search, as a new `COLLABORA_DOCUMENT` search result type.

- **Extraction** — `CollaboraTextExtractService`: fetches bytes via `FileServiceAdapter` (no Collabora), parses in-process with `officeparser@7.2.1`, cleans markup, returns `null` on no-text/parse-error/fetch-error/over-cap. Source-file-size cap is a **dedicated config key** `search.collabora_document_max_source_size` (default 25 MiB), enforced twice: pre-fetch against the DB `size` column and post-fetch against the actual byte count. Cleaned text is **truncated to 1,000,000 chars** (FR-016) before indexing.
- **Ingest** — lazy, on owning-entity reindex; loads docs from both `CalloutFraming` and `CalloutContribution`, extracts **sequentially** (one document in memory at a time), writes `content` + `flowStateID` to the new **`office-document`** index, then force-merges it to one segment. Bulk requests are **byte-chunked** (10 MB sub-requests) and a failed batch is recorded without aborting the reindex. No-text docs still indexed (findable by `displayName`).
- **Result type** — new `ISearchResultCollaboraDocument` DTO + enum value, split into framing (`isContribution=false`) / contribution (`isContribution=true`) results. Query-time auth filter (no index-baked permissions). No excerpt, no match-source field. `profile.url` resolves to the parent callout URL (new `COLLABORA_DOCUMENT` case in the URL generator).
- **Query** — reworked relevance (affects all result types, deliberate): `profile.displayName.text^3` and `profile.tags^2` field boosts, typo-tolerant fuzzy matching scoped to `content` only, and a phrase-proximity ranking boost on `content`. Matched terms are reported back via ES highlights (`ISearchResult.terms`, server#3702).
- **Observability** — reason-tagged Winston log + APM metric + reindex `Task` record on every skip/failure (`no-text` | `parse-error` | `fetch-error` | `over-cap`).
- **Schema** — additive (`schema.graphql` regenerated; no BREAKING).

## Tests
Extract-service unit + result-type/auth integration + query-builder/highlight specs + URL-generator branch specs. Lint + full unit suite green.

## Not in this PR
- T022 (operational backfill via `adminSearchIngestFromScratch`) and T024 (e2e quickstart) — need a booted stack (ES+DB+file-service).
- Quoted-phrase (`"exact phrase"`) query semantics — quotes are currently ignored (bag-of-words, OR). Needs a product decision + `match_phrase` support; candidate follow-up story.

## Cross-repo
Index alias matches the reporting-orchestration template (`alkemio-data-*office-document-*`); apply that template first. Client-web renders the new type once this schema merges.

Spec: `agents-hq/specs/009-office-doc-search/` · tasks: `tasks/server.md`

## Summary by CodeRabbit

* **New Features**
  * Collabora documents are now indexed and returned as a new search result type, including dedicated URL generation.
  * Search results now display relevance terms derived from highlighted matches.
* **Bug Fixes**
  * Improved search relevance with boosted profile matching, typo-tolerant content matching, and phrase boosting.
  * More robust office-document indexing: better handling of oversized/unsupported/empty documents, with consistent truncation and skipping.
* **Chores**
  * Added `search.collabora_document_max_source_size` and reduced the default `storage.file.max_file_size` to 15 MiB.


polibb:
Critical — breaks the feature
- Empty profile.url for every Collabora document result (CONFIRMED). The URL generator's switch has no COLLABORA_DOCUMENT case and falls through to a bare return ''. The client passes this straight into href. This breaks the PR's core acceptance criterion ("one-action open result").  Fix: add the missing case, mirroring the existing whiteboard case.

manual testing:
`[url-generator] Using cached url for entity: http://localhost:3000/adminspacepublic`
clicking on a search result actually just closes the dialog in /home and leaves user at /home

High — silent correctness/data-loss bugs
- Tag-boost targets a field nothing writes (CONFIRMED). Query boosts profile.tagsets.tags^2, but ingestion only ever writes profile.tags — the boost is a permanent no-op, and the PR's own test asserts the wrong field name, so it won't catch this.
- profile.displayName^3 boost only fires on exact-match keyword field — word-level title matches route through .text at weight 1 instead, so the "title matches rank higher" claim isn't actually true for partial matches.
- Unbounded concurrent extraction (Promise.all per batch) — up to 25 MiB of documents extracted concurrently during a full reindex, despite the code's own comment claiming sequential awaiting. Corroborated independently by 3 agents. The existing asyncMapSequential helper is the intended tool here — use it.
- Task-record clobbering — updateTaskResults is an unlocked Redis GET→mutate→SET, called concurrently; concurrent skip-records get lost to last-write-wins. Note: even serializing the callers isn't a full fix since fetchAndIngest/ingestBulk fire un-awaited calls that can still interleave — flag that the durable fix is an atomic append or per-task queue, not just a mutex.
- One oversized bulk call aborts the entire reindex task — because office-document is ingested last, a single failed bulk call means the alias flip never runs and all 9 other freshly-rebuilt indices get silently discarded.

Medium — spec/description mismatches (worth calling out explicitly since you asked for this)
- PR body claims "no query change" — actually a full query-builder rewrite.
- PR body claims reuse of STORAGE_MAX_FILE_SIZE — actually a separate, undocumented 25 MiB cap.
- Size cap only guards the compressed file size — zip-bomb inflation is unbounded after decompression.
- Cap trusts the DB file.size column rather than actual fetched byte count (TOCTOU-style trust issue).
- Ingest batch size raised 30→50 for posts/whiteboards/memos with no payload chunking — undocumented, lower severity than the office-doc issue but same family.
- An entire unspecced <em>/matched-terms highlighting subsystem shipped with no corresponding spec entry.
- Dead isContribution field written to the index that nothing reads.

Architecture/SOLID

- Adding a new result type requires coordinated edits across ≥6 separate per-type lists in the search layer, with variadic call sites — this compiles fine even if you miss one, causing silent result loss for that type. Worth a comment recommending a single registry/discriminated-union pattern instead of parallel lists.
- Fuzzy clause widens matching for existing whiteboard results as a side effect of the rewrite (unintended ranking change for content unrelated to this PR's stated scope).
- Exact content hits now score up to 3x — a global ranking change "smuggled in" alongside the new result types, not mentioned in the PR description.

polibb:
Any reason I can get a maximum of 4 results?

image

image

image

---

I got more than 4 only once, when posts were returned as well:

image

but it's listing separately the posts from the documents (and not all of them at that, even though all of them are uploaded the same way - import in a new post, taking the document title as post title), and filtering for posts doesn't exist, so they're just visible on "All" not on "Documents"

image

the post link takes you to the post in the space though, unlike the link of a document result

P.S. Q: we have relevance ordering ('resilience' appears 18x in D10, once each in D01/P10/P13/X10) ?

---

image

phrase search is not honored. Your result (doc-07, xls-11, xls-03, ppt-01) is exactly the OR-union of the individual terms: quantum {doc-07, ppt-01, xls-03} ∪ meadow {doc-07, xls-11}. The quotes were ignored and the query ran as bag-of-words with a default OR operator (an AND would have returned only doc-07). If exact-phrase queries are a requirement, you'll need match_phrase/positional support — right now quoting does nothing

do we want that?

---

🌈 Consolidated model:

Analysis chain: lowercase yes, unicode/Cyrillic yes, diacritic-folding no, stemming no, asciifolding no.
Query behavior: fuzzy AUTO (1 edit for 3–5 chars, 2 for ≥6), Damerau transpositions on, prefix_length ≈ 2 (first two chars protected), default operator OR, no phrase/match_phrase support.
Extraction coverage: body text, table cells, speaker notes, all worksheets — but not doc footers/headers or embedded chart text.

hero101:
@polibb Thorough review — thank you. Went through every item; here's the disposition. Code fixes are staged locally and land in the next push.

## Fixed (next push)

- **Empty `profile.url` (Critical)** — confirmed: `generateUrlForProfileNotCached` had no `COLLABORA_DOCUMENT` case and fell through to `''`. Added the case: the document's profile URL resolves to the **parent callout URL** (framing → the callout page renders the document; contribution → the callout page lists it). A true deep-link (`…/documents/<id>` route) needs a client route + url-resolver segment — a document has no `nameID` — so callout-level linking is the correct server-side fix for now. Covered by three new URL-generator specs.
- **Tag boost targeted a non-existent field** — confirmed: ingestion writes `profile.tags` (joined string), never `profile.tagsets.tags`. Boost changed to `profile.tags^2`; the test asserting the wrong field is fixed too.
- **`displayName^3` fired only on whole-string keyword match** — confirmed against the `alkemio-data-base-fields` mapping (`displayName` is `keyword` with a `.text` subfield). Boost now targets `profile.displayName.text^3`, so word-level title matches actually rank higher.
- **Unbounded concurrent extraction** — confirmed: `asyncMap` is `Promise.all`. Switched to `asyncMapSequential` (one document in memory at a time), comment corrected.
- **Oversized bulk aborting the whole reindex** — two-part fix: bulk payloads are now byte-chunked (10 MB sub-requests, single oversized doc isolated in its own request), and a thrown batch failure is recorded as an errored batch instead of propagating — so the alias flip and the other 9 indices survive.
- **Cap trusts the DB size column (TOCTOU)** — added a post-fetch re-check against the actual fetched byte count (skip = `over-cap`, with both numbers logged).
- **Task-record interleaving from this PR** — the un-awaited `updateTaskResults` in `fetchAndIngest` is now awaited, and sequential extraction serializes the per-skip records. Agreed the durable fix is an atomic append in `TaskService` (GET→mutate→SET on a shared cache entry is pre-existing infra used by all callers) — that's a separate story; happy to file it.
- **PR body mismatches** — body rewritten to match the code: dedicated 25 MiB cap (not `STORAGE_MAX_FILE_SIZE`), truncation at 1M chars, the query rework (boosts + fuzzy + phrase-proximity + highlight matched-terms), `office-document` index name, and the ranking-change disclosure.

## Not changing — reasoning

- **Zip-bomb inflation** — real residual risk: `officeparser` inflates in memory with no inflation cap, so only the compressed size is guardable pre-parse. Sequential extraction now bounds the blast radius to one document at a time inside an offline ingest task (not the request path), and the 25 MiB compressed cap bounds typical inflation. A hard fix means a streaming/inflation-capped unzip layer or an external extraction service — out of scope here; can be a follow-up if we want it.
- **Dead `isContribution` in the index** — deliberate and documented in the spec's `data-model.md`: "write-only in ES … kept in the mapping for future ES-side category filtering (or removal; follow-up in tasks/server.md Phase 6)". The result service intentionally recomputes it from the DB at query time (authoritative over a possibly-stale index).
- **Per-type parallel lists (registry/discriminated-union)** — agree it's a real maintainability hazard (6+ coordinated edit sites, silent result loss if one is missed), but restructuring the search layer is well beyond this PR. Worth its own refactor story.
- **Fuzzy widening whiteboards / exact content hits scoring higher** — intentional, kept: fuzzy is deliberately scoped to `content` only (per-field Levenshtein rewrite cost), and phrase-proximity is a rescoring boost, not a matcher — it never changes which documents return, only their order. It applying to whiteboards too is consistent rather than accidental. Now disclosed in the PR body instead of smuggled.
- **Batch size 30→50 for posts/whiteboards/memos** — kept: it's row-count batching for throughput, and the new byte-chunking makes payload size independent of row count, which was the actual risk.

## Answers to the testing questions

- **"Max 4 results"** — client-side: `CrdSearchOverlay.tsx` hardcodes `SEARCH_RESULTS_COUNT = 4` per category page (server default budget is 40 split across categories). More results come via cursor pagination ("load more"), not a bigger first page. If 4 feels low for the Documents tab, that's a one-line client change.
- **Posts listed separately from documents** — by design (ADR 0002): an *imported* document creates a Post whose title matches, so the *post* entity matches by title (a POST result) while the *document* matches by content (a COLLABORA_DOCUMENT result). They're different entities in different indexes. The Documents tab filter listing only document types (no posts) is the client's filter config.
- **Relevance ordering** — yes: BM25 term frequency (18× 'resilience' in D10 outscores 1× elsewhere, modulo length normalization), plus the phrase-proximity boost when terms appear close together.
- **Phrase search** — your analysis is exactly right: quotes are ignored, terms run as bag-of-words with OR. Supporting `"exact phrase"` needs quote-aware parsing + a `match_phrase` clause — product decision; I've listed it under "Not in this PR" as a candidate follow-up story.
- **Consolidated model** — accurate on all counts. One addition: doc headers/footers and embedded chart text aren't surfaced by officeparser's AST, hence not extracted; sheet names and speaker notes are recovered explicitly (its `toText()` drops them).

hero101:
009 Office Document Content Search — test summary

  Manual test plan executed against a local stack via the search GraphQL API, with a member and a non-member session.

  What was tested
  - Content search across Office formats — Word body text, Excel cell values and sheet-tab names, PowerPoint slide text and speaker notes
  - Access control — non-member gets zero results for a private-space document; member gets it (permission enforced at query time)
  - Relevance ranking — title/frequency boosts; found & fixed a tag-boost defect (below)
  - Flow-state-scoped search — a document in flow state A is correctly hidden when searching state B
  - Language & analysis contract — Cyrillic lowercase folding, no diacritic folding, no compound-word splitting; verified in Bulgarian and Dutch
  - Typo tolerance — fuzzy matching with a protected 2-char prefix
  - Pagination — 4 results per category + load-more cursor, no overlap
  - Phrase/quote handling — quotes ignored (OR-union of terms), by current contract
  - Freshness — a deleted document leaves the index after reindex
  - Empty/image-only doc — found by name, never by content; no-text skip recorded
  - No content leakage — results carry no document text excerpt

  Result: 16/17 machine-verifiable cases pass; 1 accepted by-design (see below).

  Bug found & fixed (and pushed): tag-only documents ranked dead-last because content matches were double-counted in the search query. Fixed (content^0 so
  content scores once). Forcing a tag strictly above content was intentionally dropped — it's not stable across reindexes (multi-shard BM25); tag ≈ 
  content parity accepted.

  Fixture fix: a Bulgarian test document was missing a sentence (a search term was untestable); regenerated it and authored the missing Dutch fixture.

polibb (approved):

PR #6218, opened 2026-07-02
fix(mcp): create_whiteboard_in_space accepts a top-level space nameID, not only a UUID

valentinyanakiev:
## Problem

Live on acceptance (2026-07-02): a user asked the assistant to create a whiteboard in their space, approved the confirmation, and got

> I couldn't complete that action — Space not found: zimbabve. I haven't made any changes.

`zimbabve` is the space's **real nameID** (and displayName — typo is in the data). The model passed it because that's the only space identifier it can obtain:

- `list_whiteboards` declares `context.spaceId` in its result shape but **never populates it**; the items carry the whiteboard URL, whose path segment is the space *slug*.
- The tool's own description says *"resolve the space id from a name via search_content / list_whiteboards"* — yet `execute()` resolved via `getSpaceOrFail`, which is **UUID-only**.

So the happy path documented in the tool description was impossible with a slug, and the failure surfaced only *after* user approval.

## Fix

- `uuidValidate(spaceId)` → `getSpaceOrFail` (any level, unchanged behavior for UUIDs).
- Otherwise → `getSpaceByNameIdOrFail` (top-level space nameID — unambiguous; subspace nameIDs are not globally unique, so subspaces still require the UUID).
- Input schema + description now state exactly what's accepted; the not-found error tells the model what to pass instead of dead-ending.

Authorization is untouched — resolution only finds the CalloutsSet; `createCalloutOnCalloutsSet` still enforces `CREATE_CALLOUT` on the same path as GraphQL.

## Testing

- New spec: UUID → `getSpaceOrFail`; slug → `getSpaceByNameIdOrFail`; unresolvable → guidance error, no mutation call.
- MCP suite: 18 files / **149 passed**; biome + tsc clean.

## Review pass (applied in-PR)

A multi-agent review of this PR surfaced and fixed here:
- **Shared resolver**: the id-or-nameID policy now lives in `SpaceLookupService.getSpaceByIdOrNameIdOrFail` (using `isUUID` from class-validator, the codebase convention) instead of an inline branch in one tool.
- **`analyze_contributions` had the same bug, silently**: a slug in `scope: "space:{id}"` was matched raw against `space.id` and returned an *empty result set with no error*. It now resolves through the shared method and errors with guidance.
- **Hardened `getSpaceByNameIdOrFail` / `getSubspaceByNameIdInLevelZeroSpace`**: `...options` was spread after `where`, letting a caller-supplied `where` clobber the nameID/level guard — now merged safely (same order as `getSpace`).

## Follow-up (not this PR)

`list_whiteboards` should populate the declared `context.spaceId` so the model can use UUIDs directly — tracked as part of the assistant hardening epics (alkem-io/alkemio#1948–#1952 scope area).

## Summary by CodeRabbit

* **New Features**
  * Space references now accept either a UUID or a top-level space slug in more places.
  * Whiteboard creation and contribution analysis now provide clearer guidance on how to enter space identifiers.

* **Bug Fixes**
  * Improved space lookup safety so required filters can’t be overwritten by extra query options.
  * Fixed space-not-found handling to return a consistent, clearer error message.
  * Subspace lookups now preserve the correct level and parent-space constraints.

* **Tests**
  * Added coverage for UUID vs slug handling and lookup guard behavior.

### DIFF
--- a/src/domain/space/space.lookup/space.lookup.service.spec.ts
+++ b/src/domain/space/space.lookup/space.lookup.service.spec.ts
@@ -107,6 +107,56 @@ describe('SpaceLookupService', () => {
     });
   });
 
+  describe('getSpaceByIdOrNameIdOrFail', () => {
+    const SPACE_UUID = 'e0e2e08f-04cc-40dc-b40c-089b6cc27689';
+
+    it('routes a UUID to getSpaceOrFail (any level)', async () => {
+      const mockSpace = { id: SPACE_UUID } as ISpace;
+      entityManager.findOne.mockResolvedValue(mockSpace);
+      const byNameId = vi.spyOn(spaceRepository, 'findOne');
+
+      const result = await service.getSpaceByIdOrNameIdOrFail(SPACE_UUID);
+
+      expect(result).toBe(mockSpace);
+      expect(byNameId).not.toHaveBeenCalled();
+    });
+
+    it('routes a non-UUID to getSpaceByNameIdOrFail (top-level nameID)', async () => {
+      const mockSpace = { id: SPACE_UUID, nameID: 'my-space' } as ISpace;
+      const byNameId = vi
+        .spyOn(spaceRepository, 'findOne')
+        .mockResolvedValue(mockSpace as Space);
+
+      const result = await service.getSpaceByIdOrNameIdOrFail('my-space');
+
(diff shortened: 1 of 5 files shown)


valentinyanakiev:
@coderabbitai review

valentinyanakiev:
@coderabbit review

PR #6207, opened 2026-06-29
feat(auth-reset): config-driven queue + dedicated worker role

hero101:
## Summary

Splits the `alkemio-auth-reset` RabbitMQ consumer out of the main server into a dedicated, config-driven worker role so it can scale independently (KEDA scale-to-zero).

- **Config is the single source of truth.** The auth-reset queue name and worker flag live in `alkemio.yml` (`microservices.rabbitmq.auth_reset`), threaded through all three call sites that name the queue: the `main.ts` bind, the `AUTH_RESET_SERVICE` publisher proxy, and the controller retry-republish.
- **Main server publishes, no longer consumes.** Normal deployments emit reset events but do not bind the queue.
- **Dedicated worker** (`worker=true`) binds ONLY the auth-reset queue, with its own resources for a KEDA scale-to-zero deployment.
- **Bootstrap split** into `main.server.ts` / `main.worker.ts` entrypoints; `MicroservicesModule` decomposed into role-scoped modules (`auth-reset.worker.module.ts`, `worker.event-bus.module.ts`, `communication.rpc.module.ts`).
- Added `nest enableShutdownHooks` for clean worker shutdown.

## Changes

- `clientProxyFactory` / `connectMicroservice` widened to accept a string queue name.
- New config keys in `alkemio.yml` + `src/types/alkemio.config.ts`.
- Specs updated for `subscription.factory` and `auth-reset.controller`.

## Test plan

- [ ] Main server starts, publishes reset events, does NOT consume `alkemio-auth-reset`.
- [ ] Worker (`worker=true`) binds only the auth-reset queue and processes events.
- [ ] `pnpm lint` + `pnpm test:ci:no:coverage` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated worker mode for authorization/license reset processing (headless consumer without HTTP).
  * Introduced a configuration-driven RabbitMQ queue for auth-reset so the server and worker stay aligned.

* **Bug Fixes**
  * Improved reset retry publishing to always target the configured auth-reset queue.
  * Ensured the worker-only flow doesn’t start competing RabbitMQ consumers.

* **Documentation**
  * Expanded startup/config guidance to clarify server vs. worker behavior and queue isolation.

---
Part of #6205

PR #6220, opened 2026-07-02
feat(auth-reset): local docker-compose autoscaler for the reset worker

hero101:
## Why

The auth-reset **worker** role (from #6207 / #6217) had no way to run locally. This adds it to the dev stack and autoscales it like prod (KEDA scale-to-zero) — so resets actually get drained on your machine, with zero manual steps.

## How

`quickstart-services.yml` gains two services:
- **`auth-reset-worker`** — the Alkemio image in worker mode. Starts at **0** replicas (competing consumers).
- **`auth-reset-scaler`** — a tiny sidecar that watches the `alkemio-auth-reset` queue and scales the worker `0↔N` via the docker socket. Local stand-in for KEDA.

## How to use

Nothing new to run — it starts with the stack:

```bash
pnpm run start:services      # brings up deps + the scaler (worker at 0)
pnpm start:dev               # host server = the publisher
# trigger authorizationPolicyResetAll -> workers spin up, drain, scale back to 0
```

Tune it per-developer without committing: copy `.env.docker.local.example` → `.env.docker.local` (gitignored) and set any of:

| Env | Default | Meaning |
|-----|---------|---------|
| `AUTH_RESET_MAX_REPLICAS` | `3` | max workers when the queue has resets |
| `AUTH_RESET_POLL_INTERVAL` | `5` | seconds between queue checks |
| `AUTH_RESET_IDLE_ROUNDS` | `3` | empty polls before scaling back to 0 |

Both `start:services` and the scaler load `.env.docker.local` on top of `.env.docker`, so your values win.

## Scope

Local dev tooling only (`quickstart-services.yml`, `package.json`, two new files). No app/source changes.

## Summary by CodeRabbit

* **New Features**
  * Added a local autoscaling setup for the auth-reset worker, including a companion service that scales worker replicas based on queue depth.
  * Added optional per-developer Docker environment overrides for customizing scaling behavior and compose project settings.
  * Updated the service start flow to automatically load the local override file when present.
* **Bug Fixes**
  * Improved worker handling so it drains gracefully before scaling down and avoids reducing replicas while work is still in progress.

### DIFF
--- a/.env.docker.local.example
+++ b/.env.docker.local.example
@@ -1,0 +1,27 @@
+# Per-developer LOCAL overrides for the docker stack.
+#
+# Copy this file to `.env.docker.local` (which is gitignored — see the
+# `.env.docker.*` / `!.env.docker.*.example` rules in .gitignore) and put any
+# machine-specific overrides here. `pnpm run start:services` loads it AFTER
+# `.env.docker`, so values here win. Nothing in this file is committed.
+#
+#   cp .env.docker.local.example .env.docker.local
+#
+# ---------------------------------------------------------------------------
+# Auth-reset autoscaler knobs (consumed by docker compose at parse time as
+# ${AUTH_RESET_*:-default} and injected into the auth-reset-scaler sidecar).
+# Uncomment + edit to override the defaults shown.
+# ---------------------------------------------------------------------------
+
+# Max worker replicas the scaler spins up when resets are queued (default 2).
+#AUTH_RESET_MAX_REPLICAS=3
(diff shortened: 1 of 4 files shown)


valentinyanakiev (changes requested):
Doesn't need to go to release/65, rebase to develop

hero101:
@valentinyanakiev it's based on the code on the release branch. Can't go without it. People are experiencing issues resetting locally on the release branch. It can be hold off until the release is merged

PR #6259, opened 2026-07-12
ci(test-deploy): serialise Hetzner-test deploys against the nightly

valentinyanakiev:
## Why

A `workflow_dispatch` of **Build, Migrate & Deploy to Test on Hetzner** that lands *during* a test-suites nightly rolls the single-replica server pod and runs DB migrations while e2e tests are in flight. On **2026-07-12** a `develop` deploy at **~00:31 UTC** did exactly this mid-run (`deployTimestamp` on the deployment; new pod `Available` 00:37) — the `non-interactive-login` endpoint 404'd and the run cascaded to failure across many files.

This was the real cause behind the nightly subspace/entitlement failures — **not** the long-assumed "~5s transit-path idle timeout." A runner-side packet capture (tcpdump on the Scaleway runner during a full run) proved connections to the test node live from <1s up to **608s** with **zero mid-stream resets**, refuting any 5s network cut (see test-suites#563, server#6258).

## What

- **`.github/scripts/test-cluster-lock.sh`** — a mutual-exclusion lock backed by an atomic `kubectl create configmap` in the test cluster. The cluster is the single lock authority, so it's race-free with no new infrastructure. Stale holders (crashed runs) are auto-reclaimed after 60 min.
- **`build-deploy-k8s-test-hetzner.yml`** — new `acquire-lock` job (after `build`) → `migrate` → `deploy` → `release-lock` (`always()`). The docker build stays unlocked; only migrate+deploy hold the lock.

The paired change in **test-suites** (`feat/test-deploy-lock`) has the nightly acquire the same lock **after** its own `deploy-all-services` and hold it across `cleanup-db` + `e2e`. Because the nightly takes the lock only after its own triggered deploys finish, those run while the lock is free — **no deadlock**. An external deploy during a nightly now waits it out.

## Testing

- Lock script unit-exercised against a stubbed `kubectl`: atomic acquire, contended-wait+timeout, wrong-holder release is a no-op, correct-holder release, and stale-TTL reclaim all behave.
- YAML + job-graph validated: `build → acquire-lock → migrate → deploy`, `release-lock` `always()` after all three.

> Pre-commit hook skipped: it runs a full `tsc` that fails on pre-existing, unrelated OIDC test errors (`ioredis` not installed locally). This change is `.github/`-only and cannot affect TypeScript.

Refs: test-suites#563, server#6258. Pairs with test-suites `feat/test-deploy-lock`.

## Summary by CodeRabbit

* **Chores**
  * Added an automated mutual-exclusion lock for shared Kubernetes test-cluster access.
  * Supports `acquire` and `release` commands with configurable max wait behavior.
  * Automatically waits for lock availability and reclaims expired locks.
  * Prevents releasing a lock unless the caller matches the lock owner, with verification to avoid false success.

### DIFF
--- a/.github/scripts/test-cluster-lock.sh
+++ b/.github/scripts/test-cluster-lock.sh
@@ -19,6 +19,16 @@
 # the nightly acquires only after its own triggered deploys have finished, those
 # deploys run while the lock is free — no deadlock.
 #
+# Robustness (test-suites#574 review): the lock's whole job is correctness under
+# concurrency, so failures must never fake a result.
+#   - Lock state is read in a SINGLE `kubectl get` (holder + timestamp always
+#     come from the same object version — no torn read).
+#   - A *failed* read (transient API error) never triggers a stale-reclaim; only
+#     a successfully-read timestamp older than LOCK_TTL_SECONDS does. This stops
+#     an API blip from reclaiming a live lock and letting two runs proceed.
+#   - `release` verifies the delete instead of assuming success, so a caller is
+#     never told the lock is free while it is still held.
+#
 # Usage:
 #   test-cluster-lock.sh acquire <holder> [max_wait_seconds]
 #   test-cluster-lock.sh release <holder>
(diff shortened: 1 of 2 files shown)


valentinyanakiev:
Addressed the lock-script review in the latest push:

- **Critical — torn read / stale-reclaim on transient failure:** holder + timestamp now come from a **single** `kubectl get` (`read_lock`). A *failed* read no longer computes staleness — `read_lock` distinguishes NotFound (rc 2) from a transient API error (rc 1), and reclaim fires only on a numeric timestamp actually read. An API blip can no longer reclaim a live lock (which would let a deploy and a nightly run at once).
- **Major — `create` failure misread as "held":** a create failure with no lock present is now retried, not treated as contended.
- **Major — `release` false success:** the delete is verified (with an already-gone re-check) instead of `|| true` + an unconditional success log; a transient read during release reports no false success and leaves the stale-TTL as backstop.

Verified against a stubbed `kubectl`: transient read during acquire preserves the live lock; release on a transient read reports no false success; stale reclaim still works on a real old timestamp.

Declined the `TestScenarioFactory` retry "heavy lift" (cleanup-between-retries + tests): the env is wiped nightly by `cleanup-database.yml`, so orphans from a failed attempt are harmless, and the lib has no test harness — noted on #575.

PR #6260, opened 2026-07-12
fix(oidc): never crash the server on issuer discovery failure

valentinyanakiev:
## Why

`OidcService.onModuleInit` awaits `discoverWithRetry` (added in #6241), which — after its 10 attempts (~90s of capped backoff) — **`throw`s**. That throw escapes `onModuleInit` as an **unhandled promise rejection and kills the process (exit 1)**. So when the identity chain (Hydra + traefik) is slow to settle during a deploy/restart wave and discovery outlasts the retry budget, the **whole server crashes on boot**.

On **2026-07-12** this happened on a fresh test deploy: the server crashed at boot (OIDC discovery to `identity.test-alkem.io` timed out 10×), and the disruption cascaded ~**50 nightly suites** (`Space ID is required` setup failures) — see the test-suites#563 investigation. #6241 turned crash-*looping* into a single crash-after-90s, but a boot-fatal discovery remained.

## What

OIDC discovery is used **only** by the interactive login controller (`oidc.controller.ts`) — `getClient()`/`getIssuer()`/`getPreAuthSigningKey()` have no other callers. Token validation and non-interactive-login don't touch it. So a failed discovery must **not** be boot-fatal.

- `onModuleInit` is now synchronous and **fire-and-forget**: `void this.initDiscovery()`. The server boots immediately and serves everything (incl. token validation) regardless of identity readiness.
- `initDiscovery()` runs `discoverWithRetry` in **rounds, indefinitely**, with a 15s gap between rounds. It **never rejects**, so it can't crash the process, and OIDC **self-heals** once the issuer is reachable — no restart needed.
- `getClient()`/`getIssuer()` throw a clear *"not yet initialised — retry shortly"* for interactive-login requests that land before discovery completes (instead of the whole API being down).

This is also the correct **production** behaviour: a transient identity blip should degrade interactive login briefly, not take the API offline.

## Testing

- Biome clean; no `OidcService` spec exists to update, and no spec references `onModuleInit`/`discoverWithRetry`.
- Logic verified by inspection: boot no longer awaits discovery; the background loop catches `discoverWithRetry`'s throw and retries; success sets issuer+client and logs.

> Local `tsc` flags `Cannot find module 'openid-client'` — a pre-existing artifact of an incomplete local install (that import is unchanged); CI has deps.

Refs: #6241, test-suites#563.

## Summary by CodeRabbit

* **Bug Fixes**
  * OIDC discovery now runs in the background and retries continuously until the identity provider becomes reachable.
  * Startup is no longer blocked or treated as failed due to slow or unavailable OIDC discovery.
  * Discovery guards prevent overlapping runs, and error messages now indicate discovery hasn’t finished yet.
* **Tests**
  * Added Vitest coverage for resilient startup, background retry/self-healing behavior, and correct discovery invocation and error states (using fake timers).

### DIFF
--- a/src/core/auth/oidc/oidc.service.spec.ts
+++ b/src/core/auth/oidc/oidc.service.spec.ts
@@ -1,0 +1,111 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { OidcService } from './oidc.service';
+
+// Control issuer discovery by mocking openid-client's static `Issuer.discover`.
+const { discoverMock } = vi.hoisted(() => ({ discoverMock: vi.fn() }));
+vi.mock('openid-client', () => ({
+  Issuer: { discover: discoverMock },
+}));
+
+const OIDC_CONFIG = {
+  issuer_url: 'https://identity.test-alkem.io/',
+  web_client_id: 'alkemio-web',
+  web_redirect_uri: 'https://test-alkem.io/auth/callback',
+};
+
+const makeService = (): OidcService => {
+  const configService = { get: vi.fn().mockReturnValue(OIDC_CONFIG) } as any;
+  const logger = { log: vi.fn(), warn: vi.fn(), error: vi.fn() } as any;
+  return new OidcService(configService, logger);
+};
+
+// A discovered issuer exposes a `Client` constructor; `new issuer.Client(...)`
+// must yield the resolved client instance (a real class, so the constructor's
+// object return is honoured exactly like openid-client's).
+const makeFakeIssuer = () => {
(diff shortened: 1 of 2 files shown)


valentinyanakiev:
Addressed the nitpick in `a` follow-up commit: added a 4th test — *"runs discovery only once — the guard blocks overlapping and post-success rounds"* — that exercises `initDiscovery`'s `if (this.client || this.discovering) return` for both an in-flight overlap (two `onModuleInit()` calls) and a post-success re-init. 4/4 green, Biome clean.

PR #6271, opened 2026-07-15
Fix export whole-day calendar events having wrong date

ccanos:
## Schema Contract Checklist (Feature 002)

If this PR affects the GraphQL schema, complete ALL items below:

- [ ] Ran `npm run schema:print` (and optionally `npm run schema:sort`) to regenerate `schema.graphql`.
- [ ] Retrieved base snapshot (e.g. `git show origin/develop:schema.graphql > tmp/prev.schema.graphql`).
- [ ] Ran `npm run schema:diff` to produce `change-report.json` and `deprecations.json`.
- [ ] Reviewed classifications; only expected changes present.
- [ ] (Optional) Ran `npm run schema:validate` and artifacts passed validation.
- [ ] Committed ONLY `schema.graphql` (not the JSON artifacts).

### Change Report Summary

Paste the key counts from `change-report.json` (example format below) — remove any zero lines if desired:

```
Additive: X
Deprecated: Y
Deprecation Grace: Z
Breaking: B (override applied? yes/no)
Premature Removals: P
Invalid Deprecations: I
Info: N
```

### Deprecations Added / Updated

List new or updated deprecations with schedules (`REMOVE_AFTER=YYYY-MM-DD | reason`). Indicate any in grace period.

### Breaking Changes (If Any)

If BREAKING changes are intentional:

- Rationale:
- Risk assessment / mitigation:
- CODEOWNER approval with phrase `BREAKING-APPROVED` requested? (link)

### Other Notes

---

Reference docs: `specs/002-schema-contract-diffing/quickstart.md` for full workflow and troubleshooting.

PR #6284, opened 2026-07-20
fix: export whole-day calendar events on the correct date, TZ-independent (workspace#023-wholeday-calendar-timezone)

### LINKED ISSUE #6267: BUG: wrong time in calendar
**Describe the bug**
When I upload an Alkemio event to my outlook / google calendar, the dates change

**To Reproduce**

1. Go to an calendar event
2. Add to your outlook or google calendar
3. See the dates change

**Expected behavior**
Similar dates when I download

**Screenshots**

Image

Image

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Areas that will be affected as part of the fix**
_Optional, to be added by developer if relevant_

### LINKED ISSUE #6279: Whole-day calendar events export/display on the correct date (timezone-independent) — follow-up of server#6271
**Follow-up of** alkem-io/server#6271 (Carlos Cano — the first step) · **Closes** alkem-io/server#6267 · **Spec (source of truth):** `alkem-io/agents-hq` → `specs/023-wholeday-calendar-timezone/`

## Summary

A **whole-day** calendar event exports/displays on the **wrong day** — one day early — for any user whose timezone differs from the server's. On production that's **every user east of UTC (all of Europe)**: a whole-day event created for *3 December* imports into Google as *2 December* (server#6267).

A whole-day event is a **bare calendar date** ("all day on 3 December") — no time, no timezone. But it's transported as a timezone-dependent *instant*: the client sends the creator's **local** midnight, and the server collapses that instant to a single day in a **different** zone (UTC on the Alpine/UTC server), so the day shifts.

## Why server#6271 wasn't enough

server#6271 (merged — the first step):
- ✅ **Kept:** the RFC 5545 *exclusive* whole-day end date (ICS/Outlook no longer end a day early).
- ⚠️ **Incomplete:** derived the start day from **server-local** parts — correct only when server TZ == creator TZ. Production runs UTC (Alpine default, no `TZ`); the fix was validated on the dev stack pinned to `Europe/Amsterdam` ("works on my machine") and is **still broken on dev/acceptance**. It did not close server#6267.

## The fix

One invariant across **server** + **client-web**:

> A whole-day event's start is **UTC-midnight of the intended date**; all whole-day date math is in **UTC**, never re-projected through a local offset.

- **client-web** — write as UTC-midnight (not local `startOfDay`); display as a local "floating" date so every viewer sees the right day; whole-day durations computed in UTC (DST-safe).
- **server** — derive `DTSTART`/`DTEND` from **UTC** parts (keeps server#6271's exclusive end).
- **Timed events unchanged** · **no schema/GraphQL change** · **fix-forward, no data migration** (existing events self-correct on next edit).

### 1) multi-day events highlighted too many days

While reviewing the whole-day fix we found a related, pre-existing bug in the calendar / badge / batch-export display: a **multi-day event covered too many days**. A 2-day whole-day event (23 → 25 July) lit up **five** calendar cells (23–27 — see Additional Context).

Image

Image

**Cause:** the client's `endDateFromDuration` added `durationDays` *on top of* `durationMinutes`, but `durationMinutes` already holds the **full span** (the server documents it as authoritative — *"durationMinutes always holds the full event duration, including days converted to minutes"*), so the days were double-counted (a 2-day span rendered as 4).

**Fix:** derive the end from `durationMinutes` alone (DST-safe — whole days via calendar-day arithmetic); `durationDays` is a derived field and is never added on top. Fixes the calendar grid, the date badge, and the batch ICS export. Applies to timed multi-day events too.

### 2) whole-day events have no time-of-day (added while testing)

Found while testing the fix (pre-existing, not caused by the TZ work), tracked here as added scope:

- [x] The whole-day event edit form must not show **Start time / End time / Duration** — only Start date + End date (the span is the date range).
- [x] A whole-day event's `durationMinutes` is derived from the Start→End date range on **save** — a stale sub-day duration must not survive: a single-day whole-day event is zero-length, and editing a multi-day event down to a single day **persists** (no revert).

Repro: a single-day whole-day event showed a disabled "Duration: 2 HOURS · ends at 2:00 AM", and editing a 4-day event down to a single day reverted to the original end on save (the same-day branch kept the old 72h span). Growing single → multi worked; shrinking multi → single did not. Fixed on `client-web` `feat/023-…`; spec **US5 / FR-012 / SC-008**.

Changing a 4-day event into a 1-day event keeps the 96-hours duration and brings it back to 4-day eevent after save:

Image

changing it to 1-day event (Jul 20):

Image

and you can choose time duration for 1-day events like it's not 24-hours:

Image

Image

---

## Acceptance criteria

- **Export correct everywhere** — for every creator-browser TZ × server TZ (`UTC`, `Europe/Amsterdam`, `Europe/Sofia`, a west-of-UTC zone), a whole-day event for date *D* exports (Google / Outlook / ICS `DTSTART;VALUE=DATE`) with start = *D* — **0 cells off by a day** (incl. NL-on-UTC); the whole-day end is the RFC 5545 exclusive end in all three targets.
- **Display correct everywhere** — shows date *D* in the timeline for viewers in every TZ; the edit dialog round-trips the chosen dates.
- **Multi-day exact** — a multi-day event covers exactly its days in the calendar grid, date badge, and batch export (a 2-day event highlights 3 cells, never 5); DST-crossing spans keep the exact day count.
- **Timed events unchanged** (regression guard).
- **server#6267 closed** — no new "wrong day in calendar" reports for new whole-day events.

---

## QA test plan

Covers **both** PRs — [`server#6284`](https://github.com/alkem-io/server/pull/6284) and [`client-web#10049`](https://github.com/alkem-io/client-web/pull/10049) — which must be tested **together** (they ship in the same release; the fix is a client↔server convention).

### Setup

- **Server timezone matters.** The bug only shows when the browser TZ ≠ server TZ. Dev/acceptance servers run **UTC**, which is the reproducing configuration — test there, not against a local stack pinned to `Europe/Amsterdam` (that is the "works on my machine" config that hid the bug in #6271).
- **Change the browser timezone**, not just the OS clock: Chrome DevTools → ⋮ → More tools → Sensors → Location, or run the browser with `TZ=... `. Test at minimum: `Europe/Sofia` (UTC+2, east — the reported case), `Europe/Amsterdam` (UTC+1, the NL userbase — **also broken today**), `Pacific/Honolulu` (UTC−10, west), and `UTC` itself.
- Reload the page after switching timezone.

### A. Whole-day date correctness (the reported bug — US1/US2)

| # | Steps | Expected |
|---|---|---|
| A1 | Browser = Sofia (UTC+2). Create a whole-day event, Start **3 Dec 2026**, End **4 Dec 2026**. | Card / date badge / detail view / calendar grid all show **3–4 December** (not 2–3). |
| A2 | On A1's event, "Add to Google Calendar". | Google shows **3 Dec – 4 Dec** inclusive. |
| A3 | On A1's event, "Add to Outlook". | Outlook shows **3–4 December**. |
| A4 | Download the single-event `.ics` and open it in a text editor. | `DTSTART;VALUE=DATE:20261203` and `DTEND;VALUE=DATE:20261205` (RFC 5545 **exclusive** end = last covered day + 1). Import into Outlook/Thunderbird → shows 3–4 Dec, **not ending a day early**. |
| A5 | Repeat A1–A4 with the browser in **Amsterdam (UTC+1)**. | Same dates. (This case is broken on production today — it is not only a Sofia problem.) |
| A6 | Repeat A1–A4 with the browser in **Honolulu (UTC−10)**. | Same dates (the anchoring is symmetric — no drift the other way either). |
| A7 | Repeat A1–A4 with the browser in **UTC**. | Same dates — **no regression** in the case that already worked. |
| A8 | Open the **same** event from a *second* browser in a *different* timezone (e.g. created in Sofia, viewed in Honolulu). | Both viewers see **3 December**. |
| A9 | Reopen the edit dialog on the event. | Start/End pickers show the originally chosen dates — no ±1 drift, and no drift after repeated open→save→reopen cycles. |

### B. Whole-day edge dates

| # | Steps | Expected |
|---|---|---|
| B1 | Single-day whole-day event (Start = End = 20 Jul). | Saves without error; shows one day; ICS `DTSTART 20260720` / `DTEND 20260721`. **This previously failed server validation** (`start >= end`) — verify no "Invalid calendar event date range" error on the Google/Outlook links, the `.ics` download, **or** the notification email. |
| B2 | Whole-day event **31 Dec 2026 → 1 Jan 2027**. | Correct across the year boundary in display and export. |
| B3 | Whole-day event spanning a **DST transition** (e.g. 28 → 30 March 2026, EU spring-forward on 29 March). | Exactly 3 days covered; `DTEND` = `20260331`; no ±1 day/hour drift in grid or export. |

### C. Multi-day span rendering (scope creep #1 — US4)

| # | Steps | Expected |
|---|---|---|
| C1 | Whole-day event **23 → 25 July** (3 covered days). | Calendar grid highlights **exactly 23, 24, 25** — the old bug lit up 5 cells (23–27). |
| C2 | Whole-day event 23 → 24 July (2 covered days). | Exactly **2** cells highlighted, plus the date badge showing the same span. |
| C3 | **Timed** multi-day event (e.g. 16 Jul 12:56 → 23 Jul 04:03). | Grid, date badge and card header all span exactly 16–23 July — this path changed too and must not over- or under-count. |
| C4 | **Batch export**: "Export events to ICS" for a calendar containing C1 + C3. | Whole-day entry ends on the exclusive next day; timed entry's `DTEND` matches its real end instant. Import into Google/Outlook and compare spans against the app. |

### D. Whole-day has no time-of-day (scope creep #2 — US5)

| # | Steps | Expected |
|---|---|---|
| D1 | Open the event form, toggle **Whole day** on. | **Start time, End time and Duration disappear entirely** (not just greyed out). Only Start date + End date remain; the row re-flows to 2 columns without layout breakage on desktop **and** mobile width. |
| D2 | Create a whole-day 4-day event (20 → 23 Jul), save, reopen, set **End date = Start date (20 Jul)**, save. | It **persists as a single day**. Reopen → still 20 → 20 Jul. (Previously it reverted to the 4-day span.) |
| D3 | Reverse: single-day whole-day → widen to 4 days, save, reopen. | Persists as 4 days (this direction worked before — regression guard). |
| D4 | Take an existing **timed** event with a 2-hour duration, toggle **Whole day** on, save. | No "ends at 2:00 AM" / stale duration anywhere; the span is purely the date range. |
| D5 | Take a whole-day multi-day event, toggle **Whole day** *off*. | Duration resets to a sane sub-day default (30 min) — the multi-day span must **not** leak into the duration picker or into the export. |

### E. Regressions to guard — timed events (must be unchanged)

These paths were touched by the shared helpers; verify they behave exactly as before the change.

| # | Steps | Expected |
|---|---|---|
| E1 | Create a same-day **timed** event 14:00 → 16:00. | Displays 14:00–16:00 in the viewer's local timezone; Google/Outlook/ICS import at the correct **wall-clock time** in the importer's timezone; `DTSTART:` is a real UTC instant (`…Z`), *not* a `VALUE=DATE`. |
| E2 | Create an **overnight** timed event 18:00 → 09:00 next day. | Accepted by the form (no false "end before start" error) and exported over the correct two days. |
| E3 | Create a timed event and view it from a **different** browser timezone. | Time shifts correctly with the viewer's zone (timed events are instants — this is *correct* behaviour and must not be "fixed"). |
| E4 | Attempt a timed event with **end exactly equal to start** (zero length). | Still **rejected** — the validation was relaxed *only* for whole-day events. |
| E5 | Edit an existing timed event and change **only the title/description** (leave the dates and duration untouched), save. | Duration and dates are unchanged after save. (The server now assigns `durationMinutes`/`durationDays` unconditionally, so verify a no-op edit doesn't zero or alter them.) |

### F. Other surfaces to sweep

- **Event notification email** (a whole-day event triggers a "calendar event created" notification): the date in the email body and any calendar attachment must be the picked date, and a single-day whole-day event must not error out the notification.
- **Space sidebar / upcoming-events list / event detail page** — whole-day events show the correct day in each.
- **Public / non-logged-in views** of the timeline, if reachable.
- **Recurring reload / cache**: hard-refresh after changing the browser timezone to confirm nothing is cached at the wrong day.

### G. Known & accepted — do NOT raise as bugs

- **Pre-existing whole-day events are not migrated** (FR-009, deliberate: re-interpreting a timezone-naive column is lossy and would risk breaking currently-correct rows). An event created *before* this release may still show/export one day early **until it is edited and re-saved**, at which point it self-corrects. Please verify that self-correction works: open an old whole-day event, save it unchanged, and confirm the date becomes correct.
- **Timed events shifting with the viewer's timezone** is correct by design — they are real instants.
- The two PRs must be **deployed in the same release**; testing the client against an old server (or vice versa) is not a supported configuration.

### Exit criteria

- Sections A–D fully green across all four browser timezones against a **UTC** server — **0 cells off by a day**.
- Section E shows no change in timed-event behaviour versus the current acceptance environment.
- server#6267 reproduction steps no longer reproduce.

---

## Notes

- **Timed events out of scope** — already correct (real UTC instant + `X-WR-TIMEZONE`).
- **No back-data migration** — deferred (re-interpreting a timezone-naive column is lossy and can regress currently-correct rows).
- **To verify** — whole-day date rendering in the alert/email path (server `notification.external.adapter.ts`, the other half of server#5946); raise a linked follow-up if affected.
- **Repos:** `server`, `client-web`. Design / plan / contract / ADR: `alkem-io/agents-hq` `specs/023-wholeday-calendar-timezone/` + `docs/decisions/0001-wholeday-calendar-utc-midnight-contract.md`.

polibb:
Closes the export half of the whole-day calendar timezone bug. Follow-up of #6271 (superseded-in-part — its exclusive-end fix is kept).

## Problem
A **whole-day** event is a bare calendar date, but the export collapses the stored instant to a single day using the server's process timezone. On the UTC production server that lands a day early for every creator east of UTC (all of Europe). #6271 switched to *server-local* parts — correct only when the server TZ equals the creator's — which is why it was still broken on dev/acceptance.

## Change (export only — no schema, no migration)
- Derive whole-day `DTSTART`/`DTEND` (and the Google/Outlook date-only values) from **UTC** parts of the stored instant (`formatDateOnlyUtc` / `formatUtcDateHyphenated`), so the exported day is identical under every server timezone.
- Compute the RFC 5545 exclusive end (`+1` day) in **UTC** (`setUTCDate`) — **keeps #6271's genuine exclusive-end fix**, now DST-safe.
- **Timed events unchanged** (real UTC instant + `X-WR-TIMEZONE`).
- The client (paired PR) sends whole-day dates as UTC-midnight so this holds end to end.

## Tests
- New TZ-matrix spec — identical output under `TZ=UTC / Europe/Amsterdam / Europe/Sofia / America/Los_Angeles` (the west-of-UTC case that produced `20261202` now yields `20261203`); DST-crossing multi-day guard.
- Existing whole-day tests updated to the UTC-midnight convention; timed export is a byte-for-byte regression guard.
- Full `pnpm test` suite green; **no schema change** (`schema:diff` clean).

## Schema Contract
No GraphQL/schema change — `schema:diff` is empty.

## Links
- Cross-repo spec (source of truth): [`alkem-io/agents-hq` `specs/023-wholeday-calendar-timezone/`](https://github.com/alkem-io/agents-hq/pull/65) (+ `docs/decisions/0001-…`)
- Story: #6279 · Closes: #6267 · Supersedes-in-part: #6271
- **Paired client-web PR — ships in the same release** (must not be split).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed whole-day calendar exports to use consistent UTC date-only values across ICS, Outlook, and Google Calendar.
  - Corrected whole-day range handling, including the required exclusive end date and scenarios where start equals end.
  - Ensured event updates correctly persist `0` duration values when converting multi-day events to single-day.

- **Tests**
  - Added timezone-invariance coverage for whole-day export formatting (including child-process runs under a non-UTC timezone).
  - Added unit coverage for updating calendar event duration fields to `0`.

- **Documentation**
  - Clarified how whole-day date offsets are represented and how exports apply the exclusive `+1 day` rule.

### DIFF
--- a/src/domain/timeline/event/calendar.event.calendar-links.spec.ts
+++ b/src/domain/timeline/event/calendar.event.calendar-links.spec.ts
@@ -5,8 +5,66 @@ import {
   formatDatesForCalendar,
   generateCalendarUrls,
   generateICS,
+  validateCalendarDateRange,
 } from './calendar.event.calendar-links';
 import { ICalendarEvent } from './event.interface';
+
+describe('validateCalendarDateRange', () => {
+  const ID = 'evt-1';
+
+  it('accepts a single-day whole-day event (start === end)', () => {
+    // durationMinutes = 0 ⇒ end === start; the exclusive +1 day is added at export.
+    expect(() =>
+      validateCalendarDateRange(
+        '2026-07-20T00:00:00.000Z',
+        '2026-07-20T00:00:00.000Z',
+        ID,
+        true
+      )
+    ).not.toThrow();
+  });
+
+  it('accepts a multi-day whole-day event (end after start)', () => {
+    expect(() =>
+      validateCalendarDateRange(
+        '2026-07-20T00:00:00.000Z',
+        '2026-07-22T00:00:00.000Z',
+        ID,
+        true
+      )
+    ).not.toThrow();
+  });
+
+  it('rejects a zero-length timed event (start === end)', () => {
+    expect(() =>
+      validateCalendarDateRange(
+        '2026-07-20T10:00:00.000Z',
+        '2026-07-20T10:00:00.000Z',
+        ID,
+        false
+      )
+    ).toThrow();
+  });
+
+  it('rejects an end before the start (whole-day and timed)', () => {
+    expect(() =>
+      validateCalendarDateRange(
+        '2026-07-22T00:00:00.000Z',
+        '2026-07-20T00:00:00.000Z',
+        ID,
+        true
+      )
+    ).toThrow();
+    expect(() =>
+      validateCalendarDateRange(
+        '2026-07-20T12:00:00.000Z',
+        '2026-07-20T10:00:00.000Z',
+        ID,
+        false
+      )
+    ).toThrow();
+  });
+});
 
 describe('CalendarEventCalendarLinks', () => {
   it('formats calendar dates for Google and Outlook', () => {
--- a/src/domain/timeline/event/calendar.event.calendar-links.spec.ts
+++ b/src/domain/timeline/event/calendar.event.calendar-links.spec.ts
@@ -1,7 +1,7 @@
 import {
   calculateCalendarEventEndDate,
   foldIcsLine,
-  formatDateOnlyLocal,
+  formatDateOnlyUtc,
   formatDatesForCalendar,
   generateCalendarUrls,
   generateICS,
@@ -38,38 +38,23 @@ describe('CalendarEventCalendarLinks', () => {
     expect(result.wholeDay).toBe(true);
   });
 
-  it('formatDateOnlyLocal extracts the local wall-clock YYYYMMDD', () => {
-    // Constructed at local midnight so the assertion holds in any runner TZ.
-    const localMidnight = new Date(2026, 1, 20, 0, 0, 0); // 2026-02-20 local
-    expect(formatDateOnlyLocal(localMidnight.toISOString())).toBe('20260220');
+  it('formatDateOnlyUtc extracts the UTC calendar day YYYYMMDD', () => {
+    expect(formatDateOnlyUtc('2026-02-20T00:00:00.000Z')).toBe('20260220');
   });
 
-  describe('whole-day time-zone handling (Europe/Amsterdam)', () => {
-    // Reproduces the reported bug: a whole-day event stored as local midnight
-    // is read back (node-postgres, timestamp-without-tz) as the previous day in
-    // UTC. Slicing the UTC ISO produced an ICS date one day early; the export
-    // must emit the local calendar day instead.
-    const originalTz = process.env.TZ;
+  describe('whole-day date-only handling', () => {
+    // A whole-day event is a bare calendar date, sent by the client as
+    // UTC-midnight of the intended day and exported as that UTC calendar day —
+    // timezone-independent (see calendar.event.calendar-links.wholeday-tz.spec.ts).
 
-    beforeEach(() => {
-      process.env.TZ = 'Europe/Amsterdam';
-    });
+    it('exports the UTC calendar day with an exclusive end for a single-day event', () => {
+      // Whole-day "23 July 2026", single day, stored as UTC-midnight.
+      const result = formatDatesForCalendar(
+        '2026-07-23T00:00:00.000Z',
+        '2026-07-23T00:00:00.000Z',
+        true
+      );
 
-    afterEach(() => {
-      process.env.TZ = originalTz;
-    });
-
-    it('exports the local calendar day for a whole-day event, not the UTC day', () => {
-      // Single-day whole-day event stored at Amsterdam local midnight:
-      // 2026-07-23 00:00 local is 2026-07-22T22:00Z. The last covered day is the
-      // 23rd (the end instant is later the same local day), so DTSTART is the
-      // 23rd and the exclusive DTEND is the 24th.
-      const startIso = '2026-07-22T22:00:00.000Z';
-      const endIso = '2026-07-22T22:30:00.000Z';
-
-      const result = formatDatesForCalendar(startIso, endIso, true);
-
-      // Without the local-date fix DTSTART would slip to the 22nd (UTC).
       expect(result.icalStart).toBe('20260723');
       expect(result.icalEnd).toBe('20260724');
       expect(result.outlookStart).toBe('2026-07-23');
--- a/src/domain/timeline/event/calendar.event.calendar-links.ts
+++ b/src/domain/timeline/event/calendar.event.calendar-links.ts
@@ -307,12 +307,22 @@ export const isIsoDateString = (value: string): boolean =>
 export const validateCalendarDateRange = (
   startIso: string,
   endIso: string,
-  eventId: string
+  eventId: string,
+  wholeDay = false
 ): void => {
   const start = new Date(startIso).getTime();
   const end = new Date(endIso).getTime();
 
-  if (Number.isNaN(start) || Number.isNaN(end) || start >= end) {
+  // A whole-day event may be a single day (start === end, i.e. durationMinutes 0);
+  // its exclusive end is added at export. A timed event must end strictly after
+  // its start.
+  const invalid =
+    Number.isNaN(start) ||
+    Number.isNaN(end) ||
+    start > end ||
+    (!wholeDay && start === end);
+
+  if (invalid) {
     throw new ValidationException(
       'Invalid calendar event date range',
       LogContext.CALENDAR,
--- a/src/domain/timeline/event/calendar.event.calendar-links.ts
+++ b/src/domain/timeline/event/calendar.event.calendar-links.ts
@@ -65,29 +65,25 @@ export const formatDatesForCalendar = (
   const endIso = toIsoString(end, 'endDate');
 
   if (wholeDay) {
-    // A whole-day event carries a calendar date, not an instant. The stored
-    // `timestamp` (without time zone) is read back by node-postgres in the
-    // process-local time zone — the same reference the client renders in — so
-    // the date-only value MUST be derived from the local wall-clock components,
-    // not from the UTC ISO string. Slicing the UTC ISO lands on the previous
-    // day for time zones east of UTC (e.g. Europe/Amsterdam, where the server
-    // runs: local midnight 2026-07-23 is 2026-07-22T22:00Z in UTC).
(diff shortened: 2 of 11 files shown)


polibb:
Paired client-web PR: **alkem-io/client-web#10049** — ships in the same release (must not be split).

bobbykolev (approved):

Issue #5290, opened 2025-07-31
BUG: When L1 converted to L0, redirect to it is wrong


Comoque1:
**Describe the bug**

**To Reproduce**

1. Go to acc env
2. Convert L1 to L0
3. Go to account page
4. Click on the converted space - user is redirected to 404 page - the old L1 url
<img width="540" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBekRvQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f70fd2f048606fef387873f2201ab08d0ca8b20d/image.png" alt="image.png" />

**Expected behavior**
User is redirected to the new space.

SimoneZaza:
@Comoque1 does this issue persist after refreshing the (account) page?

hero101:
Tracked in [alkem-io/server#6020](https://github.com/alkem-io/server/issues/6020) (sub-issue of epic alkem-io/alkemio#1846, whiteboard step 1). This bug remains open and will be closed manually after the fix is verified end-to-end.

Comoque1:
The issue is still reproducible with latest develop changes.

Comoque1:
The above comment was related to [Callout transfer when flow is changed from template](https://github.com/alkem-io/client-web/issues/9353).

This one is actually fixed

PR #6338, opened 2026-08-05
fix(spaces): sweep the innovation flow profile URL cache on subtree invalidation

### LINKED ISSUE #6020: Invalidate URL cache on space transfer and L1→L0 conversion
## Description

As a Global Admin
I want stale URL cache entries cleared on space transfer and L1→L0 promotion
So that admins and users hit the new URL immediately, not a 404 to the old path

Triaged from alkem-io/alkemio#1846 (whiteboard step 1). Cache invalidation is already wired for callouts; spaces miss it.

## Acceptance criteria

- [ ] Transferring a space invalidates URL cache entries for that space
- [ ] Promoting L1→L0 invalidates URL cache entries for the source path
- [ ] Endpoints used by the client to resolve the promoted space return the new path (verified against alkem-io/client-web#9481)
- [ ] If full invalidation is impractical, a 30-min TTL is acceptable as a documented fallback
- [ ] Verified against alkem-io/server#5290 and alkem-io/client-web#9481 (existing bugs remain open until manually closed — do not auto-close)

## Additional Context

Whiteboard: "Bug: URLs cache is not being cleared after transfer for spaces. This does happen for Callouts." Reuse the callout-side invalidation path.

Parent epic: alkem-io/alkemio#1846 (whiteboard step 1).

hero101:
Closes #6020 (QA follow-up — this PR completes the story, so merging it closes #6020. #5290 and alkem-io/client-web#9481 are **not** closed by this PR and stay open for manual closure per the story's acceptance criteria).

## Context

The bulk of story #6020 already shipped in #6037 (`beb689ef6`): `SpaceService.invalidateUrlCacheForSpaceSubtree()` plus its wiring into space transfer (`account.resolver.mutations.ts`) and the L0/L1/L2 conversion mutations. This PR closes the one residual gap QA found afterwards:

> `innovationFlow.profile.url` is not updated after promotion of L1 to L0 — @Comoque1, 2026-06-16

## Root cause

`invalidateUrlCacheForSpaceSubtree()` swept exactly two things: `space.about.profile.id`, and everything reachable via `revokeUrlCachesForCalloutsInSpaces()` (callout framing, framing whiteboards, contribution profiles). The **innovation flow profile** — `space.collaboration.innovationFlow.profile` — was never in either sweep.

`ProfileType.INNOVATION_FLOW` resolves through `getInnovationFlowUrlPathOrFail()`, which derives the URL from the owning space's path, and `generateUrlForProfile()` caches it under the profile id. So after an L1 → L0 promotion the cache entry survived and kept serving the pre-promotion path until TTL.

## Fix

`space.collaboration.innovationFlow.profile` is now revoked for every space in the subtree, alongside the about profile.

**Why extend the TypeORM `relations` rather than add a SQL sweep next to `revokeUrlCachesForCalloutsInSpaces`:** the callout sweep is raw SQL because callouts/contributions are N-per-space, so relation loading would blow up the row count. `space → collaboration → innovation_flow → profile` is 1:1 the whole way, so it is a plain left join on the `find()` that already runs — zero extra round trips, and it reuses the existing per-profile `try`/`catch` isolation and Winston log shape verbatim. `loadEagerRelations: false` was added at the same time so the newly joined `collaboration` / `innovationFlow` / `profile` rows don't drag their eager `authorization` policies along; only profile ids are read here.

**Innovation flow states:** checked, and they have no gap. `InnovationFlowState` carries no `Profile` — `displayName` / `description` are plain columns on `innovation_flow_state` — so states hold no URL cache entry and need no sweep. Noted in a code comment so the next reader doesn't have to re-derive it.

## Tests

`src/domain/space/space/space.service.spec.ts`:

- **New regression test** `revokes the innovation flow profile of every space in the subtree` — asserts both `flow-profile-*` ids are revoked and that the single `find()` actually requests the `collaboration.innovationFlow.profile` relation. Verified **red on `develop`** (`expected "vi.fn()" to be called with arguments: [ 'flow-profile-space-root' ]`), green with the fix.
- New `skips spaces whose collaboration carries no innovation flow profile` (L0 spaces mid-conversion / spaces without collaboration).
- New `logs and keeps sweeping when revoking one profile throws` — pins the error-isolation contract: one failed revoke must not abort the rest of the subtree or the callout sweep, and must log with the documented `{ message, spaceId, profileId }` shape.
- Existing subtree tests extended with innovation flow fixtures.

While in there, the `invalidateUrlCacheForSpaceSubtree` block was made order-independent. `urlGeneratorCacheService` is a `@golevelup/ts-vitest` proxy auto-mock whose methods only materialise on access, so `vi.spyOn(urlGeneratorCacheService, 'revokeUrlCache')` only worked because an earlier `describe` in the file had already assigned the property — the tests failed the moment they were run with `-t`. The doubles are now assigned outright in a local `beforeEach`. Same class of latent order-dependence `docs/testing-flakiness.md` calls out.

## Gates

- `pnpm lint` (tsc --noEmit + biome) — clean
- `pnpm build` — clean
- `pnpm test:ci:no:coverage` — 681 files / 7742 tests passed, 0 failures

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL-cache invalidation for spaces, including details, guidelines, innovation flows, framing memos, collaborative documents, contributions, and calendar events.
  * Cache cleanup now continues when individual entries fail.
  * Space name changes now save correctly before related caches are refreshed.
* **Tests**
  * Expanded coverage for missing profiles, cache failures, and additional content relationships.

### DIFF
--- a/src/domain/space/space/space.service.spec.ts
+++ b/src/domain/space/space/space.service.spec.ts
@@ -153,7 +153,7 @@ describe('SpaceService', () => {
       // Mock the URL cache service - use direct assignment for mock objects
       const revokeUrlCacheSpy = vi.fn().mockResolvedValue(undefined);
       urlGeneratorCacheService.revokeUrlCache = revokeUrlCacheSpy;
-      (urlGeneratorCacheService as any).revokeUrlCachesForCalloutsInSpaces = vi
+      (urlGeneratorCacheService as any).revokeUrlCachesForContentInSpaces = vi
         .fn()
         .mockResolvedValue(undefined);
 
@@ -268,7 +268,7 @@ describe('SpaceService', () => {
       // Mock the URL cache service - use direct assignment for mock objects
       const revokeUrlCacheSpy = vi.fn().mockResolvedValue(undefined);
       urlGeneratorCacheService.revokeUrlCache = revokeUrlCacheSpy;
-      (urlGeneratorCacheService as any).revokeUrlCachesForCalloutsInSpaces = vi
+      (urlGeneratorCacheService as any).revokeUrlCachesForContentInSpaces = vi
         .fn()
         .mockResolvedValue(undefined);
 
@@ -422,7 +422,7 @@ describe('SpaceService', () => {
       revokeSpy = vi.fn().mockResolvedValue(undefined);
       revokeCalloutsSpy = vi.fn().mockResolvedValue(undefined);
       (urlGeneratorCacheService as any).revokeUrlCache = revokeSpy;
-      (urlGeneratorCacheService as any).revokeUrlCachesForCalloutsInSpaces =
+      (urlGeneratorCacheService as any).revokeUrlCachesForContentInSpaces =
         revokeCalloutsSpy;
     });
 
--- a/src/domain/space/space/space.service.spec.ts
+++ b/src/domain/space/space/space.service.spec.ts
@@ -609,6 +609,38 @@ describe('SpaceService', () => {
         LogContext.SPACES
       );
     });
+
+    it('revokes in bounded batches rather than fanning the whole subtree out at once', async () => {
+      const rootId = 'space-root';
+      // Each space contributes 4 profile ids plus its space-keyed ids, so 20
+      // spaces is comfortably more than one batch.
+      const childIds = Array.from({ length: 19 }, (_, i) => `space-${i}`);
+      const allIds = [rootId, ...childIds];
+
+      stubDescendants(childIds);
+      vi.spyOn(spaceRepository, 'find').mockResolvedValue(
+        allIds.map(spaceWithProfiles) as any
+      );
+
+      let inFlight = 0;
+      let maxInFlight = 0;
+      revokeSpy.mockImplementation(async () => {
+        inFlight++;
+        maxInFlight = Math.max(maxInFlight, inFlight);
+        // Yield so every revoke started in the same batch overlaps.
+        await Promise.resolve();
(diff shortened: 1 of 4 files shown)


hero101:
Round 3 — outside-diff / suppressed review items, judged:

**Declined — CodeRabbit, `space.service.ts:903-956`, "Emit a domain event for the rename" (Major).**
The guideline is real and I agree with it in principle, but the direct write it flags is not introduced by this PR. `updateSpacePlatformSettings` has used `this.save(space)` since `209987c28`, and `SpaceService` calls `save()` in 9 places. This PR only *reordered* that existing call so invalidation runs after the commit rather than before it — which is strictly closer to the event-driven ordering CodeRabbit is asking for, not further from it. Converting the space rename flow to emit a domain event is an architectural change to a shared service, well outside a URL-cache invalidation fix, and doing it for one of nine call sites would leave the service half-migrated. Worth its own story; not this PR.

**Accepted — Copilot, `space.service.ts:1041`, inaccurate concurrency comment.** Correct: `Promise.all` isolates failures and starts revokes concurrently, but the sweep still awaits the slowest. Comment reworded to say it is bounded by the slowest entry rather than by their sum.

**Accepted — Copilot, `url.generator.service.cache.ts:57`, misleading method name.** Correct: it now sweeps calendar events too. Renamed `revokeUrlCachesForCalloutsInSpaces` -> `revokeUrlCachesForContentInSpaces`. One production call site; `tsc` clean.

Issue #4970, opened 2025-02-21
BUG: Transferring callout from subspace of space 1 to subspace of space 2 makes it not visible


Comoque1:
**Describe the bug**

**To Reproduce**

1. Go to DEV env
2. Create space 'A' and a subspace 'AA' to it
3. Create callout
4. Create space 'B' and a subspace 'BB' to it
5. Transfer the callout from subspace 'AA' to the subspace 'BB' - request is successful
6. Check for the callout - it is not available on subspace 'BB'

**Expected behavior**
Callout is available on subspace 'BB'

**Additional context**
After transferring the callout to space 'B' and back to subspace 'BB' it becomes available

**Areas that will be affected as part of the fix**
_Optional, to be added by developer if relevant_

hero101:
Tracked in [alkem-io/server#6021](https://github.com/alkem-io/server/issues/6021) (sub-issue of epic alkem-io/alkemio#1846, whiteboard step 2). This bug remains open and will be closed manually after the fix is verified end-to-end.

Comoque1:
This issue is resolved.

PR #6339, opened 2026-08-05
fix(callouts): fall back to destination default flow state on cross-space transfer (#6021)

### LINKED ISSUE #6021: Reset callout to destination default flow state when current state has no match
## Description

As a Global Admin transferring a callout across spaces
I want the callout to fall back to the destination's default flow state when its current state name does not exist in the destination
So that the callout remains visible after the move instead of disappearing into an unmatched state

Triaged from alkem-io/alkemio#1846 (whiteboard step 2). Whiteboard notes this was working and regressed; same root cause for both bugs.

## Acceptance criteria

- [ ] Cross-space transfer where source state name is absent in destination → callout assigned destination default state
- [ ] Cross-space transfer where source state name is present in destination → callout keeps that state name (matches by name)
- [ ] Same logic applies when a destination space's innovation flow was changed via template (verified against alkem-io/client-web#9353)
- [ ] Transferred callout is visible in the destination after the move
- [ ] Regression test added so this does not slip again
- [ ] Verified against alkem-io/server#4970 and alkem-io/client-web#9353 (existing bugs remain open until manually closed — do not auto-close)

## Additional Context

Whiteboard: "Callout moved to default state if not matching", "This is I think the same cause as 4970. The callout being transferred should pick up the default state. I thought this was fixed."

Parent epic: alkem-io/alkemio#1846 (whiteboard step 2).

hero101:
## Summary

A Callout transferred across Spaces could disappear in the destination because its `flow-state` classification ended up holding a value the destination does not recognise. This makes the fallback resolve against the destination's *actual* allowed states, so an unmatched Callout lands on the destination's default state and stays visible.

Closes #6021

## Root cause

`CalloutTransferService.transferCallout` re-points the Callout's `flow-state` classification tagset at the destination `CalloutsSet`'s `TagsetTemplate` via `ClassificationService.updateTagsetTemplateOnSelectTagset`. Matching by name already worked; the *fallback* did not:

```ts
const defaultTags = tagsetTemplate.defaultSelectedValue
  ? [tagsetTemplate.defaultSelectedValue]
  : [];
```

`defaultSelectedValue` is a nullable column, and `TagsetTemplateService.updateTagsetTemplateDefinition` only overwrites it when the incoming update carries a truthy value. So a destination template can legitimately carry:

- **no default at all** &rarr; the tagset was set to `[]`, or
- **a stale default** naming a state that no longer exists &rarr; the tagset kept a value the destination does not know.

Either way `filterCalloutsByClassificationTagsets` (which the client drives with `classificationTagsets: [{ name: "flow-state", tags: ["<phase>"] }]`) matches nothing, so the Callout is filtered out of every tab and looks like it vanished.

## Fix

`ClassificationService` now resolves the fallback through a `resolveDefaultTags()` helper:

1. Prefer `defaultSelectedValue` **only when it is itself one of `allowedValues`**.
2. Otherwise fall back to the first allowed value — the same "destination default" convention already used by `CalloutsSetService.moveCalloutsToDefaultFlowState` and the flow-state template bootstrap in `CollaborationService`.
3. When the template declares no `allowedValues` at all (free-form templates), keep whatever default it declares — preserving existing behaviour.

Empty strings are discarded from both the current value and `allowedValues`, because a `simple-array` column persisted from an empty array reads back as `['']` rather than `[]`.

Behaviour-only: **no GraphQL schema change, no entity change, no migration.** Verified that `pnpm run schema:print && pnpm run schema:sort` produces a diff *identical* to the one produced on an unmodified tree, i.e. pre-existing drift on `develop` (`CalloutContributorsMapView` descriptions) and nothing from this branch. `schema.graphql` is deliberately not touched here.

## Acceptance criteria

| AC | How it is met |
| --- | --- |
| Cross-space transfer where source state name is **absent** in destination &rarr; Callout assigned destination default state | `resolveDefaultTags()` now always yields a state the destination allows. Tests: *"should fall back to the first allowed value when the template carries no defaultSelectedValue"*, *"...when defaultSelectedValue is itself stale"*, plus the pre-existing *"should fall back to default when current tag absent from target template allowedValues"*. |
| Cross-space transfer where source state name is **present** in destination &rarr; Callout keeps that state name | Unchanged match-by-name path, still pinned by *"should preserve current tag when present in target template allowedValues"*. |
| Same logic when the destination's innovation flow was changed via template | `updateInnovationFlowStatesFromTemplate` &rarr; `updateInnovationFlowStates` already refreshes `flowStatesTagsetTemplate` (`allowedValues` + `defaultSelectedValue`) — that template is the same entity shared by the destination `CalloutsSet`, so the transfer reads the post-template state set. The new guard additionally covers the case where that refresh left an unusable default. |
| Transferred Callout is visible in the destination after the move | The assigned value is now guaranteed to be one of the destination's `allowedValues`, so it matches a phase filter in `filterCalloutsByClassificationTagsets` instead of falling through every tab. |
| Regression test added | 6 new cases in `classification.service.spec.ts` under *"destination default resolution (story #6021)"*, plus a transfer-boundary test in `callout.transfer.service.spec.ts` asserting the destination's `flow-state` template is what gets handed to the classification. |
| Verified against alkem-io/server#4970 and alkem-io/client-web#9353 | Both referenced as plain text only — **no closing keywords**, they stay open for manual closure. #4970's "transfer subspace&rarr;subspace makes it invisible" is the same code path and same symptom. |

## Known adjacent defect (not fixed here)

`InnovationFlowService.createStateOnInnovationFlow` and `InnovationFlowService.deleteStateOnInnovationFlow` mutate the flow's states but **never refresh `innovationFlow.flowStatesTagsetTemplate`**. Only the wholesale `updateInnovationFlowStates` and the rename branch of `updateStateOnInnovationFlow` do.

Consequence: adding or deleting a phase leaves the template's `allowedValues` drifting from the flow's real states. Because the transfer answers "does the source state exist in the destination?" against that template, a drifted `allowedValues` can make a **genuinely matching state look unmatched** (so the Callout is needlessly reset), or make a deleted state look still-valid (so the Callout keeps a phase that no longer exists) — which undermines AC#2 in the drifted case.

Deliberately left out of this PR: it is a defect in flow *editing* with a much wider blast radius (`allowedValues` is read by every consumer of the flow, not just transfer), and folding it in pushes this past bug-sized. Flagged for a separate issue.

## Test plan

- `pnpm exec vitest run src/domain/common/classification/classification.service.spec.ts src/domain/collaboration/callout-transfer/callout.transfer.service.spec.ts src/domain/collaboration/callouts-set/callouts.set.service.spec.ts` &rarr; **85 passed**
- Full suite (pre-commit hook): **7746 passed / 7 skipped, 687 files**
- `pnpm lint` (tsc --noEmit + biome check) &rarr; clean
- `pnpm build` &rarr; clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flow-state tag handling when templates have missing, outdated, or empty defaults.
  * Preserved valid selections through case-insensitive matching while applying the template’s canonical value.
  * Automatically selected the first valid option when no usable default was available.
  * Ensured newly created tagsets receive the correct defaults during callout transfers.
  * Treated empty existing tag values as invalid and resolved them appropriately.
  * Ensured destination callout templates provide the correct flow-state values.

### DIFF
--- a/src/domain/collaboration/callout-transfer/callout.transfer.service.spec.ts
+++ b/src/domain/collaboration/callout-transfer/callout.transfer.service.spec.ts
@@ -27,7 +27,7 @@ describe('CalloutTransferService', () => {
   let profileService: ProfileService;
   let tagsetService: TagsetService;
   let storageAggregatorResolverService: StorageAggregatorResolverService;
-  let _classificationService: ClassificationService;
+  let classificationService: ClassificationService;
   let _urlGeneratorCacheService: UrlGeneratorCacheService;
   let activityService: ActivityService;
   let entityManager: EntityManager;
@@ -54,7 +54,7 @@ describe('CalloutTransferService', () => {
     storageAggregatorResolverService = module.get(
       StorageAggregatorResolverService
     );
-    _classificationService = module.get(ClassificationService);
+    classificationService = module.get(ClassificationService);
     _urlGeneratorCacheService = module.get(UrlGeneratorCacheService);
     activityService = module.get(ActivityService);
     entityManager = module.get(EntityManager);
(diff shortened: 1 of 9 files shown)


hero101:
@coderabbitai review

PR #6354, opened 2026-08-12
fix: temporarily disable Collabora document analytics attribution (hotfix for ~8s open latency)

bobbykolev:
## Temporary hotfix

Opening or attributing a Collabora document made the user wait ~8s. Root cause: a best-effort lifecycle-analytics step `await`s `communityResolverService.getCommunityForCollaboraDocumentOrFail(...)` **inline before the response returns**. That method is a TypeORM `findOne(Space)` OR-ing two five-level relation paths — it took **7,917 ms** in a measured production trace. The reporter call itself is cheap; the expensive part is resolving the space to attribute the record to.

This PR **comments out** (does not delete) the four best-effort analytics `try/catch` blocks that call `getCommunityForCollaboraDocumentOrFail`, removing the ~8s penalty. `getCommunityForCollaboraDocumentOrFail` itself is left untouched.

### Sites disabled
1. `src/domain/collaboration/collabora-document/collabora.document.resolver.queries.ts` — `collaboraEditorUrl` query, `COLLABORA_DOCUMENT_OPENED`. **This is the primary user-facing ~8s penalty.** Still returns `editorUrl`.
2. `src/domain/collaboration/collabora-document/collabora.document.resolver.mutations.ts` — `replaceCollaboraDocument`, `COLLABORA_DOCUMENT_REPLACED`. Still returns the swapped document.
3. `src/domain/collaboration/callout/callout.resolver.mutations.ts` — `importCollaboraDocument`, `COLLABORA_DOCUMENT_UPLOADED` (the whole `if (contribution.collaboraDocument) { ... }` block). Still returns via `getCalloutContributionOrFail`.
4. `src/services/collaborative-document-integration/collaborative-document-integration.service.ts` — `reportOfficeDocumentWindow`, the **background RabbitMQ consumer** path (contribution + view events). **No user waits on this one**, but it paid the same expensive query per contribution/view event, so it's disabled here too (the method now returns early so the consumer still ACKs the event without firing the query).

### Effect
Analytics records for `COLLABORA_DOCUMENT_OPENED` / `COLLABORA_DOCUMENT_REPLACED` / `COLLABORA_DOCUMENT_UPLOADED` (and the background office-document contribution/view window records) are **temporarily suppressed**. All user-facing behavior — auth checks, editor-URL return, file swap, rename, contribution return values — is unchanged.

### Follow-up (restores analytics)
The proper fix restores/replaces this logic with a cheap leaf-first space lookup on branch `fix/110-collabora-editor-url-latency` (spec draft PR #6350, feature `110-collabora-editor-url-latency`). The blocks are commented rather than deleted so that follow-up can restore them in place.

### Tests
Existing unit assertions that the reporter is called on these paths would now fail, so the analytics-specific assertions have been skipped/narrowed (each with a comment pointing to the follow-up). Unrelated assertions were not weakened:
- `collabora.document.resolver.queries.spec.ts` — `it.skip` on the "report COLLABORA_DOCUMENT_OPENED … and return the editor URL" test (its purpose is the reporter assertion; return-URL behavior is covered by other tests in the file).
- `collabora.document.resolver.mutations.spec.ts` — commented out only the `calloutCollaboraDocumentReplaced` assertion inside the replace happy-path test; the auth / file-swap / rename / return assertions still run.
- `callout.resolver.mutations.spec.ts` — `it.skip` on the "report COLLABORA_DOCUMENT_UPLOADED" test.
- `collaborative-document-integration.service.spec.ts` — `describe.skip` on `officeDocumentContributions` and `officeDocumentViews` (they verify the now-suppressed window reporting).

Gates: `pnpm lint` (tsc native + biome) clean; `pnpm build` clean; `pnpm test` — 7759 passed, 30 skipped, 0 failed.

### Comments convention
This diff touches comments. Per the workspace convention (`docs/conventions/code-comments.md`), the added comments carry **no** spec/feature/issue/PR numeric references — the follow-up is described in prose and cross-referenced here in the PR instead.

No schema change, no migration, no new dependency.

### DIFF
--- a/src/domain/collaboration/callout/callout.resolver.mutations.spec.ts
+++ b/src/domain/collaboration/callout/callout.resolver.mutations.spec.ts
@@ -574,7 +574,13 @@ describe('CalloutResolverMutations', () => {
   });
 
   describe('importCollaboraDocument', () => {
-    it('should report COLLABORA_DOCUMENT_UPLOADED for the uploading actor', async () => {
+    // TEMP hotfix: COLLABORA_DOCUMENT_UPLOADED analytics attribution is disabled
+    // to remove the ~8s getCommunityForCollaboraDocumentOrFail penalty on the
+    // upload path, so the reporter is no longer called here. Re-enabled by the
+    // proper fix (a cheap leaf-first space lookup) in the
+    // collabora-editor-url-latency follow-up PR referenced in this PR's
+    // description.
+    it.skip('should report COLLABORA_DOCUMENT_UPLOADED for the uploading actor', async () => {
       const callout = {
         id: 'callout-1',
         authorization: { id: 'auth-1' },
--- a/src/domain/collaboration/callout/callout.resolver.mutations.ts
+++ b/src/domain/collaboration/callout/callout.resolver.mutations.ts
@@ -604,41 +604,46 @@ export class CalloutResolverMutations {
       );
     await this.authorizationPolicyService.saveAll(updatedAuthorizations);
 
-    // Lifecycle analytics (US4 / FR-013): record the upload as a single-actor
+    // TEMP hotfix: analytics attribution disabled to remove the ~8s
+    // getCommunityForCollaboraDocumentOrFail penalty on the document-upload
+    // path. The proper fix (a cheap leaf-first space lookup) restores this
+    // reporting; see the collabora-editor-url-latency follow-up PR referenced in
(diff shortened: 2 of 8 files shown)


bobbykolev:
Tracking bug for this temporary fix: #6356. Proper fix follows in #6350 (feature `110-collabora-editor-url-latency`).

PR #6469, opened 2026-09-05
feat: prepare memo signing previews

antst:
## Checkpoint status

This is the reviewed checkpoint 2 branch stacked on checkpoint 1 / #6468. Under the one-PR-per-repository delivery decision, this history will be incorporated without rewriting reviewed SHAs into canonical server PR #6468 after both checkpoint gates accept this commit. This PR must not merge independently.

## Summary

- Reuses the existing memo resolver, authorization policy, Yjs-to-Markdown projection, PR1 signing-attempt operations and file-service adapter.
- Checks the mandatory `cleverbase:` linked identity before live rendering, inserts the attempt row first, and stores a fresh private snapshot plus lowercase 64-character SHA-256.
- Uploads every signing snapshot with explicit `skipDedup: true`; losing-operation cleanup therefore owns the returned fileID and cannot remove another operation's bytes.
- Serves the initiating actor's exact snapshot through the conventional private REST module with 401/403/409 outcomes, `no-store`, and `nosniff`.
- Converts only supported projection elements. Raw authored SVG and other unsupported elements become visible labels before conversion; highlight syntax is transformed only in parsed text nodes outside code/pre.
- Resolves images only from the memo bucket after READ authorization. Each source retains the 15 MiB file cap and Sharp's 16,777,216-pixel guard, then normalizes sequentially to sRGB JPEG q80, white-flattened alpha, and at most 1200px on the longest edge. The memo limit is 20 images and 16 MiB total normalized JPEG bytes.
- Carries only opaque owned image keys through JSDOM/html-to-pdfmake, substituting authorized data after conversion. Pinned pdfmake then embeds exactly one DCT image XObject per input.

There is no gateway call, signing-session state, callback, signed copy, client UI, validation, deployment, or secret in this checkpoint.

## Tests first / RED evidence

The initial behavioral specifications precede implementation in `6cf625309`, `d0eb4ada8`, and `a67e8a9e7`; the first RED run failed on the absent service, renderer, REST controller, and actor-bound attempt read imports.

Later review corrections were exercised against `a0c5eb8` before their implementation. The discriminating failures were:

- an authored SVG containing `<image href="/local/file.png">` produced a PDF image object without any document lookup, READ authorization, or file-service read;
- the raw-Markdown highlight rewrite changed `==literal==` inside inline and fenced code;
- anonymous, wrong-actor, and unready preview outcomes escaped as 500 responses rather than 401/403/409;
- equal signing uploads could deduplicate to one fileID, allowing loser cleanup to remove another operation's bytes;
- passing normalized base64 through JSDOM/html-to-pdfmake reached 1,430.7 MiB RSS in the isolated 20-image stage; direct Buffer input avoided that amplification but incorrectly embedded 40 DCT objects for 20 inputs and doubled the PDF to 27,040,191 bytes.

The review-fix RED chronology is reported from the working-tree runs and architect witnesses; it is not represented as a separate test-only commit. No history was rewritten or retroactively described as an original RED commit.

## GREEN evidence at `b947bbc4b69460226d628fcd033a1f24855a1de2`

- Signed commit; the repository pre-commit hook passed typecheck, formatting, and the full `test:ci:no:coverage` suite.
- `pnpm run test:content-signing:coverage`: 14 files / 290 tests passed against actual PostgreSQL and the pinned real file-service image.
- Review-time changed executable coverage from `coverage-final.json`: 160/160 lines, 174/174 statements, 31/31 functions, and 51/51 branches (100% each). Decorator-only module wiring has no V8 executable denominator and is verified by typecheck/build.
- Scoped coverage JSON SHA-256: `c7141b4e9985452d478abe27e471e2b0ada15f92612ad9b984457ae0690e92ce`.
- `pnpm vitest run` over the 13 focused unit/renderer files: 277/277 passed.
- `pnpm lint`: passed (native TypeScript plus Biome over 3,316 files).
- `pnpm build`: passed.
- `pnpm schema:print && pnpm schema:sort`: passed; the schema diff is one additive, described `prepareMemoSigning` mutation and its result type. Standalone `schema:validate` requires CI-produced `change-report.json`/`deprecations.json`; their validator tests passed in the full suite.

The real-service matrix includes row-first prepare versus actual `MemoService` deletion, cleanup/retry, conditional finalization, same-byte concurrent preparations with independent fileIDs, and an actual authorized image through projection -> renderer -> real file-service -> actor preview with exact stored/preview bytes and hash.

## Renderer performance and resource evidence

Fresh-process Apple measurements of the complete production renderer:

| fixture | elapsed | peak RSS | PDF / image objects |
|---|---:|---:|---|
| baseline | 58.9 ms | 255.5 MiB | 6,533 bytes |
| 100 KB structured text | 547.2 ms | 314.1 MiB | 374,433 bytes |
| 5 real screenshots/photos + 100 KB | 827.9 ms | 372.0 MiB (+116.5) | 661,038 bytes; 5/5 DCT |
| one compressible 4096x4096 source | 122.2 ms | 352.0 MiB (+96.5) | 14,897 bytes; 1/1 DCT |
| 20 distinct high-entropy 2048x2048 sources + 100 KB | 1,811.0 ms | 478.8 MiB (+223.3) | 13,901,310 bytes; 20/20 DCT |

The retained local coverage run rendered the representative 50 KB / five-repository-image fixture in 1,076.9 ms and the 100 KB / 20-image fixture in 2,197.0 ms under coverage instrumentation, both under the 10-second target and with one DCT XObject per image. CI must retain its own runner timings; one local sample is not called p95.

The <150 MiB increment criterion is workload-specific to the realistic fixture and passed. The allowed high-entropy maximum is explicitly +223.3 MiB and remains a documented residual, not a universal <150 MiB claim. The 15-second collaboration and 30-second per-file-service timeouts can accumulate; there is no end-to-end 30-second cancellation guarantee or synchronous-layout worker.

## RCA / ownership boundaries

The linked-identity check is before Yjs read/render because signing cannot proceed without the provider binding. Rendering first would spend resources and create evidence for an ineligible actor.

The attempt is inserted before live read so concurrent memo deletion either cascades the row before upload finalization or encounters the retained-row/file preflight. Conditional finalization proves the memo/attempt still exists. When it loses, compensation deletes only the fresh `skipDedup: true` upload; no lock, refcount, state, or platform-wide dedup change was added.

The SVG exfiltration was caused by blocklist sanitization: html-to-pdfmake preserved authored SVG and svg-to-pdfkit opened nested local image references outside pdfmake's URL/local access policies. The fix is the supported projection-element allowlist at the renderer boundary, not an SVG sanitizer subsystem.

The memory spike was caused by base64 image strings traversing JSDOM/html-to-pdfmake. Opaque keys now cross the DOM, followed by one narrow post-conversion substitution. Direct Buffers were rejected because pinned pdfmake double-embedded them. Sequential normalization and the 16 MiB normalized-byte bound limit the owned producer before layout; the bound exception is preserved instead of being swallowed by the unsupported-image fallback.

## Known representation limits

The current Yjs-to-Markdown projection flattens table spans and non-paragraph editor nodes. Images in another Alkemio bucket fail preparation. Roboto covers the tested Dutch, Spanish, Bulgarian Cyrillic, German, French and Greek samples; symbols, emoji and CJK outside the bundled font render a visible replacement box in preview and signed PDF. PDF text extraction may omit an unmapped glyph, but silent visual omission is not accepted. A broader bundled font is a later product choice, not a font framework in this checkpoint.

## LOC accounting

Direct tree diff from checkpoint 1 final SHA `646c2ab4a36f64a3d9b228e56dd4aef44d5a1a66`:

- handwritten production/config: 474 additions, 9 deletions, within the approved checkpoint 2 ceiling of 475;
- tests and test configuration: 1,684 additions, reported separately and never reduced to fit the production ceiling;
- generated schema: 7 additions;
- generated lockfile: 158 additions;
- checkpoint 3 remains capped at 215; checkpoint 2+3 cap at 690, checkpoints 1-9 at 1,920, and the complete feature at 2,020. Consolidating checkpoints into one repository PR does not reset any slice ceiling.

## Not covered here

- Gateway start/status/result and canonical signer matching are checkpoint 3.
- Callback/complete/final signed-copy orchestration are checkpoint 4.
- Client UI is checkpoint 5, local quickstart is checkpoint 6, dev orchestration is checkpoint 7, the single infra-ops PR carries checkpoints 8/9, and on-demand Verify is checkpoint 10a/b.
- No merge, deployment, ACC/PROD mutation, or residual-risk acceptance is authorized.


antst:
@coderabbitai full review

antst:
@coderabbitai full review

Issue #71, opened 2020-09-16
Tags creatable + updateable 


techsmyth:
## Description
As a user of the api, I want to be able to add tags to any taggable entity so that I can via a compact api work with the data

## Acceptance criteria

- [ ] ability to add tags to the followinging entities: User, UserGroup, Ecoverse, Challenge, ...

## Additional Context
Note: may involve creating a "Profile" type entity on  the above entities

PR #6509, opened 2026-09-14
fix(forum-hierarchy): authorize reconcile removals and fix the completion contract

bobbykolev:
Follow-up to #6484, addressing the architect review on [alkem-io/matrix-adapter#71](https://github.com/alkem-io/matrix-adapter/pull/71) (findings 1, 2, 3 and 6). Pairs with the adapter change in that PR — **merge order: adapter first, then the lib release, then this.**

## The problem

The reconcile pass authorized removals **by omission**. It sent each category its desired child set, and any child the adapter found attached but absent from that set became a removal candidate.

The desired set is a snapshot read once, before the pass starts. So "absent from the snapshot" covers two very different things:

- a discussion that genuinely moved away — should be removed;
- a discussion created or recategorised **while the pass was running** — must not be.

The second kind lost its only edge into the hierarchy, and the pass reported `COMPLETED, failed=0` while doing it. No timeout is involved, and no retry recovers it — the edge is simply gone. Add-then-remove ordering does not help: it bounds *when* the removal happens, not *whether* it was ever justified.

## The fix

Removals are now authorized per room, and the authorization comes from positive establishment rather than from absence:

```
removable(category C) = every room the add phase confirmed attached under some category other than C
```

One precondition, both High findings:

- A room created or recategorised after the snapshot is in **no** category's established set, so nothing authorizes removing it and its correct edge survives (finding 1).
- A room whose move failed — destination space missing, or the add rejected — is likewise unestablished, so its **old** edge is kept rather than dropped in favour of a destination that never materialised (finding 2).

A failed call contributes nothing to the established set, not even partially. The response reports which desired ids did not resolve, but not which individual writes were rejected, so after a failure the pass cannot tell an established room from a refused one — and guessing permissively is exactly what authorizes removing a room's last edge.

**Residual risk within a single pass** is bounded and self-healing: the worst outcome is a discussion sitting under its *previous* category for one pass, converging on the next. Cross-pass safety needed a separate fix — see the correction at the end of this description.

## Also in this change

- **`SPACE_NOT_FOUND` is a natural skip only for an empty category.** A retired category resolves to no space on every pass by design and must not block termination. A category that still holds discussions and has no Matrix space is incomplete work — those rooms have nowhere to be attached. Space creation stays in the provisioning/sync path; this reports the prerequisite rather than silently creating anything.

- **Completion is judged on convergence, not the absence of errors** (finding 6). Unresolved desired rooms, extra edges kept because nothing authorized removing them, and outstanding pointer repairs all leave `failed=0` while leaving real work undone — so a repeat-until-clean runbook was told to stop while the forum was still drifted. The final sweep's `unconverged` count now drives task status. The audit row carries it alongside a separate `parentPointersUnprocessable` count, kept apart from deferrals because no future pass can clear it. A dry run is exempt: reporting drift is its purpose.

- **`CommunicationAdapter` stamps an absolute expiry** on every `set_children` request, derived from the RPC timeout governing that same call — so the previously divergent 8s / 10s / 30s envelope has one source of truth (finding 3). The adapter's own deadline starts when it dequeues, so without this a request that outwaited the caller would still execute, writing from a snapshot the caller had already superseded.

- **`converged` defaults to `false` when absent from the wire**, so an adapter predating the field reports "not finished" rather than a convergence nothing verified.

- **An in-process guard refuses a second overlapping pass.** Deliberately **not** a distributed lock and **not** a guard across the 6 API replicas — see the comment on `passInFlight`. It is sufficient only because the authorization rule above removes the data-loss outcome, leaving duplicated work and two confusing audit rows as the worst case.

## Deliberately not addressed

**Finding 5 (write-budget accounting)** — `maxOperations` stays advisory rather than enforced at the write site. Tracked in alkem-io/alkemio#2121 with the reasoning.

**Finding 4 (planning cost exhausting the deadline)** — not guessed at. It needs real category fan-out and alias latency measured against a restored ACC dataset before an execution envelope can be chosen honestly.

## Testing

New spec `admin.communication.forum.hierarchy.reconcile.authorization.spec.ts` covers the raced child that must survive, per-room authorization scope, both destination-failure shapes (missing space and rejected write), an unresolved room, the forum-level call authorizing nothing, the completion contract under deferred and unprocessable work, dry-run exemption, and the reentrancy guard including its release after a throw.

The moved-room authorization test was **confirmed RED against the previous behaviour** — it is discriminating, not a restatement of the implementation.

One existing expectation changed on purpose: a pass over a forum that holds a discussion in a category with no Matrix space now completes `ERRORED` rather than `COMPLETED`. A companion test pins that a genuinely empty category still skips cleanly.

`tsc --noEmit` clean, biome clean, 173 tests green across the changed areas.

## Related

| | |
|---|---|
| alkem-io/matrix-adapter#71 | Adapter half — the `removable_child_context_ids` / expiry / `converged` protocol this depends on. **Must merge first** |
| alkem-io/agents-hq#146 | Spec 061 correction — the concurrency claim that asserted a safety property which did not hold |
| alkem-io/alkemio#2121 | Finding 5 follow-up (budget accounting), deliberately out of scope here |
| alkem-io/alkemio#2114 | Story |
| #6484 | The PR this follows up; it merged ahead of the adapter, which is why `develop` already calls a topic no deployed adapter handles |

## Sequencing

The two adapter artifacts have different triggers: `publish-lib.yml` fires on a **git tag** `v*` and publishes the npm lib; `build-release-docker-hub.yml` fires on a **GitHub Release** and builds the Docker image. Creating a Release does both, since it creates the tag too.

1. **matrix-adapter#71 merges** to its `develop`.
2. **Hetzner dev deploys itself** — `build-deploy-k8s-dev-hetzner.yml` runs on push to `develop` and does `kubectl set image` with the commit SHA. No `dev-orchestration` PR is needed; its base manifests are bootstrap defaults that drift and are not the live source of truth.
3. **Tag + release `v0.8.19`** on matrix-adapter → npm lib published, Docker image built.
4. **Repin this PR** from the `pkg.pr.new` preview to `0.8.19`, push, wait for green, merge.
5. **agents-hq#146** merges independently (docs only).
6. **ACC / PROD are GitOps** — bump `alkemio/matrix-adapter:v0.8.17 → v0.8.19` in `infrastructure-operations` alongside the server release carrying this PR.

### Why the `test` check is red

`check-vendor-lib-pins.sh` rejects the preview specifier — correctly, since a preview artifact cannot produce a release image. Because that guard is the **first step in the job**, nothing after it runs:

```
failure  Block non-registry dependency pins on protected branches
skipped  Install dependencies
skipped  Run tests with coverage
```

So this PR has **no CI test signal at all** until step 4. Standing in for it, the full suite was run locally against the exact preview package pinned here: **799 files / 9426 tests passed, 28 skipped**, plus `tsc --noEmit` and biome clean. I will not ask for a merge before that check is green on its own.

## Operator runbook, once deployed

1. `dryRun: true` first (the default) and read the report. On a converged forum it returns all zeros and writes nothing — confirmed against restored ACC data, see the evidence on matrix-adapter#71.
2. Only then apply. Repeat until **`unconverged` reaches 0** — that is the termination condition now, not `failed == 0`.
3. **Prerequisite:** `newsletter` and `tips-and-tricks` have no Matrix space on ACC. Empty today so they skip cleanly, but the moment a discussion lands in one the pass reports `ERRORED` rather than silently succeeding. Space creation belongs to the provisioning/sync path and must run first.

---

## Correction — cross-pass safety (2026-09-14, second review)

An earlier revision of this description claimed the residual risk was "bounded and self-healing … it can never end up under none." **That was wrong**, and it is retracted.

Removal authorization is built from confirmations gathered earlier in the same pass. Those are *historical observations*, not live guarantees — a concurrent pass can invalidate one before it is consumed. With a room moved A → B between two passes' snapshots, and every adapter call processed sequentially:

| step | pass P (snapshot: A) | pass Q (snapshot: B) | edges |
|---|---|---|---|
| 1 | confirms A, incl. its destructive visit to A; pauses | | A |
| 2 | | establishes B | A, B |
| 3 | | removes A, authorized by its confirmation of B | B |
| 4 | | finishes | B |
| 5 | resumes at B, removes B, authorized by its earlier confirmation of A | | **none** |

The room ends with no category edge and **both passes report `COMPLETED, failed=0, unconverged=0`**. `converged` does not help — it is computed per pass against a view that was already stale. No concurrent Matrix write, timeout, failed request, ghost prune or pointer repair is involved.

The in-process guard never prevented this, because the second pass runs in a different API replica with its own flag. The argument that the flag was "sufficient because nothing can be lost" rested on exactly the premise that turned out to be false.

**Fixed by mutual exclusion across replicas.** A Redis lease keyed on the forum (NX + PX + owner token, the same idiom as the OIDC refresh mutex), re-confirmed immediately before every call that can remove an edge; losing it is a hard stop (`aborted: 'ownership-lost'`) rather than a reason to consume authorizations gathered under someone else's lease. Execution is fenced by reusing the request expiry already in the protocol: each request carries the lease's own expiry, and a new owner can only acquire after the previous lease fully expired — so every request the previous owner issued has expired too, and the adapter rejects it before any read or write. No hand-off protocol needed.

Redis rather than a Postgres advisory lock: a session-scoped lock is unsafe on a pooled connection, and a transaction-scoped one would mean holding a transaction open across Matrix RPCs. Redis unreachable refuses the pass rather than running it unowned.

Tests: the exact interleaving with two service instances over one shared Redis, ownership loss mid-pass, release on completion and on throw, and the expiry bound. The mutual-exclusion test was confirmed to fail when mutual exclusion is disabled. The old `reentrancy` block was removed rather than updated — it covered the in-process flag this replaces and could not have caught what that flag failed to prevent. Full server suite green: 800 files / 9430 tests.

Spec 061 is corrected in alkem-io/agents-hq#146, including a standing note not to restore any "never parentless" assertion unless the interleaving above is demonstrably prevented for the mechanism being described.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved forum hierarchy reconciliation reliability across multiple application instances.
  - Prevented overlapping reconciliation runs from making conflicting changes.
  - Safely stops destructive updates if coordination is lost while preserving non-destructive repairs.
  - Added request expiry handling to prevent stale operations from modifying forum data.
  - Improved handling of incomplete or missing reconciliation responses and status reporting.

- **Reliability**
  - Reconciliation coordination is now maintained through renewable ownership, with safe cleanup after successful or failed runs.


bobbykolev:
### On the red `test` check

It is the vendor-pin guard, not a test failure — and worth flagging that it hides the suite entirely:

```
failure  Block non-registry dependency pins on protected branches
skipped  Install dependencies
skipped  Native TypeScript 7 typecheck gate
skipped  Run tests with coverage
```

The guard is the first step in the job, so nothing after it runs. This PR therefore gets **no CI test signal at all** until matrix-adapter#71 merges, `v0.8.19` publishes, and the pin here moves from the preview to the release. That is the guard working as designed (a preview artifact cannot produce a release image), but it does mean the red check should not be read as evidence either way.

Since CI cannot supply that signal yet, the full suite was run locally against the exact preview package this branch pins (`612208c`, verified to contain the new contract fields):

```
Test Files  799 passed | 8 skipped (807)
     Tests  9426 passed | 28 skipped (9454)
```

Plus `tsc --noEmit` clean and biome clean. The 173 tests in the changed areas (platform-admin/communication + the communication adapter) are included in that run.

Adapter side is independently green in CI on `c7e6a5c` — lint/test, container build, and library publish, including the openapi staleness gate.

The real CI run for this branch happens after the repin in step 3 of the merge order in the description. I will not ask for a merge before that check is green on its own.

antst:
**Merge/release requirement: once matrix-adapter#71 is merged and its fixes are tagged/released, this server PR must bump `@alkemio/matrix-adapter-lib` to the resulting published npm version before merging.** Updating or deploying the Go adapter does not update the library installed by the server.

Currently `package.json` pins:

```json
"@alkemio/matrix-adapter-lib": "https://pkg.pr.new/@alkemio/matrix-adapter-lib@612208c"
```

That preview is useful for development, but it must be replaced with the exact registry release containing the final adapter contract and safety fixes—currently planned as `0.8.19`.

Required sequence:

1. Resolve the remaining [cross-replica safety blocker on adapter #71](https://github.com/alkem-io/matrix-adapter/pull/71#issuecomment-5665735620), including its server-side changes, and finalize the shared contract.
2. Merge adapter #71 and tag/release the resulting commit. Confirm the **npm library publish** succeeds; a Docker image or GitHub Release alone is not evidence that the package is available.
3. In this server PR, replace the preview with the exact published version in **`package.json` and `pnpm-lock.yaml`**. For the currently planned version:

   ```bash
   pnpm add --save-exact @alkemio/matrix-adapter-lib@0.8.19
   ```

   If the final release number changes, use that actual release containing the fixes.
4. Push the dependency/lockfile update and rerun CI against that registry artifact. Verify the installed package contains `removable_child_context_ids`, `expires_at_unix_ms`, `operation_id`, `converged`, and `parent_pointers_unprocessable`, plus any protocol additions needed by the remaining fix.
5. Merge the server follow-up only after its checks pass on that final dependency. Deploy the compatible adapter service before the consuming server release.

**Why this matters:** this server code relies on the new request/response contract; generating or releasing the producer's library does not automatically change the consumer's pinned dependency. The release must also be reproducible from registry dependencies. The current `test` job fails at the non-registry dependency guard, before installing dependencies or running tests, so the local test results against the preview do not replace a green CI run against the final published package. Keep that guard intact and satisfy it with the release-version bump.

bobbykolev:
@antst you were right, and the counterexample reproduces exactly in this code — I traced it rather than taking it on faith. `removableForCategory` unions `established` across the whole pass, so a confirmation recorded in phase A authorizes removing that room from **every** other category for the rest of the pass, including the one that legitimately holds it by then. Fixed in `6d1825a`.

**Retracted:** my "can never end up under none" claim. It was wrong, and it was wrong in three places — this PR, matrix-adapter#71, and spec 061 (alkem-io/agents-hq#146). All three now carry the interleaving table and, in the spec, a standing note not to restore the assertion unless it is demonstrably true for whatever mechanism is being described. I had verified single-pass safety and over-generalised it.

**The fix** is your requested change 1, done the way the repo already does locking: a Redis lease keyed on the forum (NX + PX + owner token, same idiom as `core/auth/oidc/refresh-lock.ts`), taken before the pass reads anything, re-confirmed immediately before every call that can remove an edge, hard-stopping as `ownership-lost` if it is gone. Redis rather than a Postgres advisory lock because a session-scoped lock is unsafe on a pooled connection and a transaction-scoped one would mean holding a transaction across Matrix RPCs, which you ruled out.

**On your point 2 (outstanding work when ownership changes)** — I did not add a hand-off protocol, because the expiry from the last round already gives execution-side fencing for free. Each request now carries the *lease's* expiry rather than one derived independently. A new owner can only acquire after the previous lease fully expired, so every request the previous owner issued has expired too, and the adapter rejects an expired request before performing any read or write. If you think that argument has a hole, the residual I can see is a single already-issued Matrix write landing within one Synapse round-trip of expiry — say so and I will add an explicit grace on acquire.

**Your point 3** is implemented as `admin.communication.forum.hierarchy.reconcile.cross.pass.spec.ts`: two service instances over one shared Redis, P held mid-pass while Q attempts to start, plus ownership-loss mid-pass, release on completion and on throw, and the expiry bound. I verified it is discriminating — disabling mutual exclusion turns the key test red. The old `reentrancy` block is deleted rather than updated, since it tested the in-process flag this replaces and could not have caught what that flag failed to prevent.

Two things I deliberately did not do, so you can reject them explicitly:

- **The add-only phase is not gated on ownership.** It can only ever create an edge the desired state asks for, so gating it seemed like over-correction. There is a test pinning that it stays ungated.
- **No destructive ACC run yet.** ACC currently has zero drift, so a dry run structurally cannot exercise this path — proving it needs a seeded recategorisation on the restored dump. Happy to do that before you merge if you want it.

Full server suite green: 800 files / 9430 tests. Note this PR's `test` check stays red until the lib is published and the pin moves off the preview — the vendor-pin guard is the job's first step, so nothing after it runs.

bobbykolev:
### Destructive path exercised on restored ACC data

The earlier ACC run only proved the safe path — the forum was already converged, so the pass wrote nothing. Seeded a real recategorisation on the restored dump (`challenge-centric` → `releases`, database only, leaving the Matrix edge stale) and ran the removal path for real.

| pass | drifted | repaired | unconverged | writes |
|---|---|---|---|---|
| dry run | 2 | 0 | 2 | none |
| apply | 2 | 2 | **0** | add + remove |
| apply again | 0 | 0 | 0 | none (idempotent) |

**Add strictly precedes remove**, from Synapse's own event stream rather than from the adapter's own account of itself:

```
stream_ordering | parent                      | op     | at
198619          | #f8a78645…  (releases)      | ADD    | 16:29:17
198620          | #69816afe…  (challenge-c.)  | REMOVE | 16:29:18
```

So the room held its old edge until the new one existed. Final content confirms the re-parenting: `releases` carries `{"via":["matrix-acc.alkem.io"]}`, `challenge-centric` carries `{}`. All four audit rows persisted this time (the earlier harness passed a non-uuid actor id, which failed the insert open).

**Correction to the ACC baseline I posted earlier.** Those queries counted `m.space.child` state rows without inspecting content. Matrix removes a child by writing an *empty-content* `m.space.child`, which still occupies a state row — so a tombstoned edge would have been counted as a live one, and I would have reported drift as clean. Re-checked with content filtering: ACC carries **zero tombstones**, so the numbers in that comment were right. They were right by luck rather than by method, and the corrected query is what the table above uses.

**Still not covered live:** the cross-pass interleaving itself. Reproducing it needs two passes racing against one Synapse, which the ownership lease now prevents by construction — so the coverage for that is the unit-level spec with two service instances over one shared Redis, plus the confirmation that it fails when mutual exclusion is disabled. If you want the lease exercised against a real Redis rather than a shared fake, say so; it is the one gap I know of.

antst (dismissed):
lets roll with it

